PR #14, opened 2024-10-14
Work on general purpose JSONB support

coderdan:


README.md
```diff
@@ -222,17 +222,94 @@ CREATE TABLE users (
...
+SELECT * FROM users WHERE cs_ste_vec_v1(attrs) @> '53T8dtvW4HhofDp9BJnUkw';
+```
+
+And is the EQL equivalent of the following plaintext query.
+
+```sql
+SELECT * FROM users WHERE attrs @> '{"field": "value"}`;
+```
+
+#### `cs_ste_term_v1(val JSONB, epath TEXT)`
```
CDThomas, line 258: Could we make `epath` here json/jsonb as well? The motivation is that the JSON value would include all the info needed for generating the MAC so that Proxy doesn't need to sort out the table, column, etc based by looking at anything other than the param or literal used for the arg.

I think that we could aim to target text in SM2, but using JSON here similar to how encryption already works would simplify the Proxy (MLP) logic.

CDThomas: Also, what's the return type of `cs_ste_term_v1 `? `ore_64_8_v1`?

coderdan:
Note to self: add the things that the simplified, eJSON path doesn't support.

README.md
coderdan: Yeah, I think for now that makes the most sense. I considered adding a variant of `cs_ste_term_v1` which also takes a term type so we can handle other types of index terms but for now this should be enough. Does that sounds reasonable?

JSON.md
```diff
@@ -0,0 +1,132 @@
...
+
+```json
+{
+    "firstName": "John",
+    "lastName": "doe",
+    "scores": [1, 2, 3]
+}
+```
+
+`$.firstName` returns `[John]`
```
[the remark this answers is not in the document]
coderdan: Just to keep the same behaviour as standard JSONPath.
Screenshot 2024-10-15 at 2 10 53 PM

I used https://jsonpath.com/

PR #16, opened 2024-10-16
Go/Xorm examples with separate go package for helper functions

fimac:


go/xorm/main.go
```diff
@@ -0,0 +1,275 @@
...
+// When serializing, the table and column for that field needs to be passed.
+// So the json configuration has the correct table and column when passed to the proxy.
+//
+// This setup means that for each field that needs to be encrypted:
+// - a separate custom type needs to be created
+// eg `type EncryptedTextFieldOne string`
+//    `type EncryptedTextFieldTwo string`
+//
+// - ToDB and FromDB needs to be implemented for each custom type.
+
```
fimac, line 55: There might be a better way to do this.

I wanted to ensure that insert statements didn't need to be changed and only plaintext values were retrieved. Using the conversion interface allows us to convert from the go app type (e.g string) to the underlying db type (jsonb).

To do that, I needed to create a custom type, to then call the ToDb/FromDb method on.

The serialization/deserialization functions have been extracted to the goeql package currently sitting locally in the same folder (naming to tbc), so they can all be reused.

PR #18, opened 2024-10-17
Update docs and JSON schema with "query-only" encryption changes

CDThomas:
This change updates the EQL query docs to use the new field `"q"` (for query). This field tells proxy to perform encryption for a specific query operation (instead of performing source encryption and encryption for all indexes).

This change also updates the JSON schema:
- add new "for query" field
- add missing STE vec index field
- remove ".1" suffix from m, u, and o index fields
- add variants for the encrypted payloads for queries. These variants are used internally by proxy for query-only encryption (for filters in `WHERE` clauses, for example).


src/cs_encrypted_v1.schema.json
```diff
@@ -15,7 +15,7 @@
     "k": {
       "title": "kind",
       "type": "string",
-      "enum": ["pt", "ct", "en"]
+      "enum": ["pt", "ct", "en", "qm", "qo", "qu", "qejp", "qsv"]
```
CDThomas, line 18: When "for query" is set on the plaintext payload, Proxy will encrypt the payload as one of the new variants here:
- `qm` = query match
- `qo` = query ORE
- `qu` = query unique
- `qejp` = query EJSON path
- `qsv`= query STE vec

Proxy will only generate the index needed for the specific query and set the relevant kind/index field in the payload.

For example, if the client sends through a plaintext payload with "for query" set to "match":
```js
{
  "v": 1,
  "k": "pt",
  "p": "abc",
  "i": {
    "t": "users",
    "c": "email_encrypted"
  },
  "q": "match"
}
```

Proxy would encrypt that as the match query variant:
```
{
  "v": 1,
  "k": "qm",
   "m": [1, 2, 3]
}
```

Note that the `"m"` key in the match query payload lines up with the same key used for the match index that's stored in the column. This allows the EQL SQL funcs to look up the index field using the same key (whether the value comes from the column or from a param):
```sql
-- $1 here should be encrypted for a match query
SELECT * FROM users WHERE cs_match_v1(email_encrypted) @> cs_match_v1($1)
```

tobyhede: Are the new `kinds` required? 
The postgres function shouldn't need the kind as long as the actual field is encoded correctly.

README.md
```diff
@@ -74,22 +74,52 @@ These statements must be run through the CipherStash Proxy in order to **decrypt
...
+{
+  "v": 1,
+  "k": "pt",
+  "p": "test@test.com",
+  "i": {
+    "t": "users",
+    "c": "email_encrypted"
+  },
+  "q": "match"
+}
```
coderdan, line 119: Nice. I like that.

README.md
```diff
@@ -103,15 +133,15 @@ Enables basic full-text search.
 
 ```rb
 # Create the EQL payload using helper functions
-payload = eqlPayload("users", "encrpyted_field", "plaintext value")
+payload = EQL.for_match("users", "encrypted_field", "plaintext value")
```
coderdan, line 136: There are probably more idiomatic ways to do this in Ruby but not something we need to worry about now. Just sharing thoughts while they're fresh.

You could do:
```rb
EQL.table("users").column("encrypted_field").match("plaintext value")
```

If you made `EQL.table("users")` a class method:
```rb
class User
  def self.eql
    EQL.table("users")
  end
end
```
And used `method_missing` to handle the column name, you could do something like:
```rb
User.eql.encrypted_field = "foo"
User.eql.encrypted_field > 10
# etc
```
Just ideas for later.

coderdan (approved):
Nice job! I like the approach. Left some musings on Ruby DSL improvements but not needed for the moment.

tobyhede:
I would make the `q` field be a valid index/encoding type and add the json path as a new accepted encoding. 

```
"q": "m"
```

README.md
```diff
@@ -506,19 +551,20 @@ The format is defined as a [JSON Schema](src/cs_encrypted_v1.schema.json).
...
+| m     | Match index       | Ciphertext index value. Encrypted by proxy.                                                                                                                                                                                                                               |
+| o     | ORE index         | Ciphertext index value. Encrypted by proxy.                                                                                                                                                                                                                               |
+| u     | Unique index      | Ciphertext index value. Encrypted by proxy.                                                                                                                                                                                                                               |
+| sv    | STE vector index  | Ciphertext index value. Encrypted by proxy.                                                                                                                                                                                                                               |
```
tobyhede, line 567: We reworked all the field names before I went on AL
https://github.com/cipherstash/cipherstash-suite/pull/976/files

src/cs_encrypted_v1.schema.json
CDThomas: They aren't strictly required (we could re-use the existing `"ct"`/ciphertext variant), but I'd prefer to model this as new variants for a few reasons.

On the Proxy/Rust side:
- The source ciphertext (`"c"` field) in the existing `"ct"`/ciphertext variant would have to become optional (since queries wouldn't use it)
- All of the indexes are optional in the existing `"ct"`/ciphertext variant, which is bit awkward to work with when encrypting for a query because behaviour-wise only one (required) index field should be set in the payload depending on the operation
- Basically, specific variants for each makes it harder for us to construct something incorrectly and means that we can lean more on the compiler to catch problems

On the PG side:
- If PG attempts to store a `q*` variant, something has gone wrong. It's convenient to be able to reject anything that isn't the `ct` variant
- It's nice for debugging to be able to be able to identify with specific variant was chosen

CDThomas:
> I would make the `q` field be a valid index/encoding type and add the json path as a new accepted encoding.
> 
> ```
> "q": "m"
> ```

I think this might not work for use cases like `websearch_to_tsquery` support, @tobyhede. Our `websearch_to_tsquery` implementation will use the same output type as a regular match index, but will have some extra tokenisation logic that only needs to happen for queries.

README.md
CDThomas: Ah oops. Happy to update when that change lands.

src/cs_encrypted_v1.schema.json
tobyhede: I think having variants for every option adds complexity. 
We could have a single "query" variant to disambiguate from ct and pt.

Coupled with having the "q" value be the identifier, we can then validate. 

```
"knd": "qt", 
"qry": "u",  
"u": "etc",
```

if knd == qt
  qry is required and qry.value (u) should exist as a field.

tobyhede:
@CDThomas Do you have an example of what that means?

README.md
CDThomas: Very fair. At the moment I'm mostly aiming for examples that get the point across and would be easy to port to other languages. Agreed that it'll make sense to implement a more idiomatic/ergonomic interface for Ruby at some point.

src/cs_encrypted_v1.schema.json
CDThomas: I think that would be an OK middle ground if we're concerned about the number of variants.

Conceptually there's a variant for each and I rather make that explicit than have to model it as a single type that requires extra validation logic to handle what the compiler could catch for us.

CDThomas:
> @CDThomas Do you have an example of what that means?

`websearch_to_tsquery` is a PG function that allows basic "web search" syntax on top of Postgres full text. There are some docs [here](https://www.postgresql.org/docs/current/functions-textsearch.html).

For example, the search `'cat or dog'` could be used to search for documents that contain the words "cat" or "dog" in them.

In our implementation, the search `'cat or dog'` will need to be handled differently to the text `'cat or dog'` that's stored in a  column (`or` as a logical operator vs as a word in the text itself). Both will use match indexes/bloom filters, but the search needs some extra parsing/tokenisation logic.

README.md
calvinbrewer: Worth just writting all examples in psuedo code?

README.md
```diff
@@ -74,22 +74,52 @@ These statements must be run through the CipherStash Proxy in order to **decrypt
 **Example:**
 
 ```rb
-Users.findAll(&:encrypted_email)
```
calvinbrewer, line 77: @CDThomas I was trying to show a basic SELECT statement here withour any WHERE clauses and highlighting that responses are simply decrypted. I like what you've done here but maybe for the `cs_match_v1` example?

CDThomas: That makes sense @calvinbrewer. I reverted the changes in this section (and already updated the match example).

CDThomas:
Merging this one since the docs here reflect what we've already started to implement on the Proxy side, but we can adjust the docs if we end up simplifying more.

PR #22, opened 2024-10-18
Move over files from migrator (includes fix for indexes clobbering)

fimac:
Moves over all the sql files from migrator, tests, justfile etc.

Also includes fix for this https://linear.app/cipherstash/issue/CIP-854/adding-indexes-for-an-additional-table-clobbers-the-1st-table-config

Fix was merged in a while ago, but got reverted somewhere along the way.
https://github.com/cipherstash/cipherstash-suite/pull/979/files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,3 +180,6 @@ cipherstash-proxy.toml
 
 # turbo repo
 .turbo
+
+# ignore release folder
+release/
--- a/.gitignore
+++ b/.gitignore
@@ -180,6 +180,3 @@ cipherstash-proxy.toml
 
 # turbo repo
 .turbo
-
-# ignore release folder
-release/
--- a/justfile
+++ b/justfile
@@ -1,0 +1,38 @@
+set dotenv-load
+set positional-arguments
+
+
+test_dsl:
+  #!/usr/bin/env bash
+  set -euxo pipefail
+
+  PGPASSWORD=$CS_DATABASE__PASSWORD dropdb --force --if-exists --username $CS_DATABASE__USERNAME --port $CS_DATABASE__PORT cs_migrator_test
+  PGPASSWORD=$CS_DATABASE__PASSWORD createdb --username $CS_DATABASE__USERNAME --port $CS_DATABASE__PORT cs_migrator_test
+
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -179,4 +179,7 @@ dist
 cipherstash-proxy.toml
 
 # turbo repo
-.turbo
\ No newline at end of file
+.turbo
+
+# ignore release folder
+release/
```
CDThomas, line 185: Might be worth checking in the release DSL file and pointing individual language examples to it. We could also add a check in CI that the generated file is up to date.

justfile
```diff
@@ -0,0 +1,38 @@
...
+  connection_url=postgresql://$CS_DATABASE__USERNAME:@localhost:$CS_DATABASE__PORT/cs_migrator_test
+  PGPASSWORD=$CS_DATABASE__PASSWORD psql $connection_url -f sql/dsl-core.sql
+  PGPASSWORD=$CS_DATABASE__PASSWORD psql $connection_url -f sql/dsl-config-schema.sql
+  PGPASSWORD=$CS_DATABASE__PASSWORD psql $connection_url -f sql/dsl-config-functions.sql
+  PGPASSWORD=$CS_DATABASE__PASSWORD psql $connection_url -f sql/dsl-encryptindex.sql
+
+  # tests
+  # PGPASSWORD=$CS_DATABASE__PASSWORD psql $connection_url -f tests/core.sql
+  PGPASSWORD=$CS_DATABASE__PASSWORD psql $connection_url -f tests/config.sql
+  # PGPASSWORD=$CS_DATABASE__PASSWORD psql $connection_url -f tests/encryptindex.sql
```
CDThomas, line 21: Why are the core and encryptindex tests commented out? Is that intentional?

CDThomas (approved):

justfile
fimac: oops, no, we just commnted them out to reduce noise. I'll revert

.gitignore
fimac: that's a good idea. I'll update to commit it.

I'll do the check in CI separately as want to get back to the poc search_doc testing.

PR #26, opened 2024-10-21
Update json payload query support

fimac:
Updates the payload to include a q value.

Adds query functions to use for each type to serialize the query into the required format.

Updates goeql tests.

Points eql sql install to central release folder that contains the latest sql for eql.

### DIFF
--- a/languages/go/xorm/e2e_test.go
+++ b/languages/go/xorm/e2e_test.go
@@ -51,10 +51,10 @@ func TestWhereQueryOnUnencryptedColumn(t *testing.T) {
 		t.Fatalf("Could not insert new example: %v", err)
 	}
 
-	var example Example
+	var returnedExample Example
 	text := "sydney"
 
-	has, err := engine.Where("non_encrypted_field = ?", text).Get(&example)
+	has, err := engine.Where("non_encrypted_field = ?", text).Get(&returnedExample)
 	if err != nil {
 		t.Fatalf("Could not retrieve example: %v", err)
 	}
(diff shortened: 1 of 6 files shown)


languages/go/xorm/e2e_test.go
```diff
@@ -63,10 +63,10 @@ func TestWhereQueryOnUnencryptedColumn(t *testing.T) {
 		t.Errorf("Expected has to equal true, got: %v", has)
 	}
 
-	assert.Equal(t, newExample.NonEncryptedField, example.NonEncryptedField, "NonEncryptedField does not match")
-	assert.Equal(t, newExample.EncryptedIntField, example.EncryptedIntField, "EncryptedIntField does not match")
-	assert.Equal(t, newExample.EncryptedTextField, example.EncryptedTextField, "EncryptedTextField does not match")
-	assert.Equal(t, newExample.EncryptedJsonbField, example.EncryptedJsonbField, "EncryptedJsonbField does not match")
+	assert.Equal(t, newExample.NonEncryptedField, returnedExample.NonEncryptedField, "NonEncryptedField does not match")
```
fimac, line 66: Updated these assertions as the expected value should be the first value in the assertion.

```
func assert.Equal(t assert.TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{}) bool
```

languages/go/xorm/e2e_test.go
```diff
@@ -82,6 +82,12 @@ func TestMatchQueryLongString(t *testing.T) {
 	}
 
 	examples := []Example{
+		{
+			NonEncryptedField:   "brisbane",
+			EncryptedIntField:   23,
+			EncryptedTextField:  "another string that shouldn't be returned",
+			EncryptedJsonbField: jsonData,
+		},
```
fimac, line 90: Added another example to insert to ensure failures are picked up

languages/go/xorm/e2e_test.go
```diff
@@ -119,7 +125,7 @@ func TestMatchQueryLongString(t *testing.T) {
 		t.Errorf("Expected has to equal true, got: %v", has)
 	}
 
-	assert.Equal(t, returnedExample.EncryptedTextField, EncryptedTextField("this is a long string"), "EncryptedTextField should match")
+	assert.Equal(t, EncryptedTextField("this is a long string"), returnedExample.EncryptedTextField, "EncryptedTextField should match")
```
fimac, line 128: same as above, expected should be first in the assertion

auxesis (approved):
Nice work @fimac, thanks for putting this together. 

I have added some commits to simplify and make the code and docs a little more idiomatic.

PR #36, opened 2024-10-23
Generate EQL uninstall 

tobyhede:
Build now generates:
 - cipherstash-encrypt-uninstall.sql
 - cipherstash-encrypt.sql

At no extra charge:

- add release files to `gitignore`
- makes sure every entity has a `drop` 
- various small tweaks


auxesis (commented):
One question I have about the `release/*.sql` files:

Should we be committing those artifacts to the repo?

The general practice is to not, but we seem to be doing so.

Normally a git repo should contain all the bits required to build artifacts, but the build system is responsible for storing and serving those artifacts.

fimac:
> One question I have about the `release/*.sql` files:
> 
> Should we be committing those artifacts to the repo?
> 
> The general practice is to not, but we seem to be doing so.
> 
> Normally a git repo should contain all the bits required to build artifacts, but the build system is responsible for storing and serving those artifacts.

I think it would be good to have it there, so we can point the orm examples to that file instead of copying it over into each example.

fimac:
@tobyhede it would be good to include the uninstall.sql for ore types as well. https://github.com/cipherstash/encrypt-query-language/blob/main/sql/database-extensions/postgresql/uninstall.sql

auxesis:
> I think it would be good to have it there, so we can point the orm examples to that file instead of copying it over into each example.

That's a fair point, it is convenient to be able to reference it inside the repo. 

That said, there is a quick URL to access [the latest EQL release](https://github.com/cipherstash/encrypt-query-language/releases/latest/download/cipherstash-eql.sql) from GitHub:

```
curl -sLo cipherstash-eql.sql https://github.com/cipherstash/encrypt-query-language/releases/latest/download/cipherstash-eql.sql
```

PR #46, opened 2024-10-28
Allow jsonb columns as encrypted targets

tobyhede:
Allows `jsonb` columns as well as the `cs_encrypted_v1` domain type.

Adds a force to `cs_encrypt_v1()`

Calling `cs_encrypt_v1(true)` will skip the schema check before entering the encrypting state.

### DIFF
--- a/sql/021-config-functions.sql
+++ b/sql/021-config-functions.sql
@@ -228,17 +228,24 @@ $$ LANGUAGE plpgsql;
 
 DROP FUNCTION IF EXISTS cs_encrypt_v1();
 
-CREATE FUNCTION cs_encrypt_v1()
+CREATE FUNCTION cs_encrypt_v1(force boolean DEFAULT false)
   RETURNS boolean
 AS $$
 	BEGIN
-    IF NOT cs_ready_for_encryption_v1() THEN
-      RAISE EXCEPTION 'Some pending columns do not have an encrypted target';
+
+    IF EXISTS (SELECT FROM cs_configuration_v1 c WHERE c.state = 'encrypting') THEN
+      RAISE EXCEPTION 'An encryption is already in progress';
     END IF;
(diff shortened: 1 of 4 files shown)


auxesis (changes requested):
Need some docs, please @tobyhede.

sql/README.md
```diff
@@ -0,0 +1,13 @@
+### Adding SQL
+
+
+- Never drop the configuration table as it may contain customer data and needs to live across EQL versions
+- Everything else should have a DROP IF EXISTS
+- Functions should be DROP and CREATE, instead of CREATE OR REPLACE
+ - Data types cannot be changed once created, so dropping first is more flexible
+- Keep DROP and CREATE together in the code
```
auxesis, line 8: Highlight technical terms

```suggestion
- Everything else should have a `DROP IF EXISTS`
- Functions should be `DROP` and `CREATE`, instead of `CREATE OR REPLACE`
 - Data types cannot be changed once created, so dropping first is more flexible
- Keep `DROP` and `CREATE` together in the code
```

sql/021-config-functions.sql
```diff
@@ -228,18 +228,25 @@ $$ LANGUAGE plpgsql;
 
 DROP FUNCTION IF EXISTS cs_encrypt_v1();
 
-CREATE FUNCTION cs_encrypt_v1()
+CREATE FUNCTION cs_encrypt_v1(force boolean DEFAULT false)
```
auxesis, line 231: Can we please get some documentation on what this force argument does, and how one might use it?

auxesis (changes requested):
Thanks @tobyhede. 

One small suggestion, and one request.

auxesis (approved):
Thanks for the changes @tobyhede.

PR #47, opened 2024-10-29
Update go xorm tests

fimac:
- updates proxy, and goeql
- adds additional tests for jsonb ops that we have in main currently and inline with poc
- deletes examples because they are a duplication of what we have in the e2e tests.
- updates to download eql from releases and install on init of docker
- removes dropping and recreating db when running tests  not necessary.
Screenshot 2024-10-30 at 12 42 08 PM


languages/go/xorm/run.sh
```diff
@@ -26,28 +26,9 @@ subproject_teardown() {
   docker compose down
 }
 
-subproject_examples() {
```
fimac, line 29: Removed this command as no longer necessary (see comment in description). The tests are the source of truth for examples. No need to duplicate.

languages/go/xorm/migrations.go
```diff
@@ -3,29 +3,8 @@ package main
...
 
-func InstallEql(engine *sql.DB) {
-	log.Println("Start installing custom types and function")
-	installDsl(engine)
-
-}
-
-// Installing EQL
-func installDsl(engine *sql.DB) {
-	path := "../../../release/cipherstash-encrypt-dsl.sql"
```
fimac, line 17: removed as this is now being installed on init of docker with a download directly from eql

languages/go/xorm/main.go
```diff
@@ -213,57 +175,15 @@ func main() {
...
-	setupDb()
-
 	// Connect to go test directly and create table
 	createTable()
 
-	// Install EQL and add config
-	installEql()
-}
-
-func runExamples() {
```
fimac, line 234: As noted in the description, these examples are duplicates of the e2e tests. Removing as there's no need for both, tests are the source of truth for working examples.

languages/go/xorm/main.go
```diff
@@ -123,44 +123,6 @@ func (eb *EncryptedBoolField) FromDB(data []byte) error {
 	return nil
 }
 
-func setupDb() {
```
fimac, line 126: Unnecessary to have this. The db is created when docker starts

auxesis (approved):
Big fan of these simplifications. 

Nice work @fimac ❤️

PR #51, opened 2024-10-30
Adds some cd-ing for safety in the justfile bash scripts

damncabbage:
… for cases where you might be using the `justfile` from outside the repo.

I happen to be calling `just -f ~/code/encrypt-query-language/justfile reset` for my benchmarking setup; I use `.envrc` for environment management, and actually `cd`-ing over there would mean I lose said environment setup.

(`justfile_directory()` has been part of Just since v0.94 or so, so it's safe to use.)

### DIFF
--- a/justfile
+++ b/justfile
@@ -5,6 +5,7 @@ set positional-arguments
 test:
   #!/usr/bin/env bash
   set -euxo pipefail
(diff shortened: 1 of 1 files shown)


auxesis (approved):
Love it, thanks @damncabbage.

PR #52, opened 2024-10-30
Add test coverage for go zero value reads writes

fimac:
When inserting a jsonb value as nil, or an empty map, or not including the field on insert, the value coming through to todb is an empty map[]. This is because go sets these values to the go zero value https://go.dev/ref/spec#The_zero_value.

This PR will add a change to check for this in the EncryptedJsonb type in goeql so that we don't serialize the value into the  jsonb shape. The proxy will encrypt the value as "null", and will encrypt this. But, when trying to read, this error will return 
`handle] rewrite_results Error: Unknown variant code `11``. 

I have been testing by pointing to this branch locally https://github.com/cipherstash/goeql/pull/3

~~I will do a release of goeql and then open this PR.~~
Bumped goeql and ran tests:

Screenshot 2024-10-30 at 5 01 47 PM

### DIFF
--- a/languages/go/xorm/e2e_test.go
+++ b/languages/go/xorm/e2e_test.go
@@ -551,7 +551,170 @@ func TestJsonbTermsOp(t *testing.T) {
 		t.Fatalf("Could not unmarshal %v", err)
 	}
 	assert.Equal(t, expected_id, jsonData)
+}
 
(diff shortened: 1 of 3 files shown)


auxesis (approved):
Nice work @fimac.

PR #53, opened 2024-10-30
Granular errors on encrypted column constraint

tobyhede:
Extract constraint checks into functions with clear error messages

### DIFF
--- a/sql/010-core.sql
+++ b/sql/010-core.sql
@@ -106,7 +106,7 @@ AS $$
   END;
 $$ LANGUAGE plpgsql;
 
--- Should include an ident field
+-- Should include a version field
 DROP FUNCTION IF EXISTS _cs_encrypted_check_v(jsonb);
 CREATE FUNCTION _cs_encrypted_check_v(val jsonb)
   RETURNS boolean
--- a/sql/010-core.sql
+++ b/sql/010-core.sql
@@ -14,17 +14,88 @@ BEGIN
 END
 $$;
 
-DROP FUNCTION IF EXISTS _cs_encrypted_check_kind(jsonb);
+--
+-- CT payload should include a c field
+--
+DROP FUNCTION IF EXISTS _cs_encrypted_check_k_ct(jsonb);
+CREATE FUNCTION _cs_encrypted_check_k_ct(val jsonb)
+  RETURNS boolean
+AS $$
+	BEGIN
+    IF (val->>'k' = 'ct' AND val ? 'c') THEN
+      RETURN true;
+    END IF;
+    RAISE 'Encrypted kind (k) of "ct" missing data field (c):  %', val;
+  END;
+$$ LANGUAGE plpgsql;
 
-CREATE FUNCTION _cs_encrypted_check_kind(val jsonb)
-  RETURNS BOOLEAN
-LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE
-BEGIN ATOMIC
-  RETURN (
-    (val->>'k' = 'ct' AND val ? 'c') OR
-    (val->>'k' = 'sv' AND val ? 'sv')
-  ) AND NOT val ? 'p';
-END;
+--
+-- SV payload should include an sv field
+--
+DROP FUNCTION IF EXISTS _cs_encrypted_check_k_sv(jsonb);
+CREATE FUNCTION _cs_encrypted_check_k_sv(val jsonb)
+  RETURNS boolean
+AS $$
+	BEGIN
+    IF (val->>'k' = 'sv' AND val ? 'sv') THEN
+      RETURN true;
+    END IF;
+    RAISE 'Encrypted kind (k) of "sv" missing data field (sv):  %', val;
+  END;
+$$ LANGUAGE plpgsql;
+
+
+-- Plaintext field should never be present in an encrypted column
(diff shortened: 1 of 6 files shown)


sql/020-config-schema.sql
```diff
@@ -44,28 +44,66 @@ DROP FUNCTION IF EXISTS _cs_config_check_indexes(jsonb);
...
+
+--
+-- Should include a tables field
+-- Tables should not be empty
+DROP FUNCTION IF EXISTS _cs_config_check_tables(jsonb);
+CREATE FUNCTION _cs_config_check_tables(val jsonb)
+  RETURNS boolean
+AS $$
+	BEGIN
+    IF (val ? 'tables') AND (val->'tables' <> '{"A":"a"}'::jsonb) THEN
```
CDThomas, line 80: What's the reason for comparing `val->'tables'` to `'{"A":"a"}::jsonb` here?

sql/021-config-functions.sql
```diff
@@ -301,7 +301,7 @@ $$ LANGUAGE plpgsql;
 
 DROP FUNCTION IF EXISTS cs_add_column_v1(table_name text, column_name text);
 
-CREATE FUNCTION cs_add_column_v1(table_name text, column_name text)
+CREATE FUNCTION cs_add_column_v1(table_name text, column_name text, cast_as text DEFAULT 'text')
```
CDThomas, line 304: I think the `DROP` statement also needs to be updated since a new arg was added here.

sql/010-core.sql
```diff
@@ -14,17 +14,94 @@ BEGIN
...
+-- Plaintext field should never be present in an encrypted column
+DROP FUNCTION IF EXISTS _cs_encrypted_check_p(jsonb);
+CREATE FUNCTION _cs_encrypted_check_p(val jsonb)
+  RETURNS boolean
+AS $$
+	BEGIN
+    IF NOT val ? 'p' THEN
+      RETURN true;
+    END IF;
+    RAISE 'Encrypted includes plaintext (p) field: %', val;
```
CDThomas, line 63: Feels like a word is missing here between "Encrypted" and "includes". Might be worth rephrasing.

sql/010-core.sql
```diff
@@ -14,17 +14,94 @@ BEGIN
...
+-- Should include an ident field
+DROP FUNCTION IF EXISTS _cs_encrypted_check_i(jsonb);
+CREATE FUNCTION _cs_encrypted_check_i(val jsonb)
+  RETURNS boolean
+AS $$
+	BEGIN
+    IF val ? 'i' THEN
+      RETURN true;
+    END IF;
+    RAISE 'Encrypted missing ident (i) field: %', val;
```
CDThomas, line 76: Similar comment here about missing a word here. Might be worth rephrasing.

sql/010-core.sql
```diff
@@ -34,14 +111,11 @@ CREATE FUNCTION cs_check_encrypted_v1(val jsonb)
...
-
-      -- table and column
-      val->'i' ?& array['t', 'c'] AND
-
-      -- plaintext or ciphertext for kind
-      _cs_encrypted_check_kind(val)
+      _cs_encrypted_check_v(val) AND
+      _cs_encrypted_check_i(val) AND
+      _cs_encrypted_check_k_ct(val) AND
+      _cs_encrypted_check_k_sv(val) AND
```
CDThomas, line 117: Would it be worth adding a check that the `"k"` (kind) key is present and that it's either `"ct"` (ciphertext) or `"sv"` (ste_vec)?

CDThomas (commented):
Nice. Left a few comments, but I think this will definitely help with debugging 🙏

sql/020-config-schema.sql
tobyhede: iamverysmart

sql/021-config-functions.sql
tobyhede: Good catch, thanks

sql/010-core.sql
```diff
@@ -14,17 +14,110 @@ BEGIN
...
+AS $$
+	BEGIN
+    IF val ? 'i' THEN
+      RETURN true;
+    END IF;
+    RAISE 'Encrypted column missing ident (i) field: %', val;
+  END;
+$$ LANGUAGE plpgsql;
+
+-- Should include an ident field
```
CDThomas, line 96:
```suggestion
-- Ident field should include table and column
```

sql/010-core.sql
```diff
@@ -14,17 +14,110 @@ BEGIN
...
+AS $$
+	BEGIN
+    IF (val->'i' ?& array['t', 'c']) THEN
+      RETURN true;
+    END IF;
+    RAISE 'Encrypted column ident (i) missing table (t) or column (c) fields: %', val;
+  END;
+$$ LANGUAGE plpgsql;
+
+-- Should include an ident field
```
CDThomas, line 109:
```suggestion
-- Should include a version field
```

tests/core.sql
```diff
@@ -93,6 +93,35 @@ INSERT INTO users (name_encrypted) VALUES (
   }'::jsonb
 );
 
+-- npo kind
```
CDThomas, line 96:
```suggestion
-- no kind
```

CDThomas (approved):

PR #55, opened 2024-10-31
Add support for `GROUP BY` with `cs_grouped_value_v1`

CDThomas:
This change adds support for decrypting encrypted columns included in the `GROUP BY` clause by using a new aggregate function, `cs_grouped_value_v1`.

How it works:
- An ORE term (either in an ORE or SteVec index) can be used in `GROUP BY`
- `cs_grouped_value_v1` can be used in the `SELECT` target list to return a single source ciphertext for each group
- the ORE index is used for the equality check needed for grouping
- `cs_grouped_value_v1` allows for using the source ciphertext in the `SELECT` target list (since `SELECT`ed items must either use an aggregrate function or appear in the `GROUP BY`)
- note that including the source ciphertext in the `GROUP BY` statement wouldn't work because source ciphertexts are non-deterministic and don't support comparison operations (that's what ORE indexes are for)

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,16 @@ And is the EQL equivalent of the following plaintext query.
 SELECT id FROM examples ORDER BY field DESC;
 ```
 
+**Grouping example:**
+
+ORE indexes can be used along with the `cs_grouped_value_v1` aggregate function to group by an encrypted column:
+
+```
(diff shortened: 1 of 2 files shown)


freshtonic (approved):
Nice implementation. More explicit than the `min` / `max` hack we used to have.

fimac (approved):

PR #59, opened 2024-11-04
Raise exceptions if index fields are missing

tobyhede:
Index helper functions check for presence of the appropriate index field. 
If not present, an exception is raised.

Also include some minor changes to align with naming conventions.


freshtonic (approved):
Nicely done

PR #67, opened 2024-11-14
Update pgx to latest

fimac:
Updates pgx to the latest version https://pkg.go.dev/github.com/jackc/pgx/v5

Fixes:
https://github.com/cipherstash/encrypt-query-language/security/dependabot/10
https://github.com/cipherstash/encrypt-query-language/security/dependabot/9


cisopaul (approved):
Looks good to me

PR #79, opened 2025-01-05
Ensure EQL works with operators

tobyhede:
If operators work, values don't need to be wrapped with EQL functions to map the data.


auxesis (approved):
LGTM (although I did the build engineering work, which makes me biased)

PR #83, opened 2025-01-22
ORE Operators

tobyhede:
Add ORE operators for the `cs_encrypted_v1` type
Also generalises equality operators to support both unique and ore indexes.

Addresses: 
https://linear.app/cipherstash/issue/CIP-1148/implement-ore-operators-for-cs-encrypted-v1
https://linear.app/cipherstash/issue/CIP-1149/equality-operator-needs-to-handle-both-unique-and-ore-index-types


sql/015-operators-eq.sql
```diff
@@ -0,0 +1,525 @@
...
+CREATE FUNCTION cs_encrypted_eq_v1(a cs_encrypted_v1, b cs_encrypted_v1)
+  RETURNS boolean
+  IMMUTABLE STRICT PARALLEL SAFE
+AS $$
+  DECLARE
+    u boolean;
+    o boolean;
+  BEGIN
+    BEGIN
+      u := (SELECT cs_unique_v1(a) = cs_unique_v1(b));
```
coderdan, line 25: I'm not 100% what the threat model might be but its worth noting that this comparison won't run in constant time. It _might_ leave the possibility for an attacker to learn the value of the plaintext in certain circumstances.

There's a constant time comparison function in the EQL repo somewhere I think.

sql/015-operators-eq.sql
```diff
@@ -0,0 +1,525 @@
...
+    EXCEPTION WHEN OTHERS THEN
+      u := false;
+    END;
+
+    BEGIN
+      o := (SELECT cs_ore_64_8_v1(a) = cs_ore_64_8_v1(b));
```
coderdan, line 31: Same for this one.

sql/015-operators-eq.sql
coderdan: Something like this: https://github.com/cipherstash/encrypt-query-language/blob/2ee50af4abcc68e5346baa89687cf893e67daf89/sql/001-ore-cllw.sql#L21

tobyhede: `cs_unique_v1` is the existing function, and simply extracts the index value from the `eql::ciphertext`

sql/666-drop-operators.sql
```diff
@@ -0,0 +1,57 @@
...
+DROP OPERATOR IF EXISTS <> (cs_encrypted_v1, cs_encrypted_v1);
+
+DROP OPERATOR IF EXISTS <> (cs_encrypted_v1, jsonb);
+DROP OPERATOR IF EXISTS <> (jsonb, cs_encrypted_v1);
+
+DROP OPERATOR IF EXISTS <> (cs_encrypted_v1, cs_unique_index_v1);
+DROP OPERATOR IF EXISTS <> (cs_unique_index_v1, cs_encrypted_v1);
+
+DROP OPERATOR IF EXISTS <> (ore_64_8_v1, cs_encrypted_v1);
+DROP OPERATOR IF EXISTS <> (cs_encrypted_v1, ore_64_8_v1);
```
coderdan, line 57: Note: I haven't checked that you've captured all the operators here. I'm assuming running the script will show if that's not the case!

sql/015-operators-eq.sql
tobyhede: `cs_ore_64_8_v1` is same as unique, pulls the field from the `json`
The returned data is cast as `ore_64_8_v1` and then the underlying ore operators are used for the comparison

sql/015-operators-eq.sql
tobyhede: Unique indexes are cast as `text`, so this ends up using the default text eq operators.

tests/operators-match.sql
```diff
@@ -0,0 +1,90 @@
+\set ON_ERROR_STOP on
```
coderdan, line 1: TIL.

tests/operators-match.sql
```diff
@@ -0,0 +1,90 @@
...
+
+  END;
+$$ LANGUAGE plpgsql;
+
+
+
+-- MATCH <@ OPERATORS
+DO $$
+  BEGIN
+    -- SANITY CHECK FOR UNIQUE payloads
```
coderdan, line 65: I don't understand the comment. Is that a copy-paste error or am I missing something about what this test does?

tests/operators-ore.sql
```diff
@@ -0,0 +1,288 @@
...
+
+    ore_record = '("{""(\\""\\\\\\\\x1212121212125932e28282d03415e7714fccd69eb7eb476c70743e485e20331f59cbc1c848dcdeda716f351eb20588c406a7df5fb8917ebf816739aa1414ac3b8498e493bf0badea5c9fdb3cc34da8b152b995957591880c523beb1d3f12487c38d18f62dd26209a727674e5a5fe3a3e3037860839afd8011f94b49eaa5fa5a60e1e2adccde4185a7d6c7f83088500b677f897d4ffc276016d614708488f407c01bd3ccf2be653269062cb97f8945a621d049277d19b1c248611f25d047038928d2efeb4323c402af4c19288c7b36911dc06639af5bb34367519b66c1f525bbd3828c12067c9c579aeeb4fb3ae0918125dc1dad5fd518019a5ae67894ce1a7f7bed1a591ba8edda2fdf4cd403761fd981fb1ea5eb0bf806f919350ee60cac16d0a39a491a4d79301781f95ea3870aea82e9946053537360b2fb415b18b61aed0af81d461ad6b923f10c0df79daddc4e279ff543a282bb3a37f9fa03238348b3dac51a453b04bced1f5bd318ddd829bdfe5f37abdbeda730e21441b818302f3c5c2c4d5657accfca4c53d7a80eb3db43946d38965be5f796b\\"")""}")';
+
+
+    -- SANITY CHECK
+    ASSERT (SELECT EXISTS (SELECT id FROM users WHERE cs_ore_64_8_v1(name_encrypted) < cs_ore_64_8_v1(ore_json)));
+
+    ASSERT (SELECT EXISTS (
+      SELECT id FROM users WHERE name_encrypted < ore_cs_encrypted::jsonb
+    ));
```
coderdan, line 62: These tests are fine but in my experience ORE tests with hard-coded values can give false confidence. Because an incorrect implementation will still "work" ~50% of the time. Not for this PR but we may want to think about how to generate random ORE encrypted values for every test (like a property test) to validate correctness more robustly.

sql/015-operators-eq.sql
coderdan: Yeah that's what I figured. The standard equality operator will bail as soon as a character differs which can lead to timing side-channel attacks. A constant time comparison should always take the same amount of time regardless of which character index differs first.

coderdan (approved):
EPIC PR. Nice work.

A couple of suggestions for future improvements.

sql/015-operators-eq.sql
tobyhede: Oh right. We should fix that for unique in general then

PR #86, opened 2025-01-30
ORE Operators and Sorting

tobyhede:
Make ORDER by work for `cs_encrypted_v1` by encoding in the EQL `json` as an array rather than the PostgreSQL Ore Types.

Example ORE encoding
```
'{"o": ["121212121212594be28282d03415e7714fccd69eb7eb476c70743e485e20331f59cbc1c848dcdeda716f351eb20588c406a7df5fb8917ebf816739aa1414ac3b8498e493bf0badea5c9fdb3cc34da8b152b995957591880c523beb1d3f12487c38d18f62dd26209a727674e5a5fe3a3e3037860839afd80132248f0640e89761a123fad8155748d764f347a29e059758575a770618ab6f82d06bad973c3fb62505d9749f4f8483c8d607c61bae7c75ef09add6d91b728449726534e65379f7b3442d2a4aa2b8c3cdb90311b53dc333bbf6b213949a8990b4300473985f60c09c6a91ac963c802e319c28bafc2be66eceb3f1924081724e44d173de2091251d1ea69ec827b94ea5ab63436f0701dd2bf299e1a66a22c4b44b32b88620949736e088bc3ec6e7974426e4b392ecece0e88a7acaf510322d1726da6bc9580dad3c8717619051c220d8654a35eb7fa0a6de4be0456522054f124bbb0bdda4bc177b35a6ca20bd996f3a3499ffd00c93d4705cc4bc05f428541c3adcc36f0b9b9aebc61a88cd4bad8f034dd4a483de9bd3291e4bee06449083c83e"]}';
```

Requires a corresponding change in the proxy to encode the ore index value as an array:
https://github.com/cipherstash/proxy/pull/110

Current status: ORE works here, but does not work via proxy and I am not sure why.


sql/017-operators-ore.sql
```diff
@@ -401,24 +401,31 @@ CREATE OPERATOR <=(
...
   RETURNS integer AS $$
+   DECLARE
+    a_ore bytea[];
+    b_ore bytea[];
   BEGIN
-    RETURN compare_ore_64_8_v1(cs_ore_64_8_v1(a), cs_ore_64_8_v1(b));
+    a_ore = ARRAY[(a->>'o')::bytea];
+    b_ore = ARRAY[(b->>'o')::bytea];
+
+    RETURN compare_ore_64_8_v1(a_ore, b_ore);
```
tobyhede, line 415: Explicitly using `bytea[]`, but need new function definition for these types too.

sql/000-ore.sql
```diff
@@ -268,7 +284,19 @@ RETURNS integer AS $$
...
+    RETURN compare_ore_array(a.terms::bytea[], b.terms::bytea[]);
+  END
+$$ LANGUAGE plpgsql;
+
+DROP FUNCTION IF EXISTS compare_ore_64_8_v1(a bytea[], b bytea[]);
+
+CREATE FUNCTION compare_ore_64_8_v1(a bytea[], b bytea[])
+RETURNS integer AS $$
+  DECLARE
+    cmp_result integer;
```
coderdan, line 296: This looks to be unused now

sql/000-ore.sql
```diff
@@ -268,7 +284,19 @@ RETURNS integer AS $$
...
+  BEGIN
+    -- Recursively compare blocks bailing as soon as we can make a decision
+    RETURN compare_ore_array(a, b);
```
coderdan, line 299: Bailing early can lead to timing attacks. I suspect that's why there is a variable (now unused) to keep the result while the rest of the traversal takes place.

coderdan (commented):
I like the simplicity of just using bytea (especially if it actually works!).

These changes introduce a potential side channel vulnerability though (timing attack) because the comparison now bails early.

sql/000-ore.sql
```diff
@@ -87,169 +81,40 @@ CREATE FUNCTION compare_ore_64_8_v1_term(a ore_64_8_v1_term, b ore_64_8_v1_term)
...
+CREATE FUNCTION compare_ore_array(a bytea[], b bytea[])
 RETURNS integer AS $$
   DECLARE
     cmp_result integer;
+    len_a integer;
+    len_b integer;
   BEGIN
+
+    SELECT array_length(a, 1) INTO len_a;
+    SELECT array_length(b, 1) INTO len_b;
```
coderdan, line 102: It doesn't look like `len_a` and `len_b` are used. Maybe remove them or change the code to use them.

sql/000-ore.sql
```diff
@@ -87,169 +81,40 @@ CREATE FUNCTION compare_ore_64_8_v1_term(a ore_64_8_v1_term, b ore_64_8_v1_term)
...
 -- so if a is empty we return -1, if be is empty and a isn't, we return 1.
 -- If both are empty we return 0. This cases probably isn't necessary as equality
 -- doesn't always make sense but it's here for completeness.
 -- If both are non-empty, we compare the first element. If they are equal
 -- we need to consider the next block so we recurse, otherwise we return the comparison result.
-DROP FUNCTION IF EXISTS compare_ore_array(a ore_64_8_v1_term[], b ore_64_8_v1_term[]);
 
-CREATE FUNCTION compare_ore_array(a ore_64_8_v1_term[], b ore_64_8_v1_term[])
+
+CREATE FUNCTION compare_ore_array(a bytea[], b bytea[])
```
coderdan, line 93: Any reason not to use the `ore_64_8_index_v1` domain for the arg types here?

sql/666-drop_types.sql
```diff
@@ -1,6 +1,6 @@
 -- ANYTHING THAT NEEDS TO BE DROPPED LAST
 DROP TYPE IF EXISTS ore_64_8_v1;
-DROP TYPE IF EXISTS ore_64_8_v1_term;
+-- DROP TYPE IF EXISTS ore_64_8_v1_term;
```
coderdan, line 3: Remove this line now?

coderdan (approved):
Nice work. I bet you're glad to get this one over the line!

sql/000-ore.sql
tobyhede: Nope

PR #87, opened 2025-02-20
Use same drop setup on install and uninstall

tobyhede:
Generate a single sql file containing all of the drops and use this for both install and uninstall.


sql/000-ore.sql
```diff
@@ -184,8 +184,8 @@ DROP OPERATOR IF EXISTS = (ore_64_8_index_v1, ore_64_8_index_v1);
 
 CREATE OPERATOR = (
   PROCEDURE="ore_64_8_v1_eq",
-  LEFTARG=ore_64_8_v1,
-  RIGHTARG=ore_64_8_v1,
+  LEFTARG=ore_64_8_index_v1,
```
tobyhede, line 187: The drop failure was hiding that the type name had changed

auxesis (approved):
LGTM, thanks @tobyhede.

PR #89, opened 2025-02-25
Revert back to commit 01dcc24

CDThomas:
This change reverts this repo back to 01dcc24.

The changes reverted include commits from:
- https://github.com/cipherstash/encrypt-query-language/pull/86
- https://github.com/cipherstash/encrypt-query-language/pull/87

We mostly want to revert the changes in #86 (since they aren't working as intended with Proxy), but #87 is also included since it's more recent (and also includes some ORE-related changes that would be tedious to untangle).

Since there aren't many changes after #86, the most pragmatic option is to revert to the last known-good state and redo the install/uninstall changes by hand on top of that.

Commands used to revert:
```
git reset --hard 01dcc247e4c2b527a78d3ed5277e1464141beca2
git reset --soft ed460fc653604207210e7cc6029e1292441d3bbb
```

This change doesn't use `git revert` because there were > 20 commits to revert and merge commits also don't play well with `git revert`.


CDThomas:
Next steps after this PR:
- Raise another PR in this repo to use the new format for ORE indexes from Proxy (JSON array of hex-encoded strings), but decode into the composite type used for ORE in PG. Test Proxy against to verify that `ORDER BY` and comparison operators work as expected.
- Raise another PR to redo the work in #87

tobyhede (approved):

PR #90, opened 2025-02-25
Parse ORE indexes as JSON arrays of hex-encoded strings

CDThomas:
This change updates `cs_ore_64_8_v1` to parse ORE indexes (the `'o'` field) as JSON arrays of hex-encoded strings (instead of casting from the Postgres text format).

The corresponding change for encoding ORE indexes as JSON arrays of hex-encoded stings has already been merged in Proxy.

This is similar to some of the changes in
https://github.com/cipherstash/encrypt-query-language/pull/86, but we're parsing into the composite types for ORE indexes instead of into a plain `bytea[]`. Parsing into the composite type allows for ordering with an operator class on the output from `cs_ore_64_8_v1`.


sql/011-core-functions.sql
```diff
@@ -156,6 +156,24 @@ DROP CAST IF EXISTS (text AS ore_64_8_v1_term);
...
 	WITH FUNCTION _cs_text_to_ore_64_8_v1_term_v1_0(text) AS IMPLICIT;
 
+DROP FUNCTION IF EXISTS jsonb_array_to_ore_64_8_v1(val jsonb);
+
+CREATE FUNCTION jsonb_array_to_ore_64_8_v1(val jsonb)
+RETURNS ore_64_8_v1 AS $$
+DECLARE
+  terms_arr ore_64_8_v1_term[];
+BEGIN
+  IF jsonb_typeof(val) = 'null' THEN
```
tobyhede, line 166: TIL

tobyhede (approved):
I like it. 
Preserves the array semantics, which is mure more portable. 
Might be worth a comment explaining `jsonb_array_to_ore_64_8_v1` as I am sure I will forget what that does in about 30s.

PR #92, opened 2025-03-06
fix(ore): NULL comparisons should return NULL

tobyhede:


sql/000-ore.sql
```diff
@@ -239,13 +239,27 @@ RETURNS integer AS $$
   DECLARE
     cmp_result integer;
   BEGIN
-    IF (array_length(a, 1) = 0 OR a IS NULL) AND (array_length(b, 1) = 0 OR b IS NULL) THEN
+    IF a IS NULL AND b IS NULL THEN
+      RETURN NULL;
+    END IF;
```
yujiyokoo, line 244: Hmm, could this be `IF a IS NULL OR b IS NULL`, and get rid of `IF a IS...` and `IF b IS...` below?

yujiyokoo (approved):
Looks good

sql/000-ore.sql
tobyhede: Yes. Lol. Thank you

PR #93, opened 2025-03-07
Revert "fix(ore): NULL comparisons should return NULL"

tobyhede:
Reverts cipherstash/encrypt-query-language#92


auxesis (approved):
Approved. 

Highly likely that this is causing the build failures cipherstash/proxy#148 is trying to fix.

PR #96, opened 2025-03-20
fix: Use ~~ as match operator

tobyhede:
Using `~~` and `~~*` as match operators enable `LIKE` and `ILIKE` at no extra charge.

### DIFF
--- a/sql/016-operators-match.sql
+++ b/sql/016-operators-match.sql
@@ -2,21 +2,31 @@
 --
 -- Support for the following comparisons:
 --
---      cs_encrypted_v1 @> cs_encrypted_v1
---      cs_encrypted_v1 @> jsonb
---      cs_encrypted_v1 @> cs_match_index_v1
+--      cs_encrypted_v1 ~~ cs_encrypted_v1
+--      cs_encrypted_v1 ~~ jsonb
+--      cs_encrypted_v1 ~~ cs_match_index_v1
 --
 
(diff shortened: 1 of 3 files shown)


sql/016-operators-match.sql
```diff
@@ -2,21 +2,33 @@
...
---      cs_encrypted_v1 @> jsonb
---      cs_encrypted_v1 @> cs_match_index_v1
+--      cs_encrypted_v1 ~~ cs_encrypted_v1
+--      cs_encrypted_v1 ~~ jsonb
+--      cs_encrypted_v1 ~~ cs_match_index_v1
 --
 
-DROP OPERATOR IF EXISTS @> (cs_encrypted_v1, cs_encrypted_v1);
-DROP FUNCTION IF EXISTS cs_encrypted_contains_v1(a cs_encrypted_v1, b cs_encrypted_v1);
+-- The operator ~~ is equivalent to LIKE, and ~~*
```
freshtonic, line 10: Is `~~*` equivalent to `ILIKE`? If so you've accidentally a word

freshtonic (approved):
I think you might have missed out a word on a code comment - see my comment.

Otherwise all good!

You'll need to bump the EQL release again once merged - and update proxy and possibly protectjs too.

CDThomas (approved):
Nice, glad that using `~~`/`~~*` allows us to skip mapping for `LIKE`/`ILIKE` support.

PR #97, opened 2025-04-02
Create cs min v1 cs max v1 eql functions

yujiyokoo:
This adds `cs_min_v1` and `cs_max_v1` for ORE


freshtonic (approved):
This looks really good. I think the way you've done the NULL handling makers sense too - same semantics as normal SQL for behaviour of MIN & MAX when given NULL as input

PR #98, opened 2025-04-11
Wrap EQL into a PostgreSQL schema

tobyhede:
Touches every. single. sql. file. 

Apologies in advance.


yujiyokoo (approved):
Seems like straightforward renaming (please tell me if not)

PR #99, opened 2025-04-14
Modularise SQL

tobyhede:
Work in progress.

The story so far:

 - [x] improved dependency management for sql files
 - [x] smaller modules eg all ore types are now in `src/ore/types.sql`
 - [x] ore data for better testing
 - [x] additional test helpers to clean up testing
 - [x] smarter test runner that can run single test file  `mise run test --test encrypted/functions`


auxesis (approved):
Thanks @tobyhede!

PR #103, opened 2025-05-08
Enhance functions for operator-free environments like Supabase

tobyhede:
EQL documentation for working with [Supabase without operators](https://github.com/supabase/supautils/issues/72).

Couple of small quality of life changes to functions that emerged while documenting:

- Adds an `order_by` function that wraps the ore index term function for better DX.
- Rename `match` to `like`, to reflect the actual PostgreSQL usage.

### DIFF
--- a/SUPABASE.md
+++ b/SUPABASE.md
@@ -1,0 +1,70 @@
+# Supabase
+
+
+## No operators, no problems
+
+Supabase does not support custom operators.
+The EQL operator functions can be used in this situation.
+
+PostgreSQL operators are essentially an alias for the equivalent function, so the implementation and behaviour remains the same.
+
+| Operator | Function                                           | Example                                                           |
+| -------- | -------------------------------------------------- | ----------------------------------------------------------------- |
+| `=`      | `eql_v1.eq(eql_v1_encrypted, eql_v1_encrypted)`    | `SELECT * FROM users WHERE eql_v1.eq(encrypted_email, $1)`<br>    |
+| `<>`     | `eql_v1.neq(eql_v1_encrypted, eql_v1_encrypted)`   | `SELECT * FROM users WHERE eql_v1.neq(encrypted_email, $1)`<br>   |
+| `<`      | `eql_v1.lt(eql_v1_encrypted, eql_v1_encrypted)`    | `SELECT * FROM users WHERE eql_v1.lt(encrypted_email, $1)`<br>    |
+| `<=`     | `eql_v1.lte(eql_v1_encrypted, eql_v1_encrypted)`   | `SELECT * FROM users WHERE eql_v1.lte(encrypted_email, $1)`<br>   |
+| `>`      | `eql_v1.gt(eql_v1_encrypted, eql_v1_encrypted)`    | `SELECT * FROM users WHERE eql_v1.gt(encrypted_email, $1)`<br>    |
+| `>=`     | `eql_v1.gte(eql_v1_encrypted, eql_v1_encrypted)`   | `SELECT * FROM users WHERE eql_v1.gte(encrypted_email, $1)`<br>   |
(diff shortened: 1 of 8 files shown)


freshtonic (approved):
Love it, nice work

SUPABASE.md
```diff
@@ -0,0 +1,70 @@
+# Supabase
+
+
+## No operators, no problems
+
+Supabase does not support custom operators.
```
auxesis, line 6: Link to GitHub issue tracking support for custom operators.

```suggestion
Supabase [does not currently support](https://github.com/supabase/supautils/issues/72) custom operators.
```

SUPABASE.md
```diff
@@ -0,0 +1,70 @@
...
+The EQL operator functions can be used in this situation.
+
+PostgreSQL operators are essentially an alias for the equivalent function, so the implementation and behaviour remains the same.
```
auxesis, line 9: Be more specific about how EQL is structured 

```suggestion
In EQL, PostgreSQL operators are an alias for a function, so the implementation and behaviour remains the same across operators and functions.
```

SUPABASE.md
```diff
@@ -0,0 +1,70 @@
...
+**Function**
+```
+SELECT * FROM users WHERE eql_v1.like(encrypted_email, $1)
+```
+
+#### Case Sensitivity
+
+The EQL `eql_v1.like` and `eql_v1.ilike` functions are equivalent.
+
+The behaviour of the "match" index term that is used by the encrypted `LIKE` operators is slightly different to default PostgreSQL. Case sensitivity is determined by the index term configuration. A match index term can be configured to enable case sensitive searches. The data is encrypted based on the configuration. The like operation is always the same, and the data is different. The different operators are kept to preserve the semantics of SQL statements in client applications.
```
auxesis, line 57: Line per sentence, to improve maintainability.

Highlight technical terms.

Link to supporting documentation.

```suggestion
The behaviour of EQL's encrypted `LIKE` operators is slightly different to the behaviour of PostgreSQL's `LIKE` operator.
In EQL, the `LIKE` operator can be used on `match` indexes.
Case sensitivity is determined by the [index term configuration](./docs/reference/INDEX.md#options-for-match-indexes-opts) of `match` indexes.
A `match` index term can be configured to enable case sensitive searches with token filters (for example, `downcase` and `upcase`). 
The data is encrypted based on the index term configuration. 
The `LIKE` operation is always the same, even if the data is tokenised differently.
The different operators are kept to preserve the semantics of SQL statements in client applications.
```

SUPABASE.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+
+### Order By
```
auxesis, line 61: Highlight technical terms

```suggestion
### `ORDER BY`
```

SUPABASE.md
```diff
@@ -0,0 +1,70 @@
...
+
+Ordering requires wrapping the ordered column in the `eql_v1.order_by` function.
```
auxesis, line 63: Make it clearer that the following example demonstrates how to wrap the ordered column

```suggestion
Ordering requires wrapping the ordered column in the `eql_v1.order_by` function, like this:
```

auxesis (approved):
Looking great @tobyhede. 

I have added a bunch of suggestions, and I'll push some commits to this branch with additional minor fixups.

SUPABASE.md
auxesis: Implemented in 60b848e93aadd5849e412f8c3edd57cb3e18967a, 26c73770702b64dc605504e78a80eeb0db4297b0, 4c4a0b17c5bbd622482f4bfad35efc0dc492d67b, 439f1cb6a062cb0c3155d00ebc0145ecb96c6b98.

calvinbrewer (approved):

PR #108, opened 2025-05-22
Remove unused files, examples and playground

tobyhede:
Removes
 - old uninstall.sql
 - go javascript and python examples
 - playground


freshtonic (approved):
-3,967 lines - nice!

PR #120, opened 2025-07-08
jsonb accessors return with metadata

tobyhede:
Column meta data is required for encryption.

This pulls metadata from the column (the `i/identifier` field with table and column names) and merges with the returned stevec element.

### DIFF
--- a/src/encrypted/functions.sql
+++ b/src/encrypted/functions.sql
@@ -85,3 +85,13 @@ AS $$
   END;
 $$ LANGUAGE plpgsql;
 
+
+CREATE FUNCTION eql_v2.meta_data(val eql_v2_encrypted)
+  RETURNS jsonb
+  IMMUTABLE STRICT PARALLEL SAFE
+AS $$
+  BEGIN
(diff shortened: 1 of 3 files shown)


src/encrypted/functions.sql
```diff
@@ -85,3 +85,13 @@ AS $$
...
 $$ LANGUAGE plpgsql;
 
+
+CREATE FUNCTION eql_v2.meta_data(val eql_v2_encrypted)
+  RETURNS jsonb
+  IMMUTABLE STRICT PARALLEL SAFE
+AS $$
+	BEGIN
+     RETURN eql_v2.meta_data(val.data);
+  END;
```
auxesis, line 95: Is this indentation right?

auxesis (approved):
LGTM, thanks @tobyhede!

src/encrypted/functions.sql
tobyhede: Weird. Looked fine in editor. Fixed maybe

PR #139, opened 2025-10-27
test(sqlx): add comprehensive JSONB operator tests

tobyhede:
Add Rust/SQLx tests covering JSONB functionality in encrypted payloads.

## JSONB Array Tests
- `jsonb_array_elements_text` - array element extraction
- `jsonb_array_length` - array size validation

## JSONB Path Query Tests
- `jsonb_path_query` - complex path expressions
- `jsonb_path_query_first` - single result queries
- `jsonb_path_exists` - path existence checks
- Array-specific path operations

## JSONB Structure Tests
- Encrypted selector validation (ct, k, i, p, ob fields)
- JSONB field type verification
- Payload structure correctness

## Test Helpers
- Function call tracking verification
- Test framework meta-tests

## Migration Progress

**Progress: 24/40 SQL assertions ported**

- ✅ JSONB functions: 24/24 (arrays, paths, structure validation, encrypted selectors)
- ⏳ Equality operators: 0/16 (pending)

**Test Count:** 20 tests (19 functional + 1 helper)

These tests migrate all JSONB-related assertions from the original SQL test suite to the new Rust/SQLx framework, ensuring consistent behavior across PostgreSQL versions 14-17.

---

**Depends on:** #138 (infrastructure)
**Next:** Equality operator tests

PR #140, opened 2025-10-27
test(sqlx): add equality operator and eq() function tests

tobyhede:
Add comprehensive Rust/SQLx tests for encrypted data equality operations.

## HMAC Index Equality
- Operator tests with matching encrypted values
- Operator tests with non-matching encrypted values
- Plain JSONB comparison validation

## Blake3 Index Equality
- Operator tests with matching encrypted values
- Operator tests with non-matching encrypted values
- `eq()` function tests for exact equality
- JSONB payload structure preservation (including 'ob' field)

## Key Differences Tested
- HMAC removes 'ob' field during equality comparison
- Blake3 preserves 'ob' field for complete payload equality
- Both support operator and function-based equality checks

## Migration Status

✅ **Like-for-Like Migration: Complete** (40/40 SQL assertions ported)

- Equality operators: 16/16 (HMAC + Blake3, operators + functions + JSONB)
- JSONB functions: 24/24 (arrays, paths, structure validation, encrypted selectors)

**Test Count:** 35 tests (34 functional + 1 helper)

These tests ensure encrypted data can be reliably compared for equality using both index-based operators and explicit `eq()` function calls, maintaining SQL parity across all supported PostgreSQL versions.

---

**Depends on:** #139 (JSONB tests)
**Completes:** SQLx test migration


freshtonic (approved):
Powering through - LGTM

PR #150, opened 2025-11-11
fix(docs): include markdown and XML in package archives

tobyhede:
Updated package.sh to include all three documentation formats:
- markdown/API.md
- xml/*.xml
- html/

Previously only packaged the html/ directory. Now archives contain complete documentation set suitable for distribution.


auxesis (approved):
Good pick up, thanks @tobyhede.

PR #171, opened 2026-04-21
feat(bench): Phase 2 - benchmark table, 10K row fixture, and verification tests

tobyhede:
## Summary

- `bench` table DDL (migration 007) — `encrypted_text`, `encrypted_int`, `encrypted_bigint` columns
- `bench_data.sql` opt-in fixture seeds 10K rows with a Zipf-like skew (`setseed(0.42)` + `random()^2`, 99 ids, three independent draws per row)
- `bench_setup.sql` opt-in fixture creates 5 benchmark indexes (hash/HMAC, 3× ORE btree, GIN/bloom)
- `bench_data_tests.rs` — 12 tests covering row count, column population, index term extraction, EXPLAIN index-usage assertions, seq scan baseline
- `FIXTURE_SCHEMA.md` and `migrations/README.md` updated

DDL is in the migration; data INSERT and indexes are in separate opt-in fixtures so non-bench tests skip the seeding cost and index-usage tests can assert before/after planner behaviour. Phase 2; Phase 1 (`pg_stat_statements`) is #169.

## Test plan

- [ ] `cargo test -p eql-tests bench_data_tests` on PG 14, 15, 16, 17

### DIFF
--- a/tests/sqlx/fixtures/FIXTURE_SCHEMA.md
+++ b/tests/sqlx/fixtures/FIXTURE_SCHEMA.md
@@ -152,11 +152,10 @@ CREATE TABLE bench (
 ```
 
 **Data:**
-- 10,000 rows cycling through 100 distinct encrypted values (ore ids 1-100)
-- Cycling offsets create varied column distributions:
-  - `encrypted_text`: ids 1, 2, ..., 100, 1, 2, ... (offset 0)
-  - `encrypted_int`: ids 35, 36, ..., 100, 1, ..., 34 (offset +34)
-  - `encrypted_bigint`: ids 68, 69, ..., 100, 1, ..., 67 (offset +67)
+- 10,000 rows drawn from 99 distinct encrypted values (ore ids 1-99)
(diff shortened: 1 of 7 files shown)


auxesis (approved):
LGTM, and you love to see it. 

An adjacent issue that needs to be addressed in a separate PR — [the test.yml workflow](https://github.com/cipherstash/encrypt-query-language/blob/c89356f08c0253c341ca8ac58e6864ad6ac07a8a/.github/workflows/test-eql.yml) have blown out quite a bit in the last 60 days: 

- At 04bd5c25fc82332a11a458d7750cfd3686d64293 (in #168), from ~3 minutes to ~23 minutes
- At 89f86f665399232d6d4a155c8a5a4446793b7807 (in this PR), from ~23 minutes to ~30 minutes

The SLO for PR checks is < 5 minutes, so currently we are way past that. 

We need to spend a moment to work out how to drive that way down.

tobyhede:
@auxesis Good catch. The culprit seems to be the ORE tests.

PR #176, opened 2026-04-27
feat(ope): add CLLW Order-Preserving Encryption index type

freshtonic:
Wires the OPE types from cipherstash-suite#1835 into EQL as a parallel to ORE. Numeric plaintexts use a 65-byte fixed ciphertext stored in the 'opf' SEM field; strings use a variable-length ciphertext in 'opv'.

Unlike ORE CLLW, OPE ciphertexts compare under standard lexicographic bytea ordering — no custom per-byte protocol — so the compare functions delegate to PostgreSQL's native bytea comparator. The generic eql_v2.compare dispatch now checks ope_cllw_u64_65 and ope_cllw_var_8 after the ORE variants and before the hash fallback.

Also adds 'ope' to config_check_indexes' valid set, refreshes the add_search_config docstring, and adds 12 SQLx tests covering SEM extraction, three-way compare, generic-dispatch routing, encrypted-type operators, prefix/length handling for variable-width OPE, and config validation.

## Summary by CodeRabbit

* **New Features**
  * Added Order-Preserving Encryption (OPE) support for range queries, ordering, and faster sorting paths alongside existing ORE/unique options.
  * Search/index kinds expanded to include an OPE index type.

* **Documentation**
  * Added guidance comparing ORE vs OPE, range-query examples, updated SQL/operator support matrix, and troubleshooting checks for OPE terms.

* **Tests**
  * Added comprehensive OPE tests covering extraction, comparisons, ordering, sorting fast-paths, and configuration validation.

### DIFF
--- a/src/ope_cllw_u64_65/functions.sql
+++ b/src/ope_cllw_u64_65/functions.sql
@@ -1,0 +1,89 @@
+-- REQUIRE: src/schema.sql
+-- REQUIRE: src/common.sql
+-- REQUIRE: src/ope_cllw_u64_65/types.sql
+
+
+--! @brief Extract CLWW OPE index term from JSONB payload
+--!
+--! Extracts the fixed-width CLWW OPE ciphertext from the 'opf' field of an
+--! encrypted data payload. Used internally for range query comparisons.
+--!
+--! @param jsonb containing encrypted EQL payload
+--! @return eql_v2.ope_cllw_u64_65 CLWW OPE ciphertext
+--! @throws Exception if 'opf' field is missing when ope index is expected
+--!
+--! @see eql_v2.has_ope_cllw_u64_65
+--! @see eql_v2.compare_ope_cllw_u64_65
+CREATE FUNCTION eql_v2.ope_cllw_u64_65(val jsonb)
+  RETURNS eql_v2.ope_cllw_u64_65
+  IMMUTABLE STRICT PARALLEL SAFE
+AS $$
+  BEGIN
+    IF val IS NULL THEN
+      RETURN NULL;
+    END IF;
+
+    IF NOT (eql_v2.has_ope_cllw_u64_65(val)) THEN
+        RAISE 'Expected a ope_cllw_u64_65 index (opf) value in json: %', val;
+    END IF;
+
+    RETURN ROW(decode(val->>'opf', 'hex'));
+  END;
+$$ LANGUAGE plpgsql;
+
+
+--! @brief Extract CLWW OPE index term from encrypted column value
+--!
+--! Extracts the fixed-width CLWW OPE ciphertext from an encrypted column value
+--! by accessing its underlying JSONB data field.
+--!
+--! @param eql_v2_encrypted Encrypted column value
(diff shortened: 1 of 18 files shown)


tests/sqlx/tests/ope_tests.rs
```diff
@@ -0,0 +1,187 @@
...
+    };
+
+    QueryAssertion::new(&pool, &cmp(&a, &b)).returns_int_value(-1).await;
+    QueryAssertion::new(&pool, &cmp(&b, &a)).returns_int_value(1).await;
+    QueryAssertion::new(&pool, &cmp(&a, &a)).returns_int_value(0).await;
+    Ok(())
+}
+
+#[sqlx::test]
+async fn generic_compare_dispatches_to_opf(pool: PgPool) -> Result<()> {
```
tobyhede, line 94: Add a test where both OPE and ORE are present with opposite orderings to lock in "ORE wins."

tests/sqlx/tests/ope_tests.rs
```diff
@@ -0,0 +1,187 @@
+//! OPE (CLWW Order-Preserving Encryption) tests
```
tobyhede, line 1: Add tests for `— <=, =, >=, <>` operators and for the `opv` variant - covered in compare, but not operators

src/ope_cllw_u64_65/functions.sql
```diff
@@ -0,0 +1,89 @@
...
+--!
+--! @see eql_v2.ope_cllw_u64_65
+CREATE FUNCTION eql_v2.has_ope_cllw_u64_65(val jsonb)
+  RETURNS boolean
+  IMMUTABLE STRICT PARALLEL SAFE
+AS $$
+  BEGIN
+    RETURN val ->> 'opf' IS NOT NULL;
+  END;
+$$ LANGUAGE plpgsql;
```
tobyhede, line 70: The new helpers could be vanilla `SQL` instead of `plpgsql` 
The PG planner can inline a LANGUAGE SQL function's body into the calling query, eliminating the call entirely.

Is actually official policy, despite existing inconsistency in a couple of place (`src/ore_cllw_var_8/functions.sql` (my bad)). 

```
Prefer `LANGUAGE SQL` over `LANGUAGE plpgsql` unless you need procedural features.
```

tobyhede (approved):
Some non-blocking recommendations and tweaks:

ORDER BY is not OPE-aware — src/operators/order_by.sql:22-29 and src/operators/sort.sql:421,545 only branch on ORE, OPE needs to be added

OPE needs to be added to `eql_v2.jsonb_array(val jsonb)` 
Currentlly: `WHERE kv.key IN ('s', 'b3', 'hm', 'ocv', 'ocf')` 

Reference docs not updated (sql-support.md, index-config.md, database-indexes.md) — no ope column/row.

tobyhede:
Notice commits not signed as well

tobyhede (approved):

tobyhede (approved):
LGTM.
Additional test for NULL handling would be good, but not a blocker.

PR #179, opened 2026-05-05
ci(splinter): add advisory Supabase splinter lint check

coderdan:
## Summary

Adds a non-blocking Supabase [splinter](https://github.com/supabase/splinter) job to `test-eql.yml` so we get visibility on database-linter findings against EQL without gating merges.

- New mise task `test:splinter` (`tasks/test/splinter.sh`) downloads splinter pinned to SHA `55db5b1f`, creates stub `anon` / `authenticated` roles if absent, runs the lints inside a transaction, prints a per-rule summary, and writes a markdown findings table to `$GITHUB_STEP_SUMMARY` in CI.
- New `splinter` job runs once on PG17 (a single artifact is enough — splinter's results don't vary across PG versions for the EQL install). `continue-on-error: true` keeps it advisory.

## Why advisory for now

Against current `main`, splinter emits **163 findings**:

| Level | Rule | Count |
| --- | --- | --- |
| WARN | `function_search_path_mutable` | 162 |
| WARN | `extension_in_public` | 1 |

PR #177 fixes ~143 of the search_path warnings on PL/pgSQL functions. The remaining ~37 are LANGUAGE SQL helpers that are intentionally left unpinned to preserve inlining (the `eql_v2.jsonb_contains` / `<@` GIN-index path). Closing them cleanly needs an `OPERATOR(pg_catalog.@>)`-style rewrite — out of scope for #177.

The `extension_in_public` finding is `pgcrypto` landing in the customer's `public` schema. Movable via `CREATE EXTENSION pgcrypto WITH SCHEMA eql_v2`, but worth doing as its own change.

So this PR ships the scaffolding now and a follow-up will flip the gate once those two items land.

## Flipping to blocking

One-line change once EQL is splinter-clean: delete `continue-on-error: true` on the job.

## Bumping splinter

Edit the `SPLINTER_SHA` constant at the top of `tasks/test/splinter.sh`.

## Test plan

- [x] `mise run test:splinter` runs locally against a fresh EQL install — 163 findings, exits 1
- [x] `GITHUB_STEP_SUMMARY=… mise run test:splinter` produces a well-formed markdown summary
- [x] Workflow YAML validates
- [ ] CI run on this branch shows the splinter job as advisory (red ❌ on the job, but PR still mergeable; full findings visible in the run's step summary)


.github/workflows/test-eql.yml
```diff
@@ -65,5 +61,41 @@ jobs:
...
+    # Advisory check — currently ~163 known findings (mostly function_search_path_mutable
+    # on LANGUAGE SQL helpers retained for inlinability). Flip continue-on-error to false
+    # once EQL is splinter-clean.
+    continue-on-error: true
+
+    env:
+      POSTGRES_VERSION: "17"
+
+    steps:
+      - uses: actions/checkout@v4
```
auxesis: ci: use latest actions/checkout

```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/test-eql.yml
```diff
@@ -65,5 +61,41 @@ jobs:
...
+
+      - uses: jdx/mise-action@v3
```
auxesis: ci: use latest jdx/mise-action
```suggestion
      - uses: jdx/mise-action@v4
```

.github/workflows/test-eql.yml
```diff
@@ -65,5 +61,41 @@ jobs:
...
+        with:
+          version: 2025.11.2
```
auxesis: ci: use consistent version of mise across jobs
```suggestion
          version: 2026.4.0
```

auxesis (approved):
Thanks for this @coderdan. 

Have approved with a couple of changes, and a conventional commit message you can copy for each. 

Three other things to note: 

- **The `test:splinter` mise task size:** Firstly, I love the way it fits with the rest of our tasks. ~120 lines of shell is at the threshold where I'd consider writing it in a scripting language like Python so it's easier to read and modify in the future. No action required now, but we should look at rewriting when we come back next and make changes to it.
- **Running `test:splinter` as part of `test`:** for all the other tasks in the `test:*` namespace, we invoke them via `test`. Given our decision right now not to fail the build on Splinter failures, I think we should not call `test:splinter` in `test`.
- **Decision point for `test:splinter` to fail the build:** do you have a sense of how much work is involved in making Splinter pass? I'd like to get Splinter fully wired into our test suite, but only if it's less than half a day's work.

coderdan:
Thanks @auxesis — applied all three suggestions across both jobs (test + splinter) for consistency: `actions/checkout@v6`, `jdx/mise-action@v4`, mise `2026.4.0`. Three separate commits with your suggested messages.

On the three notes:

1. **`test:splinter` task size** — agreed, ~120 lines of shell is at the edge. I'll plan a Python rewrite next time we touch it.
2. **Not invoking `test:splinter` from `test`** — already the case; `tasks/test.sh` only runs `test:lint`, `test:legacy`, and `test:sqlx`. No change needed.
3. **Effort to make splinter pass** — best estimate is **0.5–1 day**, but with non-trivial spike risk. Breakdown:
    - **`extension_in_public` (1 finding)** — trivial. `CREATE EXTENSION pgcrypto WITH SCHEMA eql_v2` (or move post-install). ~30 min including a regression check that `digest()`/`gen_random_bytes()` callers still resolve.
    - **`function_search_path_mutable` on ~37 LANGUAGE SQL helpers** — this is the unknown. We can't `SET search_path` (kills inlining, which is why we reverted that on these in #177). The clean fix is fully schema-qualifying every operator and function reference inside the body — `OPERATOR(pg_catalog.@>)`, `pg_catalog.jsonb_typeof(...)`, `eql_v2.foo(...)`, etc. Mechanical for most cases, but: (a) the GIN-index path on `eql_v2.jsonb_contains` / `<@` needs verification that inlining is preserved (EXPLAIN check), and (b) any helper that calls another EQL function needs the qualified form to chain correctly.

    If the spike on the GIN path goes smoothly, it's ~0.5 day. If it doesn't inline cleanly post-rewrite, it could balloon — in which case the right move is probably to leave splinter advisory and accept those specific findings via an allowlist rather than restructure. Happy to spike it as a follow-up and report back before committing to the full sweep.

PR #190, opened 2026-05-06
fix(splinter): make EQL splinter-clean and flip the gate

coderdan:
## Summary

Stacked on top of #179. Once that lands, this PR makes EQL splinter-clean against the pinned ruleset and flips the splinter job from advisory to blocking.

Empirical answer to @auxesis's question on #179: with help this was ~1–2 hours, not the half-day I'd estimated.

## What changed

**`src/crypto.sql`** — pgcrypto now installs into the `extensions` schema (Supabase convention) instead of `public`. Kills the `extension_in_public` finding. Existing function `search_path` settings already include `extensions`, so callers are unaffected on a fresh install. For pre-existing pgcrypto installs, the install-time check now picks one of three branches:

| pgcrypto schema | Behavior |
| --- | --- |
| `extensions` | silent OK |
| `public` | `RAISE NOTICE` recommending `ALTER EXTENSION pgcrypto SET SCHEMA extensions` (EQL still works) |
| anywhere else | `RAISE EXCEPTION` with the same remediation — fails install rather than failing later inside an encrypted comparison when `encrypt(...)` doesn't resolve |

**`src/_pin_search_path.sql`** (new) — post-install DO block that loops over `eql_v2.*` functions and applies `ALTER FUNCTION ... SET search_path = pg_catalog, extensions, public`. Skips:

- 8 GIN-critical wrappers — `@>`, `<@`, and the 6 `jsonb_contains`/`jsonb_contained_by` overloads — must remain inlinable. PostgreSQL's `inline_function` (`src/backend/optimizer/util/clauses.c`) bails when `proconfig` is set, which would unfold to a seq scan even with the GIN index in place.
- 3 aggregates (`min`, `max`, `grouped_value`) — `ALTER AGGREGATE` has no `SET configuration_parameter` syntax and `ALTER ROUTINE/FUNCTION` reject aggregates. Their SFUNCs are pinned.

Identifies the 8 wrappers by OID + (proname, proargtypes) so the comparison is independent of the caller's search_path. Uses `oid::regprocedure` for the ALTER target so the script doesn't need to mess with search_path itself (an earlier draft did `set_config('search_path', '', true)` and broke sqlx because `_sqlx_migrations` stopped resolving inside the install transaction).

REQUIREs only files present in every build variant (no `src/config/*` or `src/encryptindex/*` because protect excludes those; no `*operator_class.sql` because supabase excludes those).

**`tasks/test/splinter.sh`** — adds a TSV allowlist keyed on (rule, schema, name, type) with one-line justifications per entry. Splits findings into raw/allowlisted/unallowlisted, prints both in stdout and the GitHub step summary, and exits 0 only when unallowlisted = 0.

**`.github/workflows/test-eql.yml`** — drops `continue-on-error: true` on the splinter job and renames it from "Supabase splinter (advisory)" — the gate is now blocking.

## Empirical evidence (the central spike question)

The PR description on #179 hand-waved that closing the search_path findings on LANGUAGE SQL helpers needed an `OPERATOR(pg_catalog.@>)`-style rewrite to preserve inlining. The actual finding was narrower:

| Function | SET search_path effect on GIN index |
| --- | --- |
| `eql_v2.jsonb_array(jsonb)` | **Still uses Bitmap Index Scan** — expression-match doesn't require inlining |
| `eql_v2.jsonb_array(eql_v2_encrypted)` | Same |
| `eql_v2.jsonb_contains(...)` (any overload) | **Drops to Seq Scan** — wrapper must inline to expose `jsonb_array(a) @> jsonb_array(b)` |
| `eql_v2.jsonb_contained_by(...)` (any overload) | Same |
| `eql_v2.@>` / `eql_v2.<@` | Same |

So the inlining-critical surface is just the 8 wrappers, not the entire jsonb path.

## Findings: 73 → 0

Against a fresh EQL install:

| | Before | After |
| --- | --- | --- |
| `function_search_path_mutable` | 66 | 0 unallowlisted (11 allowlisted: 8 GIN-critical + 3 aggregates) |
| `extension_in_public` | 1 | 0 |
| `unused_index` | 6 (test fixtures, not present on a clean install) | 0 |
| **Total** | **73** | **0 unallowlisted** |

## Test plan

- [x] `mise run test:splinter` exits 0 against a fresh EQL install (raw=11, allowlisted=11, unallowlisted=0)
- [x] `cargo test -- --test-threads=2` — 362 passed, 0 failed, 3 ignored
- [x] `containment_with_index_tests` (the GIN oracle) — 14/14 pass; bug caught and fixed when the OID-based allowlist initially misfired and pinned everything
- [x] `build_validation_tests` — 14/14 pass; protect/supabase variants don't accidentally pull in excluded files via the new pinner's REQUIREs
- [x] pgcrypto install matrix — fresh DB lands in `extensions`; pre-existing in `public` emits NOTICE; pre-existing in custom schema raises EXCEPTION
- [ ] CI run on this branch shows the splinter job blocking and green

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened PostgreSQL extension schema handling to prevent runtime cryptographic failures in Supabase environments
  * Improved function search path resolution to ensure correct schema lookups

* **Chores**
  * Enhanced CI testing workflow with allowlist support to reduce false-positive findings and improve reliability

### DIFF
--- a/.github/workflows/test-eql.yml
+++ b/.github/workflows/test-eql.yml
@@ -71,12 +71,8 @@ jobs:
           mise run --output prefix test --postgres ${POSTGRES_VERSION}
 
   splinter:
-    name: "Supabase splinter (advisory)"
+    name: "Supabase splinter"
     runs-on: ubuntu-latest-m
-    # Advisory check — currently ~163 known findings (mostly function_search_path_mutable
-    # on LANGUAGE SQL helpers retained for inlinability). Flip continue-on-error to false
-    # once EQL is splinter-clean.
-    continue-on-error: true
 
     env:
       POSTGRES_VERSION: "17"
--- a/src/crypto.sql
+++ b/src/crypto.sql
@@ -6,10 +6,52 @@
 --! Enables the pgcrypto extension which provides cryptographic functions
 --! used by EQL for hashing and other cryptographic operations.
 --!
+--! Installs pgcrypto into the `extensions` schema (Supabase convention) to
+--! avoid the `extension_in_public` lint. Every EQL function that uses
+--! pgcrypto has `pg_catalog, extensions, public` on its `search_path`, so a
+--! pre-existing install in `public` keeps working — and a pre-existing
+--! install anywhere else will be rejected at install time rather than
+--! failing later inside an encrypted comparison.
+--!
 --! @note pgcrypto provides functions like digest(), hmac(), gen_random_bytes()
---! @note IF NOT EXISTS prevents errors if extension already enabled
+--! @note If pgcrypto is already installed in `public`, EQL works but emits
+--!       a NOTICE recommending `ALTER EXTENSION pgcrypto SET SCHEMA extensions`.
+--! @note If pgcrypto is already installed in any other schema, install
+--!       fails. Relocate it first with `ALTER EXTENSION pgcrypto SET SCHEMA
+--!       extensions` (or move it into `public` if compatibility with other
+--!       consumers requires it).
 
---! @brief Enable pgcrypto extension
---! @note Provides cryptographic functions for hashing and random number generation
-CREATE EXTENSION IF NOT EXISTS pgcrypto;
+--! @brief Create extensions schema (Supabase convention)
+CREATE SCHEMA IF NOT EXISTS extensions;
 
+--! @brief Enable pgcrypto extension and validate its schema
+DO $$
+DECLARE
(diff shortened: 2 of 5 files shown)


freshtonic (approved):
I'm not across the internal details of how forcing schema search paths works on a per-function basis but I get the intent and understand the tests.

PR #197, opened 2026-05-07
fix(ope): address CodeRabbit and reviewer feedback on PR #176

freshtonic:
Follow-up to #176 addressing late CodeRabbit and reviewer feedback.

## Code changes

**`compare_ope_cllw_*` STRICT cleanup**
- Drop dead outer NULL-handling branches (function inputs `a`/`b` short-circuit to NULL via STRICT before the body runs).
- Keep the inner term-NULL branches — they handle a non-NULL `eql_v2_encrypted` whose payload lacks the `opf` / `opv` field, mirroring `compare_ore_block_u64_8_256`.
- Add `@note` in both compare.sql files explicitly distinguishing the two NULL paths.

**OPE helper inlinability**
- Convert `ope_cllw_u64_65(eql_v2_encrypted)`, `ope_cllw_var_8(eql_v2_encrypted)`, both `has_ope_*` overloads, and `order_by_ope` from `LANGUAGE plpgsql` to `LANGUAGE sql` so the planner can inline them on the sort / order_by hot path.

**Doc fixes**
- Type comments: ciphertext length now correctly reads "8 bytes per plaintext byte" (was "per plaintext bit", contradicting the 65-byte result).
- `CLWW` → `CLLW` everywhere — matches the `ope_cllw_*` type names and existing `ore_cllw_*` docs.
- `@param` tags in `ope_cllw_*/functions.sql` now name the parameter (`val`) before the type, matching valid Doxygen syntax.
- `docs/reference/sql-support.md`:
  - `MIN`/`MAX` and `COUNT(DISTINCT)` rows separate ste_vec-extracted-node requirements from outer-column sibling-index requirements. `ope` is never emitted by ste_vec extraction; it only applies to the outer column.
  - `COUNT(DISTINCT col)` outer-column cell lists `unique` / `ore` / `ope` for parity with the `MIN`/`MAX` row.

## Test changes

**Operator parity** — `<`, `<=`, `=`, `>=`, `>`, `<>` exercised via the `eql_v2_encrypted` operators on both `opf` and `opv` variants (previously only `<` and `>` were covered for `opf`; `opv` was exercised through `compare_*` but not via operators).

**ORE-vs-OPE precedence** — `ore_wins_over_opf_when_both_present` locks in the ORE-before-OPE dispatch order in `eql_v2.compare` when a payload carries both `ob` and `opf`.

**NULL-handling parity with ORE** — 18 tests covering scenarios where ORE had coverage and OPE had none:
- `{"opf": null}` / `{"opv": null}` payloads dispatch through to HMAC (mirrors `compare_hmac_with_null_ore_index`).
- `compare_ope_cllw_*` returns NULL for both NULL-left and NULL-right operands (STRICT contract codified).
- `ORDER BY` × ASC/DESC × `NULLS FIRST`/`NULLS LAST` with mixed NULL + opf rows.
- `eql_v2.sort_compare` with mixed NULL + opf rows.
- `eql_v2.min` / `eql_v2.max` over OPE-encoded values, including NULL-only → NULL.
- `BETWEEN` for both `opf` and `opv` variants.

**Brittle assertions** — `config_check_rejects_unknown_index` no longer matches the exact comma-separated list of valid indexes; checks for `ope` + the offending token instead, so adding a new valid index won't break the test.

## Open review items

- **`src/ope_cllw_u64_65/compare.sql`** (tobyhede, line 30): the dropped outer NULL branches drew the comment _"This might have unintended consequences. Let me try and find the history."_ The branches are dead under `STRICT`, but their removal diverges from the ORE pattern which keeps both outer (input-NULL) and inner (term-NULL) layers. Awaiting follow-up — happy to restore the outer branches as defense-in-depth if that's preferred.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected NULL handling in comparison functions to align with STRICT semantics; NULL inputs now short-circuit before function execution.
  * Fixed OPE extraction and presence-check behavior for encrypted columns.

* **Documentation**
  * Clarified ordered comparison behavior between OPE and ORE index configurations and their operator dispatch rules.
  * Updated comparison function documentation to reflect proper NULL handling and term extraction semantics.


src/ope_cllw_u64_65/compare.sql
```diff
@@ -27,18 +27,6 @@ AS $$
     a_term eql_v2.ope_cllw_u64_65;
     b_term eql_v2.ope_cllw_u64_65;
   BEGIN
-    IF a IS NULL AND b IS NULL THEN
```
tobyhede, line 30: This might have unintended consequences. 
Let me try and find the history.

tobyhede (approved):

docs/reference/sql-support.md
```diff
@@ -12,7 +12,7 @@ EQL ships five search index kinds that encrypt data in ways that preserve specif
 | `match`                            | `bloom_filter` (`bf`)        | Substring / token matching via `LIKE` / `ILIKE`        |
 | `ste_vec`                          | Structured encryption (`sv`) | JSONB containment and JSONB path / field access        |
 
-> **`ore` vs `ope`** — both index kinds support the same ordered-comparison surface. `ore` (Order-Revealing Encryption) is the default. `ope` (CLWW Order-Preserving Encryption) is an alternative for environments that need plain lexicographic byte comparison (e.g. pluggable storage that cannot run a custom comparator). On a column configured for `ope`, `eql_v2.compare()` and the `<` / `<=` / `>` / `>=` operators dispatch to OPE terms automatically.
+> **`ore` vs `ope`** — both index kinds support the same ordered-comparison surface. `ore` (Order-Revealing Encryption) is the default. `ope` (CLLW Order-Preserving Encryption) is an alternative for environments that need plain lexicographic byte comparison (e.g. pluggable storage that cannot run a custom comparator). On a column configured for `ope`, `eql_v2.compare()` and the `<` / `<=` / `>` / `>=` operators dispatch to OPE terms automatically.
```
coderdan: Chennette-Lewi-Weis-Wu = CLWW

coderdan (approved):
Just the CLLW vs CLWW thing. Otherwise LGTM and I'll work on a follow up.

PR #211, opened 2026-05-13
Inline range operators (<, <=, >, >=) on eql_v2_encrypted to Block ORE term comparison

coderdan:
## Summary

Mirrors the EQL 2.3 `=` → `hmac_256` inlining for the range operators on `eql_v2_encrypted`. Each `<` / `<=` / `>` / `>=` overload (three per operator) becomes single-statement inlinable SQL with body `eql_v2.ore_block_u64_8_256(a) <op> eql_v2.ore_block_u64_8_256(b)`. The planner can now structurally match bare-form range predicates against a functional btree index on `eql_v2.ore_block_u64_8_256(col)` (using the existing DEFAULT-for-type `eql_v2.ore_block_u64_8_256_operator_class`).

Motivated by a perf cliff on the benches: `WHERE val < $1 ORDER BY val LIMIT 10` at 100k rows was 6.3 s pre-change (and slow enough to expire the scoped-cipher TTL mid-decrypt during a criterion sweep). With this change:

| Query shape | 100k rows |
|---|---|
| `WHERE val < $1 ORDER BY val LIMIT 10` (natural) | **~885 ms** (down from 6.3 s) |
| `WHERE val < $1 ORDER BY eql_v2.ore_block_u64_8_256(val) LIMIT 10` (hybrid — sort key matches index expression) | **~1.4 ms** |
| `WHERE val < $1 LIMIT 10` (non-ordered range) | **~7 ms** |

## Mechanics

For the inlining chain to reach index matching, the inner operator backing functions on `eql_v2.ore_block_u64_8_256` also had to be inlinable. `eql_v2.ore_block_u64_8_256_{eq,neq,lt,lte,gt,gte}` were declared `LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE` — without this, they defaulted to `VOLATILE`, which made the outer operator body implicitly volatile and blocked inlining. The stale "FILE IS DISABLED" `!REQUIRE` comments on `src/ore_block_u64_8_256/{operators,operator_class}.sql` were also cleaned up; those files were already in the build via the find self-dep, but their dependency on `functions.sql` wasn't declared, which broke tsort ordering once another file started requiring them.

`tasks/pin_search_path.sql` and `tasks/test/splinter.sh` allowlists updated to keep the new inlinable functions unpinned.

## Behaviour change

Range operators no longer fall through the `eql_v2.compare()` priority list (Block ORE → CLLW u64 → CLLW var → OPE → hmac → literal). Columns carrying only `ore_cllw_u64_8`, `ore_cllw_var_8`, `opf`, or `opv` terms now raise from the `ore_block_u64_8_256` extractor when used with `<` / `<=` / `>` / `>=`. Callers there must rewrite to the matching extractor form, e.g. `WHERE eql_v2.ore_cllw_u64_8(col) < eql_v2.ore_cllw_u64_8($1::jsonb)`.

Existing test coverage for the dispatcher paths (13 in `comparison_tests.rs`, 12 in `ope_tests.rs`, 2 sort tests) is `#[ignore]`'d with a pointer to U-005. Re-enable once a CASE-style operator body re-introduces broader ORE/OPE support under the inlined operators (out of scope for this PR).

## Docs

- `CHANGELOG.md` `[Unreleased]` Changed entry.
- New upgrade note `U-005` in `docs/upgrading/v2.3.md`.
- `docs/reference/database-indexes.md` Range Queries section rewritten around the functional ORE recipe (including the natural-vs-hybrid ORDER BY trade-off).

## Test plan

- [x] `mise run build` clean.
- [x] Plan-assertion tests for natural-form `<` / `>` and jsonb cross-type against the functional ORE index (`comparison_tests.rs::natural_form_*_engages_functional_ore_index`).
- [x] Hybrid-form test (`hybrid_form_lt_engages_functional_ore_index_without_sort`) confirms no Sort node.
- [x] Raise-on-missing-`ob` regression test (`lt_on_column_without_ob_term_raises`).
- [x] Full SQLx suite passes on the local Postgres 17 (391 pass, 39 ignored covering the deferred breakage scope).
- [x] CI: re-run against Postgres 14 / 15 / 16 / 17.
- [x] Manual EXPLAIN smoke on the bench container at 100k rows confirms `Index Scan using *_ore_index` with `Index Cond: ore_block(value) < ore_block(...)`.
- [x] Bench smoke: `range_lt_ordered_10` 6.3 s → ~885 ms at 100k; new `range_lt_hybrid_ordered_10` lands at ~1.4 ms.

## Summary by CodeRabbit

* **Changed**
  * Range operators (<, <=, >, >=) are now inlinable and require Block ORE; bare-form range predicates can use functional B-tree indexes and will error if the block term is missing.
  * ORDER BY/WHERE interactions clarified for index-ordered streaming.

* **Deprecated**
  * Old plpgsql helper comparison functions (lt/lte/gt/gte) marked for removal in 3.0.

* **Documentation**
  * Added upgrade/migration guidance and expanded index usage and performance notes.

* **Tests**
  * Several range- and order-related tests are temporarily ignored pending migration support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/encrypt-query-language/pull/211?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/operators/<.sql
```diff
@@ -1,12 +1,16 @@
...
 -- REQUIRE: src/operators/compare.sql
+-- REQUIRE: src/ore_block_u64_8_256/functions.sql
+-- REQUIRE: src/ore_block_u64_8_256/operators.sql
 
 --! @brief Less-than comparison helper for encrypted values
 --! @internal
 --!
 --! Internal helper that delegates to eql_v2.compare for less-than testing.
---! Returns true if first value is less than second using ORE comparison.
+--! Kept for callers that invoke it directly (and used by `eql_v2.compare`
```
tobyhede: Comment is unclear - `compare` does not use `lt`

src/operators/<.sql
tobyhede, line 30: The intent of a single canonical `lt` function was to avoid duplicating logic.
The `SELECT eql_v2.ore_block_u64_8_256(a) < eql_v2.ore_block_u64_8_256(b)` now appears ibn multiple places. 

Indirection may not be worth the effort and may conflict with `inlineable` requirement, but may be worth keeping the existing structure -> single implementation with typed aliases for the operator definitions.

tobyhede (approved):

src/operators/<.sql
```diff
@@ -41,16 +46,31 @@ $$ LANGUAGE plpgsql;
...
+-- `eql_v2.ore_block_u64_8_256_operator_class`). Bare range queries
+-- (`WHERE col < $1`) engage the functional ORE index on Supabase and any
+-- install that doesn't ship `eql_v2.encrypted_operator_class`.
+--
+-- Behaviour change vs the previous dispatcher-based impl: the old
+-- `eql_v2."<"` walked `eql_v2.compare`, which dispatched through
+-- ore_block / ore_cllw_u64 / ore_cllw_var / ope. Now `<` requires the
+-- column to have `ore_block_u64_8_256` configured (i.e. carry an `ob`
+-- field). Calling `<` on a column with only `ore_cllw_*` or OPE terms
+-- will return NULL where it previously returned a Boolean — same
```
freshtonic: The comment about returning `NULL` is incorrect.

`eql_v2.ore_block_u64_8_256(jsonb)` raises - `RAISE 'Expected an ore index (ob) value in json: %'` - it does not return `NULL`. The CHANGELOG entry, `database-indexes.md`, and the new `lt_on_column_without_ob_term_raises` test all correctly say raises.

src/operators/>.sql
```diff
@@ -25,29 +28,32 @@ $$ LANGUAGE plpgsql;
...
 --!
---! @see eql_v2.compare
+--! @see eql_v2.ore_block_u64_8_256
 --! @see eql_v2.add_search_config
+-- Inlinable: see `src/operators/<.sql` for the rationale. Predicate
+-- `WHERE col > val` reduces to
+-- `WHERE eql_v2.ore_block_u64_8_256(col) > eql_v2.ore_block_u64_8_256(val)`
+-- and matches a functional ORE index built on the same expression.
+-- Breaking impact: columns with only `ore_cllw_*` or OPE terms return
+-- NULL for `>` where they previously fell through `eql_v2.compare`.
```
freshtonic: The comment about returning `NULL` is incorrect.

`eql_v2.ore_block_u64_8_256(jsonb)` raises - `RAISE 'Expected an ore index (ob) value in json: %'` - it does not return `NULL`. The CHANGELOG entry, `database-indexes.md`, and the new `lt_on_column_without_ob_term_raises` test all correctly say raises.

freshtonic (approved):
Solid, well-scoped change. One should-fix: correct the `NULL`-vs-raises claims in the `<.sql` and `>.sql` inline comments before merging.

**Worth noting**:  ~25 `comparison_tests.rs/ope_tests.rs` cases plus sort tests are now #[ignore]'d. The OPE/CLLW comparison paths (eql_v2.lt/lte/gt/gte, eql_v2.compare's priority list, the ope/ore_cllw operator support) are no longer exercised through the operators - regressions there won't be caught by the operator suite.

src/operators/<.sql
coderdan: Valid concern. Let me investigate.

src/operators/<.sql
coderdan: You're right. Good catch!
FWIW, that raise behaviour doesn't seem to have affected performance either which is great news (raise is def preferred IMO).

PR #212, opened 2026-05-13
docs: query-performance guide

coderdan:
## Summary

A new reference page that consolidates the practical advice for getting fast queries out of EQL-encrypted columns. The 2.3 operator-inlining work surfaced enough subtleties — natural vs extractor forms, the `ORDER BY` sort-key trap, the inlining chain through helpers, when a functional index will and won't match — that they warrant a dedicated page rather than being scattered across `database-indexes.md`, the upgrade notes, and tribal knowledge.

## Outline

1. Why functional indexes — small leaves, no superuser, structural planner match.
2. Operator inlining mechanics — the four conditions PG checks, the syntactic match rule, the transitive nature of the chain, EXPLAIN-based verification.
3. Natural / extractor / hybrid query forms — when each is the right default, with examples.
4. `ORDER BY`: the sort-key trap — three-shape contrast (natural / hybrid / fully extractor) with the 100k benchmark numbers as evidence (885 ms vs 1.4 ms).
5. Equality and `GROUP BY` / `DISTINCT` — including the ~425× speedup from `GROUP BY` on the inlined extractor vs the natural form's plpgsql `hash_encrypted` path.
6. `LIKE` / `ILIKE` — bloom-filter recipe and the case-sensitivity caveat.
7. JSONB containment and ste_vec field-level extraction — per-selector vs all-selector recipes.
8. Common pitfalls — index-creation ordering, missing `ANALYZE`, stale opclass indexes, pinned `search_path`, range queries on non-Block-ORE.
9. Diagnosing with `EXPLAIN` — what to look for, what to do when the plan is wrong.

## Placement

Landed at `docs/reference/query-performance.md` and linked from `docs/README.md` next to `database-indexes.md`. Happy to relocate to a new `docs/guides/` directory if we want to start separating action-oriented content from reference. Not blocking on that for this PR — easier to iterate on the content first.

## Related

Depends on the contract established by:

- [#211](https://github.com/cipherstash/encrypt-query-language/pull/211) — range-operator inlining (`<` / `<=` / `>` / `>=`). U-005 callouts in the guide reference that PR's upgrade note.
- The 2.3 `=` inlining (#193, #196) — already referenced via U-002.

## Test plan

- [x] Markdown renders cleanly (checked in preview).
- [x] All in-doc links resolve (database-indexes, sql-support, index-config, v2.3 upgrade notes).
- [ ] Review pass for accuracy on the numbers and the claims about planner behaviour.
- [ ] Decide whether to keep at `reference/` or move to a new `guides/`.

Draft because the content is a first pass — happy to iterate.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide for writing performant queries against encrypted/EQL-backed columns, covering index strategies, query patterns (equality, range, LIKE/ILIKE, JSONB), operator inlining, ORDER BY/GROUP BY considerations, troubleshooting with EXPLAIN, and index-build best practices for large tables.
  * Updated reference list to include the new performance guide.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/encrypt-query-language/pull/212?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/reference/query-performance.md
+++ b/docs/reference/query-performance.md
@@ -173,19 +173,9 @@ SELECT * FROM users WHERE email_encrypted = $1;
 --   Index Cond: ((eql_v2.hmac_256(email_encrypted))::text = (eql_v2.hmac_256(...))::text)
 ```
 
-**`GROUP BY` is a different beast.** PostgreSQL evaluates `GROUP BY col` by hashing each row's `col` value using the type's hash discriminator. For `eql_v2_encrypted`, that discriminator is `eql_v2.hash_encrypted`, which is plpgsql and called once per row. Even with a functional hash index in place, the index doesn't help `GROUP BY` directly (hash aggregation is in-memory, not index-driven).
+**`GROUP BY` is a different beast — and the extractor form is the only recipe that scales.** Use it. Even at moderate row counts the natural form's plan choice degrades pathologically; the extractor form sidesteps the trap entirely and works the same way on every Postgres deployment, including Supabase and managed services where you can't tune `work_mem`.
 
-The natural form is:
-
-```sql
-SELECT email_encrypted, count(*)
-  FROM users
-  GROUP BY email_encrypted;
-```
-
-This works, but the per-row hash is a plpgsql function call — ~hundreds of microseconds per row, not hundreds of nanoseconds.
-
-The extractor form is dramatically faster on large tables:
+The canonical recipe for `GROUP BY` (and `DISTINCT`, and `IN (subquery)`, and hash joins) on an encrypted column:
 
 ```sql
 SELECT eql_v2.hmac_256(email_encrypted), count(*)
@@ -193,9 +183,42 @@ SELECT eql_v2.hmac_256(email_encrypted), count(*)
   GROUP BY eql_v2.hmac_256(email_encrypted);
 ```
 
-The body of `eql_v2.hmac_256` is inlinable SQL — `(col).data ->> 'hm'` — so the planner folds it into the aggregation and each row pays a single jsonb lookup instead of a plpgsql function call. On 10k rows the difference is around 425×; the gap grows with the table.
+Why this is the right shape:
 
-If the column also has a `bloom_filter` for `LIKE` (which is `smallint[]`), `DISTINCT` over the extractor form benefits the same way.
+- **The group key is small.** `eql_v2.hmac_256(col)` returns a 32-byte HMAC. At 1M rows the in-memory hash table is well under 100 MB, which fits inside the default `work_mem = 4MB` per partition many times over — so the planner picks `HashAggregate` reliably, without any deployment-wide tuning.
+- **The extractor inlines.** `eql_v2.hmac_256(val)` is a single-statement SQL function whose body is essentially `(val).data ->> 'hm'`. The planner folds it into the aggregation, so each row pays a single jsonb lookup. No plpgsql function-call overhead per row, no opaque dispatch.
+- **It matches a functional index.** If you have a `unique` search config on the column and a `CREATE INDEX … USING hash (eql_v2.hmac_256(col))` — the same index you'd build for fast equality — `IN (subquery)` and hash joins against `eql_v2.hmac_256(col)` can use it directly. `GROUP BY` itself doesn't use the index (hash aggregation is in-memory, not index-driven), but everything else that hashes against the same key does.
+
+### Why the natural form doesn't scale
+
+The natural form looks fine on paper:
+
+```sql
+-- Avoid. Falls into the work_mem / planner trap below.
+SELECT email_encrypted, count(*)
+  FROM users
+  GROUP BY email_encrypted;
+```
+
+The trap is in the planner's cost model. PostgreSQL has two aggregation strategies — `HashAggregate` (in-memory hash table keyed by the GROUP BY expression) and `GroupAggregate` (sort the input, then collapse adjacent equal rows). The planner picks between them on cost, and the cost model is *very* sensitive to the size of the group key — because the hash table for HashAggregate has to fit in `work_mem`.
+
+On an encrypted column, the natural-form key is the entire `eql_v2_encrypted` payload — typically 1-2 KB per row. At 100k rows the planner estimates a 100-200 MB hash table, which exceeds the default `work_mem = 4MB` by two orders of magnitude. The planner refuses HashAggregate and falls back to GroupAggregate. GroupAggregate sorts the input, which means an O(N log N) pass over kilobyte-sized rows — the per-comparison cost is small (the `=` operator is inlinable SQL post-2.3, so each comparison reduces to a 32-byte HMAC comparison), but the sort still dominates wall-clock time and spills to disk past `work_mem`.
+
+Measured at 100k rows on the bench tables:
+
+| Query form | Plan | Time |
+| --- | --- | --- |
+| `GROUP BY col` (natural, default 4 MB work_mem) | GroupAggregate + Sort (disk spill) | ~29 s |
+| `GROUP BY col` (natural, work_mem bumped to 256 MB) | HashAggregate | ~780 ms |
+| `GROUP BY eql_v2.hmac_256(col)` (extractor) | HashAggregate | ~80 ms |
+
+Bumping `work_mem` recovers the natural form from "minutes" to "sub-second" by changing the planner's choice — a 37× improvement just from one setting. The extractor form is another ~10× on top of that and doesn't depend on a deployment-wide knob. At 1M rows the natural form's GroupAggregate sort takes around 4 minutes even with `work_mem = 512MB`; the extractor form stays sub-second.
+
+The takeaway: write the extractor form. If you have a query you can't rewrite (an ORM that's GROUP BY-ing the raw column, a third-party report tool), bumping `work_mem` to a size that fits the estimated hash table is the rescue knob — but don't make that the design.
+
+### `DISTINCT` and ste_vec field-level
+
+`DISTINCT col` follows the same rules — `SELECT DISTINCT eql_v2.hmac_256(col) FROM tbl` is the recipe. If you only need set membership and not the original encrypted value back, the extractor form is what you want regardless of table size.
 
 For ste_vec documents, field-level `GROUP BY` works analogously:
 
@@ -260,7 +283,8 @@ For field-level lookups (`data_encrypted->'email' = $1`), use the per-selector h
 - **Stale opclass index alongside a functional index.** If you migrate an old schema from `eql_v2.encrypted_operator_class` to functional indexes, drop the old opclass index. Two btree indexes on the same column compete for cache and double the maintenance cost on writes.
 - **Pinning `search_path` on an EQL function.** Adding `SET search_path = …` to an `eql_v2.*` function disables inlining and reverts queries through that function to sequential scans. The EQL build allowlists operator wrappers that must stay inlinable; if you're customising the install, preserve that allowlist.
 - **`ORDER BY` on the natural form expecting an `Index Scan`.** The Sort node is required (§4). If you need it gone, switch the `ORDER BY` to extractor form.
-- **`=` / `<>` returning zero rows silently.** Equality requires `hm`. On a column without a `unique` search config, `WHERE col = $1` either raises (on lookup) or returns zero rows. `eql_v2.hash_encrypted` raises loudly when used in `GROUP BY` / `DISTINCT` — that's the canonical smoke test for "is `hm` configured?".
+- **`=` / `<>` returning zero rows silently on a column without `hm`.** Equality requires the column to carry an `hm` HMAC term. Without it, `eql_v2.hmac_256(col)` returns NULL and the operator comparison evaluates to NULL — false in a WHERE context. `eql_v2.hash_encrypted` (the discriminator behind `GROUP BY` / `DISTINCT` / hash joins) doesn't raise on missing `hm` either — it falls back to hashing the encrypted payload bytes, which keeps the aggregate's hash table from degrading to O(N²) on a single NULL bucket but no longer gives you a runtime smoke signal. Audit at config time instead: `SELECT eql_v2.has_hmac_256(col) FROM tbl LIMIT 1`.
+- **`GROUP BY` on the raw encrypted column at scale.** §5 details the trap, but it's worth repeating in pitfalls form: `GROUP BY col` on an `eql_v2_encrypted` column past ~10k rows falls into GroupAggregate-with-disk-spill territory because the natural-form key is 1-2 KB per row and the hash table can't fit in `work_mem`. Write `GROUP BY eql_v2.hmac_256(col)` instead. The extractor form is the only `GROUP BY` recipe that holds up at production scale on default Postgres settings.
 - **Range queries (`<`, `<=`, `>`, `>=`) on columns with only `ore_cllw_*` or OPE terms.** The range operators are Block ORE only post-2.3 (see [U-005 in v2.3.md](../upgrading/v2.3.md#u-005-range-operators-are-block-ore-only)). Migrate the column to `ore` or switch the query to the extractor form for the relevant CLLW / OPE encoding.
 
 ---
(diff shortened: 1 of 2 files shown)


docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
+# Writing fast queries against EQL columns
+
+This guide is about getting query performance out of EQL-encrypted columns that's competitive with plain-PostgreSQL workloads. It explains the two practical ingredients — **functional indexes** and **operator inlining** — and shows how to combine them across the common query shapes (`=`, `<` / `>`, `ORDER BY`, `GROUP BY`, `LIKE`, JSONB containment, ste_vec field-level access).
```
yujiyokoo, line 3: Maybe good to mention the version this applies to at the start?

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+## 1. Why functional indexes
+
+The two recipes for putting a PostgreSQL index on an `eql_v2_encrypted` column are:
+
+| Recipe | Shape | Example |
+| --- | --- | --- |
+| **Functional** *(canonical)* | Index over a deterministic extractor that yields a small per-row term | `CREATE INDEX … ON users (eql_v2.hmac_256(email_encrypted));` |
+| **Operator class** *(legacy)* | Index over the whole `eql_v2_encrypted` column via a custom btree opclass | `CREATE INDEX … ON users (email_encrypted eql_v2.encrypted_operator_class);` |
+
+The operator-class recipe ships with EQL and still works, but functional indexes are the recommended path for new schemas because:
```
yujiyokoo: I reckon it might be a good idea to emphasise that functional index is recommended a little here

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+
+1. **Small leaves.** A functional index on `eql_v2.hmac_256(col)` stores only the 32-byte HMAC per row. The operator-class index stores the entire encrypted payload (often kilobytes), inflating the btree and risking the `index row size N exceeds btree version 4 maximum 2704` error on full-payload columns.
```
yujiyokoo: Nit: something like "space efficiency" may be easier to understand for some

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+
+---
+
+## 2. Operator inlining: the mechanics
+
+PostgreSQL inlines a SQL function when **all** of these conditions hold:
+
+- `LANGUAGE sql` (not `plpgsql`).
+- The body is a single `SELECT` returning the same type the function declares.
+- No `SET` clause on the function definition (in particular, no `SET search_path = …`).
```
yujiyokoo: The part inside parentheses feel slightly confusing, as `SET` blocks inlining whether it's `search_path` or not. I reckon it could be something like:

> No `SET` clause on the function definition (most commonly SET search_path, but any other `SET` clause would blocks inlining)

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+- The function has a `SET` clause (`\df+ eql_v2.…` will show `proconfig`).
+- The function is `plpgsql` (look at `prolang`).
+- An inner helper that the operator body calls is `VOLATILE` or has a `SET` clause — inlining is transitive, so an inlinable wrapper around a non-inlinable helper still won't inline.
+- The planner thinks a different plan is cheaper (e.g. an `ORDER BY pk LIMIT n` driving the primary-key index instead of your functional index). Force the question with `SET enable_seqscan = off` or stronger selectivity to see if the planner *can* use the index.
+
+---
+
+## 3. Natural form vs extractor form
+
+The operators ship with three overloads each — `(encrypted, encrypted)`, `(encrypted, jsonb)`, `(jsonb, encrypted)` — so all three predicate shapes inline equivalently. After inlining, `col = $1`, `col = '{…}'::jsonb` and `'{…}'::jsonb = col` all reduce to the same canonical expression. There's no perf penalty for any of those bindings; pick whichever fits your client.
```
yujiyokoo: nit: perf -> performance

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+
+- The function has a `SET` clause (`\df+ eql_v2.…` will show `proconfig`).
+- The function is `plpgsql` (look at `prolang`).
+- An inner helper that the operator body calls is `VOLATILE` or has a `SET` clause — inlining is transitive, so an inlinable wrapper around a non-inlinable helper still won't inline.
+- The planner thinks a different plan is cheaper (e.g. an `ORDER BY pk LIMIT n` driving the primary-key index instead of your functional index). Force the question with `SET enable_seqscan = off` or stronger selectivity to see if the planner *can* use the index.
+
+---
+
+## 3. Natural form vs extractor form
+
```
yujiyokoo, line 87: It might read better with a small introduction like:

> EQL has two ways to express the comparison with encrypted columns.

Then:

> **Natural form** ...

> **Extractor form** ...

> The operators ship with three overloads each...

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+- **The column doesn't carry the canonical term for the natural-form operator.** E.g. a column configured with `ore_cllw_u64_8` instead of `ore_block_u64_8_256` will raise from the natural-form `<` after the range-operator inlining — write `WHERE eql_v2.ore_cllw_u64_8(col) < eql_v2.ore_cllw_u64_8($1::jsonb)` instead.
+- **You want a sort key that the index can satisfy without a Sort node** (see §4).
+- **You want a per-field index on an ste_vec document** — the field-level recipes (`eql_v2.hmac_256(col, '<selector>')` and `eql_v2.hmac_256_terms(col)`) only work in extractor form because the selector isn't part of any natural-form SQL operator.
+- **You're debugging a plan and want to bypass the inlining question entirely.** Plans with the extractor written out leave nothing to inference; the predicate matches the index by direct text identity.
+
+---
+
+## 4. `ORDER BY`: the sort-key trap
+
```
yujiyokoo, line 130: Starting the section with a little more intro may be helpful. Something like:

> WHERE and ORDER BY look symmetric in SQL, but they reach a functional index through different paths. The planner inlines operators in predicates, so WHERE col < $1 rewrites itself to match an index on f(col). It does no such rewrite for sort keys: ORDER BY col and ORDER BY f(col) are not interchangeable to the planner, even when they'd sort the rows identically.

(written by Claude Code)

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+Same trap, same fix at the sv-element level. With a Standard-mode `ste_vec` index the orderable term on each sv element is `oc` (ORE CLLW), and the canonical recipe is a functional btree on the *extractor* applied to a selector:
+
+```sql
+-- Build a functional ORE-CLLW index on the orderable sv element
+CREATE INDEX users_age_oc_idx
+  ON users (eql_v2.ore_cllw(data_encrypted -> '<age-selector>'::text));
+ANALYZE users;
+```
+
+(The `eql_v2.ore_cllw_ops` opclass is `DEFAULT FOR TYPE eql_v2.ore_cllw`, so no explicit opclass annotation is needed.)
```
yujiyokoo: I don't think you need parentheses here

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+
+Three query shapes at the field level:
```
yujiyokoo: There are 2 examples here

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+
+There's a second, subtler way to lose the index-driven sort, even when you've already written `ORDER BY` correctly. When the `SELECT` projects the same column you `ORDER BY` and applies a cast to that projection — typically `value::jsonb` to coerce an `eql_v2_encrypted` column into something a generic driver can decode — PostgreSQL pushes the cast into the inner scan output and **uses the projected (post-cast) expression as the sort key**. `EXPLAIN` will show `Sort Key: ((value)::jsonb)` rather than `Sort Key: value`. That expression is syntactically distinct from anything you could index — neither a btree on `value` with the default `eql_v2.encrypted_operator_class` nor a functional btree on `eql_v2.ore_block_u64_8_256(value)` matches `(value)::jsonb`.
+
+```sql
+-- TRAP: cast in SELECT + bare column in ORDER BY
+-- Plan: Sort Key: ((value)::jsonb), no index-for-sort, falls back to
+--       SeqScan/Bitmap + Sort even though a perfectly good index exists.
+SELECT id, value::jsonb FROM events
+  WHERE encrypted_at < $1
+  ORDER BY encrypted_at LIMIT 10;
```
yujiyokoo: These should be `value` rather than `encrypted_at`?

yujiyokoo: Oh, coderabbit had already pointed it out

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+  ORDER BY encrypted_at LIMIT 10
+) sub;
+```
+
+Important: this trap only fires when the sort key column is the same as the projected-cast column. The recommended **extractor form** of §4 (`ORDER BY eql_v2.ore_block_u64_8_256(value)`) is structurally distinct from `(value)::jsonb`, so projecting `value::jsonb` alongside is safe — the sort key matches the functional index either way and the cast just runs on the LIMIT'd rows.
+
+The bench suite at `cipherstash/benches` ships a custom `EqlV2Encrypted` sqlx type and projects `value` raw across every scenario for exactly this reason: a future scenario that adds `ORDER BY value` (or `ORDER BY (value -> 'sel')` at the field level) can't accidentally walk into the trap.
+
+(Why does PostgreSQL push the cast into the scan? It's a projection-pushdown optimisation: computing `value::jsonb` early narrows the rows fed into Sort/Hash/Materialize nodes — `1100 → 36` bytes per row in the bench's ORE-encoded `i32` case. That's normally a big win; the cost model just doesn't account for "preserving sort-key matchability against an indexed expression." The optimisation is right in expectation and wrong in this specific case.)
```
yujiyokoo: I am not sure if this is the writing style you're going for but it looks a bit strange to me with monologue inside paretheses

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+  ON orders USING GIN (eql_v2.jsonb_array(data_encrypted));
+ANALYZE orders;
+
+-- Document-level containment
+SELECT * FROM orders WHERE data_encrypted @> $1::jsonb;
+-- Bitmap Index Scan on orders_data_gin_idx
+--   Index Cond: (eql_v2.jsonb_array(data_encrypted) @> eql_v2.jsonb_array('...'::eql_v2_encrypted))
+```
+
+For field-level lookups (`data_encrypted->'email' = $1`), use the per-selector hash recipe for hot paths and the all-selector GIN recipe for everything else. Both are described in §1; both require the extractor form because the selector isn't part of any native SQL operator.
```
yujiyokoo: Instead of "described in §1", I suggest "listed in §1"

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+
+`eql_v2.jsonb_path_query`, `_first`, and `_exists` are inlinable SQL functions that walk the `sv` array filtering by selector. Use them when you need the full sub-payload back; use `eql_v2.hmac_256(col, '<selector>')` when you only need an equality check on the selector's value.
+
+---
+
+## 8. A short list of common pitfalls
+
+- **Index created before data was populated through the proxy.** EQL search-config + functional index is a two-phase process: configure the index, repopulate the column through the proxy so the encrypted terms land in the payload, *then* `CREATE INDEX … ANALYZE`. The other order silently leaves the index without the values it needs.
```
yujiyokoo: I am not all that clear on this but does this only apply for proxy?

docs/reference/query-performance.md
```diff
@@ -0,0 +1,453 @@
...
+```sql
+CREATE INDEX … USING btree (eql_v2.hmac_256(col));   -- large tables
+CREATE INDEX … USING hash  (eql_v2.hmac_256(col));   -- small / medium tables
+```
+
+Measured: a `hash` functional index on a 10M-row encrypted-JSONB column ran for 17 hours to 73% and then stalled, the insert rate decaying toward zero — it never completed. The `btree` equivalent, with `maintenance_work_mem` raised, builds without drama. Hash is not wrong — it is a fine choice up to mid-six-figure row counts — but its *build* does not scale, so reserve it for tables that won't grow into the millions.
+
+### The de-TOAST floor
+
+A functional index over a large encrypted column de-TOASTs the whole stored value once per row to evaluate the extractor — and an ste_vec JSONB document is large. This cost is unavoidable and identical across access methods: a GIN, btree, or hash build on the same column all pay it. It sets the build's *floor* rate; the memory and index-type choices above decide whether you stay near that floor or fall far below it. (There is no partial de-TOAST for JSONB — `col -> 'selector'` materialises the entire document.)
```
yujiyokoo: Not sure what the intended audience is like, but maybe providing a link to de-TOAST or TOAST may help some

yujiyokoo (approved):
I have read through, and added comments where I noticed something. I did not fact-check or check the relevant code

docs/reference/query-performance.md
coderdan: Done in 7e65bcf — added a note under the intro: the guide applies to EQL 2.3 and later, where the `eql_v2_encrypted` operators became inlinable SQL functions.

docs/reference/query-performance.md
coderdan: Done in 5ed0658 — bolded that **functional indexes are the recommended path for new schemas**.

docs/reference/query-performance.md
coderdan: Done in 5ed0658 — the bullet now reads "Small leaves (space efficiency)".

docs/reference/query-performance.md
coderdan: Good point — reworded in 7e65bcf to: "`SET search_path = …` is the usual culprit, but *any* `SET` clause blocks inlining."

docs/reference/query-performance.md
coderdan: Fixed in 5ed0658.

docs/reference/query-performance.md
coderdan: Added an intro in 5ed0658. I kept it count-neutral ("a few ways… they differ only in how explicit you make the extractor") rather than "two", since the section also documents a Hybrid form alongside Natural and Extractor.

docs/reference/query-performance.md
coderdan: Added your framing as the section intro in 5ed0658 — WHERE and ORDER BY reach a functional index through different paths. Thanks!

docs/reference/query-performance.md
coderdan: Removed the parentheses in 5ed0658.

docs/reference/query-performance.md
coderdan: Fixed in 7e65bcf — now "Two query shapes".

docs/reference/query-performance.md
coderdan: Fixed in 7e65bcf — WHERE and ORDER BY now use `value`. I also switched the two fix-option examples below to `value` in 5ed0658 so the whole trap section is consistent.

coderdan: Thanks for the cross-check — fixed.

docs/reference/query-performance.md
coderdan: Unwrapped in 5ed0658 — it is now a normal paragraph ("Why does PostgreSQL push the cast into the scan? …") instead of a parenthetical aside.

docs/reference/query-performance.md
coderdan: Done in 5ed0658.

docs/reference/query-performance.md
coderdan: Good catch — it is not proxy-only. The two-phase repopulation applies to any client that writes encrypted terms into the payload, so I reworded it in 7e65bcf/5ed0658 to name Proxy or Stack with links to both repos.

docs/reference/query-performance.md
coderdan: Added a link to the Postgres TOAST storage docs on "de-TOASTs" in 5ed0658.

docs/reference/query-performance.md
coderdan: Weirdly they are required for this form. A quirk of the grammar I guess.

docs/reference/query-performance.md
coderdan: I would assume technical readers and agents mostly! Added the link.

PR #213, opened 2026-05-14
feat(config): optional ste_vec mode field with constraint validation

freshtonic:
Adds an optional `mode` field on `ste_vec` index configs, validated by a new `eql_v2.config_check_ste_vec_mode` CHECK on `eql_v2_configuration`. Allowed values are `standard` and `compat`; absent passes unconditionally so existing configurations remain valid.

## Summary by CodeRabbit

* **New Features**
  * Configuration indexes now support an optional `mode` field for `ste_vec` index settings, supporting either "standard" or "compat" values. The system validates all configurations and rejects invalid mode values with informative error messages indicating the complete set of allowed options. Validation behavior is verified across all scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/encrypt-query-language/pull/213)


tests/sqlx/tests/config_tests.rs
```diff
@@ -774,3 +774,104 @@ async fn configuration_constraint_validation(pool: PgPool) -> Result<()> {
 
     Ok(())
 }
+
+#[sqlx::test(fixtures(path = "../fixtures", scripts("config_tables")))]
+async fn ste_vec_mode_validation(pool: PgPool) -> Result<()> {
```
coderdan: This test might be easier to read/debug if split out into 3(?) tests for the different cases.

coderdan (approved):
A suggestion for test structure but otherwise LGTM

PR #217, opened 2026-05-18
fix(operators): correct planner selectivity functions on <=, >, >=

### LINKED ISSUE #216: Range operators on eql_v2_encrypted use incorrect selectivity functions (scalarltsel on > and >=)
## Summary

`src/operators/>.sql` and `src/operators/>=.sql` declare every `CREATE OPERATOR` with `RESTRICT = scalarltsel, JOIN = scalarltjoinsel`. Those are the "less-than" selectivity estimators. PostgreSQL ships symmetric estimators that should be used instead:

| Operator | Should be |
|----------|-----------|
| `>` | `RESTRICT = scalargtsel, JOIN = scalargtjoinsel` |
| `>=` | `RESTRICT = scalargesel, JOIN = scalargejoinsel` |

The inner `ore_block_u64_8_256` operators on the underlying domain type *are* declared correctly (`src/ore_block_u64_8_256/operators.sql` uses `scalargtsel` / `scalargejoinsel`), so the inconsistency is only at the `eql_v2_encrypted` outer layer.

## Impact

Wrong selectivity estimates feed the planner inaccurate row counts for `>` and `>=` predicates. Symptoms:

- Index-vs-seq-scan flips on the wrong threshold for range queries.
- Join order chosen on bad cardinality assumptions.
- `ORDER BY` + `LIMIT` plans can pick the wrong access path when the planner thinks `WHERE col > $1` is selective when it isn't (or vice versa).

The bug is silent until cardinality matters. With the range-operator inlining in #211, bare-form range queries now structurally match functional ORE indexes, which makes selectivity hints load-bearing for plan choice — so the cost of leaving this wrong is higher than it was pre-2.3.

## Scope

Three `CREATE OPERATOR` blocks in each of `src/operators/>.sql` and `src/operators/>=.sql` (six total), plus regenerated `release/` artefacts. No behaviour change for individual query results — only planner estimates. No upgrade note needed.

## Detected by

CodeRabbit on #211 review. PR scope was operator inlining, so this was deferred.

## Verification

Compare `EXPLAIN ANALYZE` plans on a 100k-row Block ORE column for `WHERE col > $1` (high-selectivity vs. low-selectivity bounds) before/after the change; expected row estimates should track actual within a constant factor.

coderdan:
## Summary

Range operators on `eql_v2_encrypted` declared the wrong `RESTRICT` / `JOIN` selectivity functions. All three overloads of `<=`, `>`, and `>=` used `scalarltsel` / `scalarltjoinsel` — the "less-than" estimators. The inner `eql_v2.ore_block_u64_8_256` `>=` block had a related miss (`scalarlesel` where `scalargesel` belongs).

Closes #216.

### What changes

| File | Operator | Was | Now |
|---|---|---|---|
| `src/operators/<=.sql` | `<=` (×3 overloads) | `scalarltsel` | `scalarlesel` |
| `src/operators/>.sql` | `>` (×3 overloads) | `scalarltsel` | `scalargtsel` |
| `src/operators/>=.sql` | `>=` (×3 overloads) | `scalarltsel` | `scalargesel` |
| `src/ore_block_u64_8_256/operators.sql` | inner `>=` | `scalarlesel` | `scalargesel` |

`<` and the rest of the inner ORE block operators were already correct.

## What RESTRICT and JOIN do

`RESTRICT` and `JOIN` tell the planner **which estimator function to use** to compute selectivity for a given operator — i.e. what fraction of rows survive the predicate.

- `RESTRICT` is called when the operator appears as `col <op> constant` (WHERE-clause filter).
- `JOIN` is called when it appears as `A.col <op> B.col` (join condition).

The planner uses those estimates to choose plans: index vs. seq scan, hash vs. nested-loop join, drive side of a join, and so on. Wrong estimates don't change query *results* — they change *plan choice*, which is where the perf hits live.

Postgres ships matched pairs: `scalarltsel`/`scalarltjoinsel` model `<`, `scalarlesel`/`scalarlejoinsel` model `<=`, `scalargtsel`/`scalargtjoinsel` model `>`, `scalargesel`/`scalargejoinsel` model `>=`. The two halves (RESTRICT and JOIN) should always be the matching pair — they have to agree on direction and boundary semantics.

The pre-fix declaration used `scalarltsel` for `>` and `>=`. That doesn't compute the complement of the right answer — it computes the **same-direction** answer. So a `>` predicate was being modeled as if it filtered to rows *below* the literal.

## Worked example

`events.encrypted_ts` is a Block ORE column with 100k rows uniformly distributed over a year. Query: `WHERE encrypted_ts > $1` where `$1` is mid-March.

| | Selectivity estimate | Planner decision |
|---|---|---|
| True | ~0.79 (Mar–Dec) | seq scan + filter (~80k rows survive, streaming is faster than 80k random fetches) |
| `scalargtsel` (correct) | ~0.79 | seq scan ✓ |
| `scalarltsel` (buggy) | ~0.21 | index scan — looks competitive at 21k rows, but reality is 80k random page fetches |

Wrong-direction estimate → wrong access path → potentially 5–10× slower for that query. Same wrong-direction logic applies to join ordering and hash-vs-nested-loop choice.

## Why this surfaced now

Latent on `main` because the previous plpgsql operator bodies couldn't be inlined — the planner couldn't structurally match a functional ORE index for these operators regardless of selectivity, so it fell back to seq scan and the estimator hint didn't matter. #211 made the operators inlinable and the planner *can* now match the functional index, at which point bad selectivity starts flipping plan choice. CodeRabbit caught the `>` / `>=` half during the #211 review; the broader audit during the fix turned up `<=` and the inner `>=` as well.

May explain some recent benchmark weirdness.


src/operators/<=.sql
```diff
@@ -34,64 +25,69 @@ $$ LANGUAGE plpgsql;
...
 CREATE OPERATOR <=(
   FUNCTION = eql_v2."<=",
   LEFTARG = eql_v2_encrypted,
   RIGHTARG = eql_v2_encrypted,
   COMMUTATOR = >=,
   NEGATOR = >,
-  RESTRICT = scalarltsel,
-  JOIN = scalarltjoinsel
+  RESTRICT = scalarlesel,
+  JOIN = scalarlejoinsel
```
freshtonic, line 65: Took me far too long to spot what changed here!

freshtonic (approved):

src/operators/<=.sql
coderdan: Haha yes indeed. I'd missed it before as well!

PR #219, opened 2026-05-18
feat(eql_v2)!: collapse ste_vec ORE terms to single oc field

coderdan:
Closes #207.

## Summary

The pre-2.3 `ocf` (fixed-width, numeric) and `ocv` (variable-width, text) ORE-CLLW split on sv elements collapses to a single `oc` field. Width information that used to be carried by the field name is now part of the plaintext via a leading domain-tag byte (`0x00` numeric, `0x01` string) and encrypted into the CLLW ciphertext. Walking the CLLW per-byte protocol over `[tag][ciphertext]` preserves cross-domain separation (numeric < string) and within-domain ordering — so a single column can hold mixed numeric and string sv elements with a consistent total order.

**OPE removed from `eql_v2_encrypted`.** Per recent product decision, OPE-based ordered comparison moves to a separate encrypted column type in a future release; `eql_v2_encrypted` now exclusively handles ORE for ordered comparisons. v2.2 didn't emit `opf`/`opv` from production `@cipherstash/protect`, so this isn't a customer-facing data migration — it's a tightening of the documented payload surface.

## What changes

**SQL source**

- `src/ore_cllw_u64_8/` + `src/ore_cllw_var_8/` → `src/ore_cllw/` (single composite type, single extractor, single compare)
- `eql_v2.ore_cllw(jsonb)` extractor: **inlinable single-statement SQL.** Returns composite with `bytes IS NULL` rather than raising on missing `oc` — required for inlinability. Allowlisted in `tasks/pin_search_path.sql`.
- `eql_v2.compare_ore_cllw` stays plpgsql: the CLLW per-byte protocol (`y + 1 == x` mod 256) can't be expressed as a single SELECT.
- Function families collapse: `ore_cllw_u64_8` / `ore_cllw_var_8` → `ore_cllw`; same for `has_*` and `compare_*`.
- `eql_v2.compare()` priority list shortens: `ob → oc → hm → literal`.
- OPE infra removed: `src/ope_cllw_*/` directories deleted; `eql_v2.order_by_ope` removed; `src/operators/sort.sql` drops the entire `'ope'` strategy + `_compare_ope_key` helper + `ope_keys` plumbing.
- `eql_v2.hash_encrypted` flipped to inlinable single-statement SQL with data-hash fallback (required for ANALYZE to succeed on sv-shaped payloads).

**Schema** (`docs/reference/schema/eql-payload-v2.3.schema.json`)

[...]

PR #221, opened 2026-05-19
feat(eql_v2): btree operator class on eql_v2.ore_cllw

coderdan:
**Stacked on #219.** Closes #220.

## Summary

Restores the btree operator class on `eql_v2.ore_cllw` so functional indexes engage for sv-element ordered queries. The pre-2025-06-24 opclass on the per-subtype CLLW types was disabled in [`69afdc8`](https://github.com/cipherstash/encrypt-query-language/commit/69afdc8) and never came back; the consolidation in #219 collapsed those types into `eql_v2.ore_cllw` and this PR adds a clean opclass on the consolidated type.

The gap was visible end-to-end in the bench data on [cipherstash/benches#14](https://github.com/cipherstash/benches/pull/14): `json/field_order/functional` at 1M took 20 s (seq scan + Sort). With this opclass, `EXPLAIN` flips to `Index Scan + Limit` — expected sub-ms.

## What this adds

- **`src/ore_cllw/operators.sql`** — same-type comparison operators (`<`, `<=`, `=`, `>=`, `>`, `<>`) on `eql_v2.ore_cllw`. Each backed by a single-statement `LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE` wrapper reducing to `eql_v2.compare_ore_cllw_term(a, b) <op> 0`. Wrappers inline so the planner can fold them into the calling query — that's what lets the index match.
- **`src/ore_cllw/operator_class.sql`** — `eql_v2.ore_cllw_ops` btree opclass, `DEFAULT FOR TYPE eql_v2.ore_cllw USING btree`. FUNCTION 1 is `eql_v2.compare_ore_cllw_term` directly (plpgsql per-byte protocol; called only by btree internals, not per-row from the calling query). Excluded from the Supabase variant via the existing `**/*operator_class.sql` glob in `tasks/build.sh`.
- **`tasks/pin_search_path.sql`** — allowlists the six operator backing functions. Pinning would break the inlining chain.
- **`tests/sqlx/tests/ore_cllw_opclass_tests.rs`** — operator wiring, cross-domain ordering via the tag byte, opclass-is-default-for-type, functional-index match (EXPLAIN shows Index Scan, no Sort), inlinability lint.

## Design notes

- **No `HASHES` / `MERGES` flags** on the operator declarations. HASHES requires a registered hash function on the type (none, and we don't want one — CLLW is for ordering). MERGES requires a corresponding operator family registered on both sides. This was the gap [#220](https://github.com/cipherstash/encrypt-query-language/issues/220)'s history identified as breaking the pre-2025-06-24 opclass.
- **Equality via `compare_ore_cllw_term = 0`**, not a `bytea_eq` shortcut. One source of truth for equality semantics; resilient to any future change in ciphertext encoding.
- **Different scope from the operators in #211.** Those (`<`, `<=`, `>`, `>=` on `eql_v2_encrypted`) inline to `ore_block_u64_8_256` and raise on non-Block-ORE columns. The new operators here live on the `eql_v2.ore_cllw` composite type — what callers reach through the extractor form. No conflict, different scope.

## Test plan

- [x] `mise run build` succeeds for all three variants (main, supabase, protect). `cipherstash-encrypt.sql` and `cipherstash-encrypt-protect.sql` contain `CREATE OPERATOR CLASS eql_v2.ore_cllw_ops`; `cipherstash-encrypt-supabase.sql` does not.
- [x] `mise run test` green — 6 new opclass tests + 6 existing ORE-CLLW tests pass.
- [x] `pg_opclass.opcdefault = true` for `eql_v2.ore_cllw_ops`.
- [x] Functional btree on `eql_v2.ore_cllw(value)` engages `Index Scan` for `ORDER BY ... LIMIT n` (verified via `EXPLAIN` with `enable_seqscan=off` on a 20-row fixture).
- [x] Re-run JSON `field_order/*` bench at 1M after merge — expected sub-ms via the new opclass.

## Merge order

This branch is stacked on #219. After #219 merges to main, rebase this onto main and flip to ready-for-review.

## Summary by CodeRabbit

* **New Features**
  * B-tree operator class added to enable functional-index matching for ordered encrypted (ORE) queries.

* **Documentation**
  * Updated index and upgrade docs to reflect functional-index behavior and removed an outdated limitation note.

* **Bug Fixes**
  * Extractor/comparator null-handling tightened to surface invalid values and ensure consistent NULL semantics for indexing.

* **Tests**
  * Added integration tests covering operator semantics, index registration, and planner/index selection.

* **Chores**
  * Build and post-install scripts hardened; inlining helper functions and test allowlists updated to preserve planner inlining.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/encrypt-query-language/pull/221?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


freshtonic:
I noticed that there are errors when mise build generates the concatenated SQL files in the EQL repo. Looks like it’s still referencing the deleted files (on main anyway)

```
cat: src/ore_cllw_u64_8/types.sql: No such file or directory
cat: src/ore_cllw_u64_8/functions.sql: No such file or directory
cat: src/blake3/types.sql: No such file or directory
cat: src/blake3/functions.sql: No such file or directory
cat: src/ore_cllw_var_8/types.sql: No such file or directory
cat: src/ore_cllw_var_8/functions.sql: No such file or directory
cat: src/ope_cllw_u64_65/types.sql: No such file or directory
cat: src/ope_cllw_u64_65/functions.sql: No such file or directory
cat: src/ope_cllw_u64_65/compare.sql: No such file or directory
cat: src/ope_cllw_var_8/types.sql: No such file or directory
cat: src/ope_cllw_var_8/functions.sql: No such file or directory
cat: src/ope_cllw_var_8/compare.sql: No such file or directory
cat: src/ore_cllw_u64_8/operators.sql: No such file or directory
cat: src/blake3/compare.sql: No such file or directory
cat: src/ore_cllw_var_8/compare.sql: No such file or directory
cat: src/ore_cllw_u64_8/compare.sql: No such file or directory
```

Also, the build script should have failed but it seems to be swallowing errors.

freshtonic (approved):
Approved, but conditional on whether the following things are not real concerns:

```
 Concerns

  1. compare_ore_cllw_term can return NULL — and it's now btree FUNCTION 1

  src/ore_cllw/functions.sql:200-204 returns NULL when a.bytes IS NULL OR b.bytes IS NULL. This is reachable in practice: the (jsonb) overload of eql_v2.ore_cllw
  (functions.sql:53-58) returns (bytes => NULL) when the input lacks oc, and the comment there says it does so deliberately for inlinability. PostgreSQL's btree FUNCTION 1
   contract requires a non-NULL int for non-NULL composite inputs; NULL-at-the-composite-level is filtered by btree's null-handling layer, but non-NULL composite with NULL
   field is not.

  In a functional index built via eql_v2.ore_cllw((value).data), any row whose payload lacks oc would index a ROW(NULL)::eql_v2.ore_cllw value, and a subsequent
  WHERE/ORDER BY traversal would call FUNCTION 1 with that as one operand. Behavior is undefined (likely silent misordering rather than a crash).

  The test fixture seeds 20 rows all with oc present, so this path isn't exercised. Two suggestions, pick one:
  - Add a fixture row with oc absent (or with bytes=NULL in the hand-crafted form) and assert either a graceful query result or an expected raise.
  - Document the expectation in the opclass comment that callers must filter / partial-index out rows lacking oc, and recommend CREATE INDEX ... WHERE (value).data ? 'oc'
  in docs/reference/database-indexes.md.

  2. Tests cover ORDER BY but not WHERE <op> index match

  The PR description and CHANGELOG both claim WHERE eql_v2.ore_cllw(col) <op> eql_v2.ore_cllw($1::jsonb) engages Index Cond — but functional_index_engages_for_order_by
  only verifies the ORDER BY ... LIMIT n shape. The WHERE form is the more common usage and a different planner path (range-quals against opclass strategies 1/2/4/5 vs.
  strategy 3 equality). A parallel test asserting Index Cond on WHERE eql_v2.ore_cllw((value).data) < eql_v2.ore_cllw($1::jsonb) would close the gap and match the claims.
```

PR #224, opened 2026-05-20
feat: fixture + bench data-generation foundation (int4 fixture, Proxy scaffolding)

tobyhede:
Foundation layer for the encrypted-domain stack. Adds the data-generation infrastructure (bench tooling + Proxy-backed fixture generators) needed by the int4 variant-family work that lands on top of this PR.

Extracted from #210 (the encrypted-domain-types-design spike). #210 remains as the full spike; this PR + the stacked int4 PR are the cleaner re-derivation for production review.

## What's in this PR

| Area | Files | Notes |
|---|---|---|
| Bench dataset tooling | `tests/benchmarks/` (docker-compose, schema, generate.sh, README) + `tasks/bench.toml` + `tests/sqlx/tests/bench_data_tests.rs` | 100K-row encrypted dataset generator; runs Proxy in a sidecar container. |
| Bench Proxy fix | `tests/benchmarks/generate.sh` (`docker restart` after schema apply) | Without the restart, Proxy serves the bench dataset in passthrough mode — no encryption applied. Mirrors the pattern already used by the fixture generators. |
| Proxy task scaffolding | `tasks/fixtures.toml` (`proxy:up` / `proxy:down` / `proxy:logs`) + `tests/docker-compose.proxy.yml` | Reused by every fixture generator. |
| Shared fixture helpers | `tasks/fixtures/_generate_common.sh` | `restart_proxy_and_wait` + `dump_fixture_table`. |
| Encrypted int4 fixture | `tasks/fixtures/generate_encrypted_int4.sh`, `tasks/fixtures/encrypted_int4_schema.sql`, `tests/sqlx/migrations/009_install_encrypted_int4_fixture.sql` | 14 rows (-100…9999) generated through real Proxy. Plain `payload JSONB` column — domain casts happen per-query in the consumer (the stacked int4 PR). |

## Out of scope

- No domain types. `eql_v2_int4*`, `encrypted_text`, `encrypted_jsonb` all live in the stacked PRs.
- No fixture-consuming tests. The fixture migration installs cleanly on its own; the test file that consumes it (`encrypted_int4_ord_ore_tests.rs`) lands in the int4 PR where its domain dependency exists.
- Text and jsonb fixtures: not included. They're sibling work that lives in #210 and isn't part of this stack.

## Test plan

- [x] `mise run build` — clean, no domain code to concatenate
- [x] `mise run reset && mise run test` — full SQLx suite green (migration 009 installs without errors; existing bench/test files compile and pass)
[...]

PR #225, opened 2026-05-20
feat: eql_v2_int4 variant family

tobyhede:
Four capability-encoded jsonb-backed domains for encrypted `int4` columns. The variant name fixes the operator surface and indexing. All live in `public` and survive `eql_v2` uninstall.

Base: #224 (fixtures).

## Variants

| Domain | Operators | Payload | Index |
|---|---|---|---|
| `eql_v2_int4` | none — all raise | `c` | none |
| `eql_v2_int4_eq` | `=` `<>` | `c` `hm` | `btree (eql_v2.hmac_256(col::jsonb))` |
| `eql_v2_int4_ord` | `=` `<>` `<` `<=` `>` `>=` | `c` `ob` | `btree (eql_v2.ord(col))` |
| `eql_v2_int4_ord_ore` | same as `_ord` | `c` `ob` | `btree (eql_v2.ord(col))` |

`eql_v2_int4_ord` is the name to use; `_ord_ore` is the scheme-explicit equivalent (identical surface).

## How ordered indexing works

- `eql_v2.ord(col)` extracts the internal `eql_v2.ore_block_u64_8_256` composite.
- That composite carries `main`'s `DEFAULT` btree operator class → plain functional index, no opclass annotation.
- **No operator class on any domain type.** The index stores the ORE term, not the whole jsonb payload — no btree-entry-size blowup.
- One index serves equality *and* range: `eql_v2.ord` is lossless on full-domain `int4`, so the order term is also the equality term. That's why ordered variants carry no `hm`.
- Ordering: `ORDER BY eql_v2.ord(col)`.

## Changed from the earlier 5-variant version (review feedback)

- `eql_v2_int4_ct` → renamed `eql_v2_int4` (plain name = storage-only).
- `eql_v2_int4_ord_ope` **removed** — #219 deleted OPE-CLLW; its foundation is gone. OPE returns later as a separate column type.
- Operator-class-on-domain approach dropped → functional index on the composite opclass.
- Dropped redundant `hm` from ordered variants and `::jsonb` casts from doc examples.

## Caveats

- **Supabase:** the ORE composite opclass is stripped from the Supabase build → ordered `int4` columns get **no indexed range on Supabase** (seq-scan). No `_ord_ope` fallback anymore.
- `eql_v2_int4_eq` index needs the explicit `::jsonb` cast — `eql_v2.hmac_256` is both a function and a type name. Disambiguating that collision is a planned follow-up.

## Docs

`docs/upgrading/v2.4.md` (U-001), `docs/reference/encrypted-int4-domain.md`, `encrypted-int4-walkthrough.md`.

## Test plan

- [ ] `mise run test --postgres 14|15|16|17`
- [ ] `mise run docs:validate` — 0 errors
- [ ] Build: composite opclass present in main release, absent from Supabase release
- [ ] Per-variant Rust/SQLx suites; `Index Scan` asserted for equality + range on ordered variants


docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+|------------------------------|--------------------------------|------------------------|------------------|
+| `public.eql_v2_int4_ct`      | none (all blockers)            | `c` only               | n/a              |
+| `public.eql_v2_int4_eq`      | `=`, `<>`                      | `hm`                   | functional btree on `((eql_v2.hmac_256(col::jsonb)))` for `=`; `<>` seq-scan |
+| `public.eql_v2_int4_ord_ore` | `=`, `<>`, `<`, `<=`, `>`, `>=` | `hm`, `ob`             | hmac functional btree (equality); btree operator class for range — name it explicitly, excluded from Supabase build |
+| `public.eql_v2_int4_ord_ope` | `=`, `<>`, `<`, `<=`, `>`, `>=` | `hm`, `opf`            | functional btree on `((eql_v2.eql_v2_int4_ord_ope_ope_key(col::jsonb)))` |
+| `public.eql_v2_int4`         | same as `_ord_ore`             | `hm`, `ob`             | same as `_ord_ore` |
+
+## Per-variant detail
+
+### `eql_v2_int4_ct` — storage only
```
coderdan: I'm not sure about the `_ct` suffix - it doesn't mean anything to most people. What about `_blob` or `_opaque`?

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+```sql
+CREATE INDEX users_age_hmac_idx ON users
+USING btree ((eql_v2.hmac_256(age::jsonb)));
+```
+
+`=` engages the index; `<>` is seq-scan (btree supports only equality).
+
+### `eql_v2_int4_ord_ore` and the default `eql_v2_int4` — HMAC + ORE-block
+
+Payload terms: `hm`, `ob`. Operators: `=`, `<>`, `<`, `<=`, `>`, `>=`.
```
coderdan: `hm` isn't needed for ints because ORE on int is lossless.

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+
+Recommended indexes:
+
+```sql
+-- Equality: functional btree on the hmac extractor.
+CREATE INDEX orders_amount_hmac_idx ON orders
+USING btree ((eql_v2.hmac_256(amount::jsonb)));
```
coderdan: Why is the cast to `::jsonb` needed? Wouldn't the type be `eql_v2_int4`?

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+
+-- Range: btree operator class — name it explicitly. A bare
+-- USING btree (amount) resolves to jsonb_ops (the domain base type's
+-- default) and will not serve ORE range.
+CREATE INDEX orders_amount_ore_idx ON orders
+USING btree (amount eql_v2.eql_v2_int4_ord_ore_operator_class);
```
coderdan: This approach probably isn't as bad for `int4` as it is for other types but it means that the whole object ends up in the index which can hit the b-tree entry size limit and hurt performance.

I also find the naming rather verbose.

A functional index could solve both problems:

```
CREATE INDEX orders_amount_ore_idx ON orders
USING btree (eql_v2.ord(amount));
```

`ord` says semantically what its actually doing and it can be overloaded based on the type to return a specific composite type which has the operator class defined.

`eql_v2.ord` `eql_v2.eq` etc become an interface that can be implemented for the various domain types. Operator classes should only be defined on internal _composite_ types (this avoids the "operator fallthrough" problem with domain types as well).

See https://github.com/cipherstash/encrypt-query-language/pull/212/changes

docs/reference/encrypted-int4-domain.md
tobyhede: `blob` already exists in postgresql

It may be surprising if `eql_v2_int4` is just an alias for `eql_v2_int4_eq` or `eql_v2_int4_ord_ore`

I was wondering if we make the ciphertext the default case - `eql_v2_int4` is encrypted only.

`eql_v2_int4_eq` is then encrypted PLUS `eq` ... seems much clearer to explain to me?

docs/reference/encrypted-int4-domain.md
tobyhede: @coderdan Yeah, I haven't pulled in the work you have been doing.
This work will need to be synced and aligned.

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+Range queries are served by the btree operator class, not a functional
+index: `compare_ore_block_u64_8_256` is PL/pgSQL, so the range wrappers
+are deliberately non-inlinable (`plpgsql IMMUTABLE`) and the planner
+matches the surviving `<` / `<=` / `>` / `>=` operator nodes to the
+operator class. This mirrors how the core `eql_v2_encrypted` type
+indexes ORE.
+
+Supabase caveat: operator classes are stripped from the EQL Supabase
+build, so `_ord_ore` range falls back to seq-scan there — use
+`eql_v2_int4_ord_ope` for indexed range on Supabase.
```
coderdan: The functional index approach (mentioned above) completely eliminates that problem.

The public domain types should never have operator classes on them. The `eql_v2.ord(x eql_v2_int4_ord_ope)` can return `text` (hex string) which uses the built in operator class.

Example:

```sql
CREATE FUNCTION eql_v2.ord(val eql_v2_int_ord_ope)
RETURNS eql_v2.hexstring -- domain type for text (built-in op-class)
LANGUAGE SQL AS $$
  val->'op'
$$;

CREATE FUNCTION eql_v2.ord(val eql_v2_int)
RETURNS eql_v2.ore_block -- composite type with custom op-class
LANGUAGE plpgsql AS $$
  BEGIN
    -- fancy ORE comparison
    RETURN -1; -- or 0 or 1
  END;
$$;
```

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+
+ORDER BY caveat: `ORDER BY col` on a jsonb-backed domain follows
+native jsonb comparison, not ORE order — `ORDER BY` requests the
+column type's default sort order, which for a domain resolves to the
+base type, not the operator class. Sort by the extractor expression:
+
+```sql
+ORDER BY eql_v2.ore_block_u64_8_256(col::jsonb)
```
coderdan: Using the interface idea above this would become:

```sql
ORDER BY eql_v2.ord(col)
```

docs/reference/encrypted-int4-domain.md
tobyhede: Not required

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+
+### `eql_v2_int4_ord_ope` — HMAC + OPE-direct
+
+Payload terms: `hm`, `opf`. Operators: `=`, `<>`, `<`, `<=`, `>`, `>=`.
+
+Recommended indexes:
+
+```sql
+CREATE INDEX prices_hmac_idx ON prices
+USING btree ((eql_v2.hmac_256(price::jsonb)));
```
coderdan: Interface would become:

```sql
CREATE INDEX prices_eq_idx ON prices
USING hash ((eql_v2.eq(price)));
```

Although, you'd really only need `ord` and a btree for ints.

CHANGELOG.md
```diff
@@ -20,7 +20,11 @@ Each entry that ships in a published release links to the PR that introduced it.
...
+Targeting `2.4.0`. See [`docs/upgrading/v2.4.md`](docs/upgrading/v2.4.md) for the consolidated upgrade notes for the new `eql_v2_int4` variant family; entries below the new one are carry-overs targeting `2.3.0` (see [`docs/upgrading/v2.3.md`](docs/upgrading/v2.3.md)).
+
+### Added
+
+- **`eql_v2_int4` variant family — five capability-encoded domain types for encrypted `int4` columns.** Pick the variant whose operator surface matches the index terms your column carries: `eql_v2_int4_ct` (storage only, no operators), `eql_v2_int4_eq` (HMAC equality only — `=`, `<>`), `eql_v2_int4_ord_ore` (HMAC equality + ORE ordering — adds `<`, `<=`, `>`, `>=`, range indexed by a btree operator class named explicitly in `CREATE INDEX`; the operator class is excluded from the Supabase build), `eql_v2_int4_ord_ope` (HMAC equality + OPE-direct ordering — same operator surface as `_ord_ore` with range engaged by a functional btree, Supabase-compatible), or `eql_v2_int4` (default, behaves as `_ord_ore`). All variants live in `public` and survive `eql_v2` uninstall. Per-variant payload requirements and index recipes: [U-001](docs/upgrading/v2.4.md#u-001-eql_v2_int4-variant-family).
```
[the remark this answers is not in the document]
tobyhede: Added the `([#225])` PR link to the changelog entry in 78f82c2.

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Variant comparison
+
+| Domain                       | Operators supported            | Payload terms required | Index-engagement |
+|------------------------------|--------------------------------|------------------------|------------------|
+| `public.eql_v2_int4_ct`      | none (all blockers)            | `c` only               | n/a              |
+| `public.eql_v2_int4_eq`      | `=`, `<>`                      | `hm`                   | functional btree on `((eql_v2.hmac_256(col::jsonb)))` for `=`; `<>` seq-scan |
+| `public.eql_v2_int4_ord_ore` | `=`, `<>`, `<`, `<=`, `>`, `>=` | `hm`, `ob`             | hmac functional btree (equality); btree operator class for range — name it explicitly, excluded from Supabase build |
+| `public.eql_v2_int4_ord_ope` | `=`, `<>`, `<`, `<=`, `>`, `>=` | `hm`, `opf`            | functional btree on `((eql_v2.eql_v2_int4_ord_ope_ope_key(col::jsonb)))` |
+| `public.eql_v2_int4`         | same as `_ord_ore`             | `hm`, `ob`             | same as `_ord_ore` |
```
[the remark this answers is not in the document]
tobyhede: Added the `c` payload term to the non-ct rows of the comparison table and the per-variant `Payload terms:` lines in 78f82c2.

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+inline chain through to a built-in `bytea` compare:
+
+```
+col < $1::eql_v2_int4_ord_ope
+  └─ eql_v2.eql_v2_int4_ord_ope_lt(eql_v2_int4_ord_ope, eql_v2_int4_ord_ope) [SQL IMMUTABLE]
+       └─ eql_v2.eql_v2_int4_ord_ope_ope_key(col) < eql_v2.eql_v2_int4_ord_ope_ope_key($1)
+            └─ (eql_v2.ope_cllw_u64_65(col::jsonb)).bytes < ... [bytea built-in]
+                 ⇒ planner matches functional btree on
+                   ((eql_v2.eql_v2_int4_ord_ope_ope_key(col::jsonb)))
+```
```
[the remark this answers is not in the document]
tobyhede: Added `text` language tags to the bare fenced code blocks in 78f82c2.

docs/reference/encrypted-int4-domain.md
```diff
@@ -0,0 +1,173 @@
...
+inlining is irrelevant to that path. This is the same mechanism the
+core `eql_v2_encrypted` type uses for ORE.
+
+## File layout
+
+```
+src/encrypted_domain/
+  types.sql                    # five domain declarations (CT/EQ/ORD_ORE/ORD_OPE/default)
+  functions.sql                # shared blocker helper; text + jsonb wrappers
+  operators.sql                # CREATE OPERATOR for text + jsonb
```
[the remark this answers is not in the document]
tobyhede: Corrected the file-layout note in 78f82c2 — `functions.sql` is now described as the shared blocker helper, the non-existent `operators.sql` line removed, and the two `*_operator_class.sql` files added.

docs/reference/encrypted-int4-walkthrough.md
```diff
@@ -0,0 +1,292 @@
...
+
+```sql
+CREATE INDEX users_age_hmac_idx
+  ON users USING btree ((eql_v2.hmac_256(age::jsonb)));
+```
+
+```
+ Index Scan using users_age_hmac_idx on users  (cost=0.28..8.30 rows=1 width=…)
+   Index Cond: (eql_v2.hmac_256((age)::jsonb) = eql_v2.hmac_256('{"hm":"…"}'::jsonb))
+```
```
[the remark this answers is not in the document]
tobyhede: Added `text` language tags to the bare fenced code blocks in 78f82c2.

docs/reference/encrypted-int4-walkthrough.md
```diff
@@ -0,0 +1,292 @@
...
+  ```
+  ERROR: Expected a ope_cllw_u64_65 index (opf) value in json: {"v":2,"k":"ct","i":{…},"c":"…","hm":"…"}
+  ```
+  Source: [`src/ope_cllw_u64_65/functions.sql:23`](../../src/ope_cllw_u64_65/functions.sql#L23). Same pattern for `hmac_256` (`hm` missing) and `ore_block_u64_8_256` (`ob` missing). Type system does not catch this — the domain is `CREATE DOMAIN ... AS jsonb` with no CHECK.
+
+- **Unsupported operator.** Every blocker resolves to `encrypted_domain_unsupported_bool` and raises before predicate evaluation. Example for `<` on `_eq`:
+  ```
+  ERROR: operator < is not supported for eql_v2_int4_eq
+  ```
+  Source: `eql_v2.encrypted_domain_unsupported_bool` in [`src/encrypted_domain/functions.sql`](../../src/encrypted_domain/functions.sql#L36). For `WHERE` predicates with planner-time-foldable constants this raises once at plan time; for general-case predicates it raises on the first scanned row. Either way, the error never falls through to native jsonb comparison — that's the policy.
```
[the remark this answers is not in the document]
tobyhede: Replaced the brittle `#L36` line anchor with a plain file link in 78f82c2.

docs/upgrading/v2.4.md
```diff
@@ -0,0 +1,80 @@
...
+## TL;DR
+
+1. Pick a variant for each encrypted int4 column based on the operators you need and the index terms your Proxy emits:
+   - `eql_v2_int4_ct` — storage only
+   - `eql_v2_int4_eq` — HMAC equality (`=`, `<>`)
+   - `eql_v2_int4_ord_ore` — HMAC + ORE-block ordering (range engages a btree operator class)
+   - `eql_v2_int4_ord_ope` — HMAC + OPE-direct ordering (range engages a functional btree)
+   - `eql_v2_int4` — default, identical to `_ord_ore`
+2. Declare columns with the chosen variant type.
+3. Attach the matching functional indexes (see the per-variant recipe in [`docs/reference/encrypted-int4-walkthrough.md`](../reference/encrypted-int4-walkthrough.md)).
```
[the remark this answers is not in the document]
tobyhede: Broadened the wording in 78f82c2 to cover the named btree operator class for `_ord_ore`/default range alongside functional indexes.

src/encrypted_domain/int4/int4_ct.sql
```diff
@@ -0,0 +1,459 @@
...
+
+-- =, <> (blockers, 3 shapes each)
+
+--! @brief Blocker for = on eql_v2_int4_ct.
+--! @param a eql_v2_int4_ct
+--! @param b eql_v2_int4_ct
+--! @return boolean (never returns; always raises)
+CREATE FUNCTION eql_v2.eql_v2_int4_ct_eq(a eql_v2_int4_ct, b eql_v2_int4_ct)
+RETURNS boolean IMMUTABLE STRICT PARALLEL SAFE
+AS $$ BEGIN RETURN eql_v2.encrypted_domain_unsupported_bool('eql_v2_int4_ct', '='); END; $$
```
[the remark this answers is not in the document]
tobyhede: Fixed in 78f82c2. `STRICT` removed from every blocker function across the five variant files and the shared helper, so blockers raise on NULL input rather than short-circuiting to NULL. A NULL-input regression test was added per variant. The real operator wrappers keep `STRICT`, where NULL-returning is correct SQL three-valued logic.

src/encrypted_domain/types.sql
```diff
@@ -0,0 +1,57 @@
...
+  ) THEN
+    CREATE DOMAIN public.eql_v2_int4_ord_ore AS jsonb;
+  END IF;
+
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_type
+    WHERE typname = 'eql_v2_int4_ord_ope' AND typnamespace = 'public'::regnamespace
+  ) THEN
+    CREATE DOMAIN public.eql_v2_int4_ord_ope AS jsonb;
+  END IF;
```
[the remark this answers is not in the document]
tobyhede: Added a `--! @brief` Doxygen comment above each of the five `CREATE DOMAIN` statements in 78f82c2.

tests/sqlx/tests/encrypted_int4_ct_tests.rs
```diff
@@ -0,0 +1,60 @@
...
+    for op in ["->", "->>"] {
+        for sql in [
+            format!("SELECT $1::jsonb::eql_v2_int4_ct {op} 'field'::text"),
+            format!("SELECT $1::jsonb::eql_v2_int4_ct {op} 0::integer"),
+            format!("SELECT $1::jsonb {op} $2::jsonb::eql_v2_int4_ct"),
+        ] {
+            let err = sqlx::query(&sql)
+                .bind(SAMPLE_PAYLOAD)
+                .bind(SAMPLE_PAYLOAD)
+                .fetch_one(&pool)
```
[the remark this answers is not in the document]
tobyhede: Fixed in 78f82c2 — the third path-op query now reuses `$1` and binds once. (In practice sqlx tolerated the extra bind, so the test was already passing; binding only what the query uses is still the correct form.)

docs/reference/encrypted-int4-domain.md
coderdan: Yeah I'm ok with that. When we discussed it this week it sounded like we wanted the default to include SEM but I don't really mind tbh. Whatever is clearest!

tobyhede: Renamed — `eql_v2_int4_ct` is gone; the plain `eql_v2_int4` is now the storage-only variant.

docs/reference/encrypted-int4-domain.md
tobyhede: Done — `hm` dropped from the ordered variants. They carry `c`, `ob`; equality rides the lossless ORE order term.

docs/reference/encrypted-int4-domain.md
tobyhede: Reworked to the functional-index approach — `eql_v2.ord_term(col)` returns the internal `eql_v2.ore_block_u64_8_256` composite, which carries the btree opclass; no operator class on any domain type. (PR #212 is docs-only — no code overlap.)

docs/reference/encrypted-int4-domain.md
tobyhede: Done — no operator class on any `eql_v2_int4*` domain; the only opclass is on the internal `eql_v2.ore_block_u64_8_256` composite.

docs/reference/encrypted-int4-domain.md
tobyhede: Done — `ORDER BY eql_v2.ord_term(col)`.

docs/reference/encrypted-int4-domain.md
tobyhede: Added as `eql_v2.eq_term(col)`, paired with `eql_v2.ord_term(col)` — reuses the existing `eq_term` extractor name rather than overloading the 2-arg `eql_v2.eq` comparator.

src/encrypted_domain/int4/int4_functions.sql
```diff
@@ -0,0 +1,343 @@
...
+-- =, <> (blockers, 3 shapes each)
+
+--! @brief Blocker for = on eql_v2_int4.
+--! @param a eql_v2_int4
+--! @param b eql_v2_int4
+--! @return boolean (never returns; always raises)
+CREATE FUNCTION eql_v2.eql_v2_int4_eq(a eql_v2_int4, b eql_v2_int4)
+RETURNS boolean IMMUTABLE PARALLEL SAFE
+AS $$ BEGIN RETURN eql_v2.encrypted_domain_unsupported_bool('eql_v2_int4', '='); END; $$
+LANGUAGE plpgsql;
```
coderdan: Is there any benefit in naming the functions for each type instead of using a common naming scheme?

So instead of:
```sql
CREATE FUNCTION eql_v2.eql_v2_int4_eq(a eql_v2_int4, b eql_v2_int4)
```

We create a function called:
```sql
CREATE FUNCTION eql_v2.eq(a eql_v2_int4, b eql_v2_int4)
```

This uses the types as the discriminator and probably means we could implement default versions which just error. Roughly analogous to a trait in Rust.

IIRC that's how postgres does it for all the builtin types and operators.

src/encrypted_domain/int4/int4_functions.sql
```diff
@@ -0,0 +1,343 @@
...
+AS $$ BEGIN RETURN eql_v2.encrypted_domain_unsupported_bool('eql_v2_int4', '>='); END; $$
+LANGUAGE plpgsql;
+
+-- ~~, ~~*, @>, <@ (blockers, 3 shapes each)
+
+--! @brief Blocker for ~~ on eql_v2_int4.
+--! @param a eql_v2_int4
+--! @param b eql_v2_int4
+--! @return boolean (never returns; always raises)
+CREATE FUNCTION eql_v2.eql_v2_int4_like(a eql_v2_int4, b eql_v2_int4)
```
coderdan: I'd prefer to start stepping back from `like`/`ilike` if we making big changes to EQL. Bloom filters don't implement `LIKE` as they are not positionally aware, don't handle wildcards etc. They are more similar to tsearch in Postgres (hence why the underlying index SEM is called `match`). When mapping we can _approximate_ like by mapping to a `match` but I don't think EQL itself should use the term `like`.

`/end-rant` :p

src/encrypted_domain/int4/int4_operators.sql
```diff
@@ -0,0 +1,127 @@
...
+CREATE OPERATOR <> (
+  FUNCTION = eql_v2.eql_v2_int4_neq,
+  LEFTARG = jsonb, RIGHTARG = eql_v2_int4,
+  NEGATOR = =, RESTRICT = neqsel, JOIN = neqjoinsel
+);
+
+CREATE OPERATOR < (
+  FUNCTION = eql_v2.eql_v2_int4_lt,
+  LEFTARG = eql_v2_int4, RIGHTARG = eql_v2_int4,
+  RESTRICT = scalarltsel, JOIN = scalarltjoinsel
```
coderdan, line 50: I think the `RESTRICT` and `JOIN` should be `scalarlesel` and `scalarlejoinsel`.

See the fixes in #217. Might be worth getting Claude to double check all of these because they're easy for our feeble meat brains to miss.

src/encrypted_domain/int4/int4_ord_functions.sql
```diff
@@ -0,0 +1,375 @@
...
+--! @see eql_v2.ore_block_u64_8_256
+--! @example
+--! -- functional index for range + equality
+--! CREATE INDEX t_col_idx ON t USING btree (eql_v2.ord_term(col));
+--! -- ordering
+--! SELECT ... FROM t ORDER BY eql_v2.ord_term(col);
+CREATE FUNCTION eql_v2.ord_term(a eql_v2_int4_ord)
+RETURNS eql_v2.ore_block_u64_8_256
+LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE
+AS $$ SELECT eql_v2.ore_block_u64_8_256(a::jsonb) $$;
```
coderdan, line 60: I'm not sure why we need the indirection. Is there a benefit to doing it this way?

src/encrypted_domain/int4/int4_ord_functions.sql
```diff
@@ -0,0 +1,375 @@
...
+--! @param a eql_v2_int4_ord Ordered encrypted int4 value
+--! @return eql_v2.ore_block_u64_8_256 ORE-block index term
+--! @throws Exception if the `ob` field is missing from the payload
+--! @see eql_v2.ore_block_u64_8_256
+--! @example
+--! -- functional index for range + equality
+--! CREATE INDEX t_col_idx ON t USING btree (eql_v2.ord_term(col));
+--! -- ordering
+--! SELECT ... FROM t ORDER BY eql_v2.ord_term(col);
+CREATE FUNCTION eql_v2.ord_term(a eql_v2_int4_ord)
```
coderdan, line 57: Curious why `ord_term` and not just `ord` - is it a reserved name or something?

src/encrypted_domain/types.sql
```diff
@@ -0,0 +1,81 @@
...
+
+DO $$
+BEGIN
+  --! @brief Storage-only encrypted int4 domain (jsonb-backed). Every
+  --!        operator is a blocker; carries ciphertext (`c`) only.
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_type
+    WHERE typname = 'eql_v2_int4' AND typnamespace = 'public'::regnamespace
+  ) THEN
+    CREATE DOMAIN public.eql_v2_int4 AS jsonb;
```
coderdan: Should we add check constraints on the domain to ensure it has the right innards?

tests/sqlx/tests/encrypted_int4_eq_tests.rs
```diff
@@ -0,0 +1,277 @@
+//! Synthetic test suite for `eql_v2_int4_eq` — HMAC equality only.
```
coderdan, line 1: The tests in this file are great as integration tests.

I reckon we could consider adding unit and property tests, too. For example:

```rust
mod eql_v2 {
    #[derive(Arbitrary)
    struct Int4Eq {
      // -- glossing over details
    }
}

// -- a bit approximate
fn sql_compare(a: &Int4Eq, b: &Int4Eq) -> bool {
    let response = get_conn()
        .query("SELECT eql_v2.eq($1, $2)", &[&a, &b])
        .expect("Failed to execute query");

    let row = response.first().expect("Expected a result");

    row.get(0)
}

quickcheck! {
    fn int4_eq(x: eql_v2::Int4Eq, y: eql_v2::Int4Eq) -> bool {
        sql_compare(&a, &b) == (x.plaintext == y.plaintext)
    }
}
```

coderdan (commented):
I'm really liking the direction here. Its a bit improvement over the single god type.

My sense is that its still perhaps a bit more complicated than it needs to be. So many functions make it harder to follow.

There might be good reasons for all the functions (please let me know if I'm missing something) but then it might make sense to hide them from a public API (e.g. `_foo`) and `@internal` doxygen tags.

src/encrypted_domain/int4/int4_ord_functions.sql
tobyhede: Yeah. This one needs more consideration.
I needed to add an `eq_term` to remove the cast requirement. Then changed this to be aligned.
`eq_term` because `eq` is already loaded for the operators `eql_v2.eq(a eql_v2_int4, b eql_v2_int4)`

tests/sqlx/tests/encrypted_int4_eq_tests.rs
tobyhede: Is a great idea. EQL does not have property tests setup, will mark for follow up.

tobyhede: Created a github issue

docs/reference/encrypted-domain-implementation-spec.md
```diff
@@ -0,0 +1,611 @@
...
+**Decided per type (its own design step):** which variants the family
+has, the payload terms each variant carries, and the index/equality
+scheme. These depend on the type's encryption scheme and are **not**
+mechanical. Examples of decisions the int4 family made that do **not**
+transfer automatically:
+
+- int4's ordered variants carry `c`, `ob` and **drop `hm`** because ORE
+  on a full-domain `int4` is lossless — the order term doubles as an
+  exact equality term. A type whose ORE is **lossy**, or whose domain is
+  not fully covered (candidates: `float`/`double`, `numeric`), must keep
```
coderdan, line 57: Worth some careful analysis but I'm 99% sure that the decimal encoding technique @freshtonic and I worked through recently means ORE representations of `float`, `double` and `numeric` are lossless, too. AFAIK, it's only text/strings that are lossy because UTF-8 canonicalization and compression.

coderdan: In fact, `floats` in ORE might actually compare more reliably than just simply HMAC. Because under the encoding scheme we devised, `-0` and `0` are both represented as `0`. `ORE(0) = ORE(0)` but `HMAC(-0) != HMAC(0)` (floats are stupid).

docs/reference/encrypted-domain-implementation-spec.md
```diff
@@ -0,0 +1,611 @@
...
+  `hm` on its ordered variants and route `=`/`<>` through `eq_term`, not
+  `ord_term`.
+- int4 ships two ordered domains (`_ord` and `_ord_ore`) as mechanical
+  twins. A type with a single ordering scheme needs only `_ord`.
+- `jsonb` does not fit the scalar storage/eq/ord shape; see §11.
+- Native edge semantics are type-specific. The design note must
+  explicitly include or exclude edge values such as float/double `NaN`,
+  infinities and `-0.0`/`0.0`; `numeric` precision/scale boundaries;
+  date min/max and calendar boundaries; timestamp timezone and DST
+  behavior; and bool's two-value domain.
```
coderdan, line 67: `bool` is worth thinking through carefully before we setting on an encryption scheme. ORE and HMAC are both basically worthless for such low-cardinality (2). It's trivial to work out which values are the encryption of `true` and `false`.

There are some randomized approaches we can use (not part of this PR but just flagging for later).

docs/reference/encrypted-domain-implementation-spec.md
```diff
@@ -0,0 +1,611 @@
...
+  file's `--! @file` header (e.g. *"Payload-term assumption: `c`, `hm`."*).
+  If a variant chooses structural `CHECK` validation beyond
+  presence/type-shape, document that choice in the same design note and
+  add matching negative tests (§7, §10).
+
+---
+
+## 4. Reference — Operator classes
+
+**Do not create an operator class on a domain type.** An opclass on a
```
coderdan, line 196: There's one possible problem with this approach that I didn't realise until I hit some problems with the `GROUP BY` benches. `GROUP BY`, `UNION` etc require an operator class on the type, eg:

```sql
CREATE TABLE foo (
  x eql_v2_int4_eq
);

SELECT x,count(*) FROM foo GROUP BY x;
```

The functional form makes the group by work but then you can't get the `c` values:

```sql
CREATE TABLE foo (
  x eql_v2_int4_eq
);

SELECT eql_v2.eq(x),count(*) FROM foo GROUP BY x;
```

It might be possible to create a custom aggregate function to address this but needs verification.

coderdan: Confirmed that we will need to keep operator classes for types that we want to be used in `UNION` and `INTERSECT` queries.

The caveat is that this op-class shouldn't be used for indexing and we should document it clearly.

The custom aggregate approach can be used for `GROUP BY` and would be required for environments that don't support customer op-class.

```sql
CREATE FUNCTION eql_v2.ct_term_sfunc(state jsonb, val eql_v2_int4_eq)
RETURNS jsonb
LANGUAGE sql IMMUTABLE PARALLEL SAFE AS $$
  SELECT COALESCE(state, val -> 'c');
$$;

CREATE FUNCTION eql_v2.ct_term_combine(s1 jsonb, s2 jsonb)
RETURNS jsonb
LANGUAGE sql IMMUTABLE PARALLEL SAFE AS $$
  SELECT COALESCE(s1, s2);
$$;

CREATE AGGREGATE eql_v2.ct_term(eql_v2_int4_eq) (
  sfunc       = eql_v2.ct_term_sfunc,
  stype       = jsonb,
  combinefunc = eql_v2.ct_term_combine,
  parallel    = safe
);

SELECT ct_term(x), count(*)
FROM foo
GROUP BY eq_term(x);
```

`UNION` isn't supported directly on Supabase for now. The workaround would be:

```sql
SELECT DISTINCT ON (eq_term(x)) x, ...
FROM (
  SELECT x, ... FROM a
  UNION ALL
  SELECT x, ... FROM b
) t;
```

coderdan (approved):
I left a lot of comments for follow-ups or future consideration but nothing blocking.

This is excellent work and I'm excited about the approach!

PR #227, opened 2026-05-20
fix: non-raising btree comparator for eql_v2_encrypted (ANALYZE fix)

coderdan:
Fixes #226.

## Problem

`ANALYZE` — and therefore autovacuum — raises `feature_not_supported` on any `eql_v2_encrypted` column carrying an `hm` term but no `ob` term (every equality-configured encrypted column).

`eql_v2_encrypted` is a composite type, and `eql_v2.encrypted_operator_class` is its `DEFAULT` btree opclass. PostgreSQL invokes the opclass's `FUNCTION 1` comparator to gather scalar column statistics during `ANALYZE`. `FUNCTION 1` was `eql_v2.compare`, which #211 made strict — it raises without a Block-ORE `ob` term (U-005). A btree support function must be total and must never raise, so `ANALYZE` failed on every `hm`-only column; autovacuum runs `ANALYZE` routinely, so this broke silently across managed installs.

Removing the opclass is **not** an option: `eql_v2_encrypted` is composite, so without a DEFAULT btree opclass PostgreSQL falls back to built-in `record` comparison (raw ciphertext) — `GROUP BY` / `DISTINCT` then silently fail to deduplicate semantically-equal values. The opclass must stay; its comparator is the bug.

## Fix

Give the btree opclass its own `FUNCTION 1`: **`eql_v2.encrypted_btree_compare`** — a total, non-raising 3-way comparator. Priority: Block ORE when both operands carry `ob`; else a total order on the `hm` term (deterministic, and `= 0` exactly when the hmac terms match — consistent with `=`); else a deterministic payload tie-break so the function stays total.

`eql_v2.compare` is untouched — it stays strict for the `<` / `>` range-operator path.

## Verification

- Scratch-DB repro: `ANALYZE` on an `hm`-only column no longer raises; an opclass btree index builds on `hm`-only data; `GROUP BY` / `DISTINCT` deduplicate correctly.
- Full sqlx suite green. `hash_operator_tests` 17/17 (5 `GROUP BY`/`DISTINCT`/`UNION` tests were failing before — the strict comparator forced the planner onto `record`-comparison sort grouping); `operator_class_tests` 4/4 with **0 ignored** — un-ignored `index_behavior_with_different_data_types` (the strict comparator had broken it) and added `analyze_on_hmac_only_column_does_not_raise`.
- `mise run docs:validate` clean.

## Files

- `src/operators/operator_class.sql` — add `eql_v2.encrypted_btree_compare`, repoint `FUNCTION 1`.
- `tests/sqlx/tests/operator_class_tests.rs` — un-ignore + new regression test.
- `CHANGELOG.md` — `Fixed` entry.


src/operators/operator_class.sql
```diff
@@ -2,27 +2,106 @@
...
+--! @param b eql_v2_encrypted Second value
+--! @return integer -1, 0, or 1
+--!
+--! @internal
+--! @see eql_v2.encrypted_operator_class
+--! @see eql_v2.compare
+CREATE FUNCTION eql_v2.encrypted_btree_compare(a eql_v2_encrypted, b eql_v2_encrypted)
+  RETURNS integer
+  IMMUTABLE STRICT PARALLEL SAFE
+  SET search_path = pg_catalog, extensions, public
```
freshtonic, line 72: @coderdan I assume the `SET search_path = ...` is OK here.

freshtonic (approved):
Approved, but see my comment.

src/operators/operator_class.sql
coderdan: Actually, I'm not 100% sure. Let me double check.

coderdan: Oh, yes its fine. This function is `pl/pgsql` so its not inlineable anyway!

PR #233, opened 2026-05-20
fix(constraints): accept SteVec payloads in _encrypted_check_c

calvinbrewer:
## Summary

- Fixes [#232](https://github.com/cipherstash/encrypt-query-language/issues/232) — `eql_v2._encrypted_check_c` now accepts the v2.3 SteVec payload shape (root `sv` instead of root `c`), so the autogenerated `eql_v2_encrypted_constraint_<table>_<column>` CHECK constraint no longer rejects JSON-encrypted inserts emitted by `@cipherstash/protect-ffi >= 0.22.0` under `mode: standard` (the new default).
- Per the v2.3 schema (`docs/reference/schema/eql-payload-v2.3.schema.json`), an `eql_v2_encrypted` root admits two mutually exclusive shapes: `EncryptedPayload` (carries `c` at the root) and `SteVecPayload` (carries `sv` at the root). The check now passes when either is present; the existing per-element ciphertext invariant on `sv` entries is unchanged (still enforced by the `eql_v2.ste_vec_entry` DOMAIN).
- Added a regression test in `tests/sqlx/tests/constraint_tests.rs` (`check_encrypted_accepts_stevec_payload`) that mirrors the issue's repro payload, verifies both `eql_v2.check_encrypted` and the `add_encrypted_constraint`-installed CHECK, and confirms the negative path (neither `c` nor `sv` at root) still raises.
- `CHANGELOG.md` `[Unreleased] → Fixed` entry added.

## Test plan

- [x] `mise run build` succeeds, both Main and Supabase variants
- [x] `cargo test --test constraint_tests` passes (8/8, including new test)
- [x] Full SQLx suite still passes locally (Postgres 17)
- [ ] CI green on PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Encrypted payload validation now accepts the v2.3 encrypted payload shape where the root ciphertext may be under the top-level sv field, improving insert compatibility.

* **Tests**
  * Added regression tests verifying acceptance of the v2.3 payload shape and rejection of malformed payloads.

* **Chores**
  * Test fixtures and migrations updated to apply production-style encrypted-payload CHECK constraints as regression guards.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/encrypt-query-language/pull/233?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/encrypted/constraints.sql
```diff
@@ -100,10 +109,10 @@ CREATE FUNCTION eql_v2._encrypted_check_c(val jsonb)
   SET search_path = pg_catalog, extensions, public
 AS $$
 	BEGIN
-    IF (val ? 'c') THEN
+    IF (val ? 'c') OR (val ? 'sv') THEN
```
coderdan, line 112: Note that the test fixtures need to be updated now as well or this check will fail the build.

If it doesn't, we may have bigger problems!

coderdan (changes requested):
Fixtures need updating, too.

tests/sqlx/migrations/003_install_ste_vec_data.sql
```diff
@@ -6,6 +6,17 @@ CREATE TABLE ste_vec
...
+-- Apply the production `add_encrypted_constraint` CHECK to the fixture
+-- table. Every row in this file is a real v2.3 SteVec payload (root `sv`,
+-- no root `c`), so the constraint also acts as a regression guard against
+-- #232: under the pre-fix `_encrypted_check_c` (which required `val ? 'c'`
+-- unconditionally) every INSERT below would fail. If this fixture loads
+-- cleanly with the constraint applied, the check function correctly
+-- admits both `EncryptedPayload` and `SteVecPayload` shapes; if it loads
+-- cleanly with `add_encrypted_constraint` somehow not wiring the CHECK,
+-- something more fundamental is broken (h/t @coderdan).
+SELECT eql_v2.add_encrypted_constraint('ste_vec', 'e');
```
coderdan, line 18: I recommend adding the check for all fixture tables to prevent any regressions.

calvinbrewer: Added the constraint to the two remaining fixture tables that hold real-shape EQL payloads in `829267d`:

- `bench` (all three encrypted columns) in `007_install_bench_data.sql` — verified by loading the 10K-row seed against the constraint locally.
- `encrypted` in `like_data.sql`.

The other fixture tables intentionally use minimal synthetic payloads that can't carry the production CHECK without rewriting fixture data:

- `ore` / `ore_text` — `{"ob": [...]}` only, exercising ORE block operator logic in isolation (no `v`/`i`/`c`).
- `agg_test` — v=1 payloads carried over from the legacy aggregate-tests SQL file.
- `encrypted` in `encrypted_json.sql` / `array_data.sql` / `order_by_null_data.sql` — used by tests that add/remove the constraint dynamically or that copy minimal-shape rows from `ore`.
- `constrained` (`constraint_tables.sql`) — exists specifically to exercise Postgres-side UNIQUE / NOT NULL / CHECK behaviour, including NULL inserts that would also collide with a production EQL CHECK.

Happy to rewrite any of those fixture seeds to use real-shape payloads in a follow-up if you'd rather have full coverage — it's a chunkier change and didn't feel in scope for this fix, but I don't want to leave the gap silently.

coderdan (approved):

PR #238, opened 2026-05-24
feat(fixtures): plug-in framework + eql_v2_int4 reference fixture

tobyhede:
## Example

Canonical example from `eql_v2_int4.rs`:

```rust
use crate::fixtures::FixtureSpec;

const VALUES: &[i32] = &[
    -100, -1, 1, 2, 5, 10, 17, 25, 42, 50, 100, 250, 1000, 9999,
];

pub fn spec() -> FixtureSpec<'static, i32> {
    FixtureSpec::new("eql_v2_int4")     // fixture name → fixtures.eql_v2_int4, _fixture_eql_v2_int4, eql_v2_int4.sql
        .index("unique")                 // HMAC equality index
        .index("ore")                    // ORE block ordering index
        .column_type("jsonb")            // committed payload column type
        .values(VALUES)                  // plaintext data — T inferred as i32 from the slice
}
```

## Notes
The core benchmark/proxy integration is unchanged from #224.

## Test plan

- `mise run test`
- `cd tests/sqlx && cargo test --lib fixtures`
- `cd tests/sqlx && cargo test --test eql_v2_int4_fixture_tests`
- `mise run fixture:generate eql_v2_int4` (manual; requires Proxy)
- CI: PG 14–17

## Summary by CodeRabbit

* **New Features**
  * Added an automated, type-checked SQLx fixture generator that can produce encrypted JSONB fixtures and a runnable generator command.

* **Documentation**
  * Documented generator usage, credential requirements, fixture schema, and regeneration workflow; updated fixture headers.

* **Tests**
  * Added unit and integration tests validating fixture specs, generation, encryption output shape, and replayability.

* **Chores**
  * Introduced opt-in feature flag and test-time dependency for generation, CI secrets for generator credentials, and extended .gitignore for generated/local artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/encrypt-query-language/pull/238?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tasks/fixtures.toml
```diff
@@ -0,0 +1,54 @@
...
+# Prerequisites:
+#   - mise run postgres:up   (Postgres with EQL installed)
+#   - mise run proxy:up      (Proxy on localhost:6432)
+#
+# Usage: mise run fixture:generate eql_v2_int4
+dir = "{{config_root}}/tests/sqlx"
+run = """
+cargo test --features fixture-gen --lib \
+  fixtures::{{arg(name="fixture")}}::generate \
+  -- --ignored --exact --nocapture
```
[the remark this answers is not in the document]
tobyhede: Addressed in 6832927: arg captured into a shell variable and validated against [a-z0-9_]+ before being interpolated into the cargo target. Invalid names exit 1 with a clear message.

tests/sqlx/tests/eql_v2_int4_fixture_tests.rs
```diff
@@ -0,0 +1,128 @@
...
+
+#[sqlx::test(fixtures(path = "../fixtures", scripts("eql_v2_int4")))]
+async fn plaintext_oracle_supports_value_filtering(pool: PgPool) -> Result<()> {
+    // The in-table `plaintext` oracle: a consuming test can filter on it
+    // directly. Exactly one row has plaintext = 42.
+    let id: i64 = sqlx::query_scalar("SELECT id FROM fixtures.eql_v2_int4 WHERE plaintext = 42")
+        .fetch_one(&pool)
+        .await?;
+    assert_eq!(id, 9, "plaintext 42 is the 9th generated value");
+    Ok(())
```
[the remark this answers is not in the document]
tobyhede: Addressed in 6832927: switched to fetch_all with assert_eq!(ids, vec![9]) so duplicate rows would now fail the test.

tests/sqlx/src/fixtures/validation.rs
```diff
@@ -0,0 +1,129 @@
...
+
+fn is_valid_column_type(s: &str) -> bool {
+    ALLOWED_COLUMN_TYPES.contains(&s)
+}
+
+/// A validated SQL identifier. Construction proves the string matches
+/// `^[a-z][a-z0-9_]*$`. Renderers interpolate via `Display`, so the bare
+/// `&str` cannot reach generated SQL once it has been validated into this type.
+#[derive(Debug, Clone)]
+pub struct Identifier(String);
```
coderdan: I got confused for a second because I thought this code was about the EQL Identifier concept (i.e. column/table).

Maybe we could rename this type to avoid confusion?

tests/sqlx/fixtures/eql_v2_int4.sql
```diff
@@ -0,0 +1,29 @@
+-- AUTO-GENERATED by `mise run fixture:generate eql_v2_int4`.
```
coderdan: If autogenerated, should it be committed? Possibly is ok just checking.

tests/sqlx/fixtures/eql_v2_int4.sql
```diff
@@ -0,0 +1,29 @@
...
+-- DO NOT EDIT BY HAND. Re-run the generator to refresh.
+--
+-- Encrypted via CipherStash Proxy (HMAC + ORE block terms).
```
coderdan: Using proxy instead of the client directly for this is probably fine but has a couple of gotchas:
1. Proxy must be using the same version of cipherstash-client as the Stack so that the EQL stuff works for both
2. You need a proxy image as well as the DB

Not deal-breakers but I think would be preferred to use cipherstash-client directly.

tests/sqlx/fixtures/eql_v2_int4.sql
coderdan: Actually, I can think of a reason _not_ to commit - stale fixtures. This bit me last week when tests passed but the fixture data was actually wrong and hid an underlying issue.

tests/sqlx/fixtures/FIXTURE_SCHEMA.md
```diff
@@ -7,13 +7,21 @@ This document defines the structure and dependencies of test fixtures used in th
 ```
 EQL Extension (via migrations)
   ├── encrypted_json.sql
-  ├── array_data.sql
+  │   └── array_data.sql (extends `encrypted` table from encrypted_json)
+  ├── like_data.sql
```
coderdan: ~~like~~ -> match

coderdan (approved):
Approved so as not to block but a couple of things worth considering as we build this out.

tests/sqlx/fixtures/eql_v2_int4.sql
tobyhede: @coderdan As noted, The core setup is unchanged from https://github.com/cipherstash/encrypt-query-language/pull/224.

Will pull client in

mise.toml
```diff
@@ -45,7 +50,16 @@ echo "Running SQLx migrations..."
...
 sqlx migrate run
 
+# Regenerate fixtures every run — they are not committed (see .gitignore).
+# Generator encrypts via cipherstash-client directly; CS_* credentials must
+# be present in the shell environment (CS_CLIENT_ACCESS_KEY +
+# CS_WORKSPACE_CRN, or the legacy CS_CLIENT_ID/CS_CLIENT_KEY pair).
+echo "Regenerating SQLx fixtures..."
+cd "{{config_root}}"
+mise run fixture:generate eql_v2_int4
+
```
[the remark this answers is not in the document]
tobyhede: Test coverage requires the fixtures. This is a hard requirement.

PR #239, opened 2026-05-24
feat: eql_v2_int4 variant family (v3)

tobyhede:
## Stack

`eql_v3` ← `fixtures-and-bench` ([#238](https://github.com/cipherstash/encrypt-query-language/pull/238)) ← **this PR** (base: `fixtures-and-bench`).

Consumes the `fixtures.eql_v2_int4` SQLx fixture from #238. Supersedes [#225](https://github.com/cipherstash/encrypt-query-language/pull/225).

## Summary

`eql_v2_int4` variant family — four jsonb-backed PostgreSQL domains for encrypted `int4`, plus a reusable scalar-domain SQL materializer that generates them from a minimal TOML manifest.

| Domain | Manifest terms | Operator surface |
| --- | --- | --- |
| `eql_v2_int4`         | `[]`      | storage only — every operator blocked |
| `eql_v2_int4_eq`      | `hm`      | `=`, `<>` |
| `eql_v2_int4_ord_ore` | `ore`     | `=`, `<>`, `<`, `<=`, `>`, `>=` |
| `eql_v2_int4_ord`     | `ore`     | same surface — recommended name |

Ciphertext envelope (`c`) is always present. Uniform extractors `eql_v2.eq_term(col)` / `eql_v2.ord_term(col)`; functional indexes engage without `::jsonb` casts, and ordered columns share one btree across equality + range.

## What's here

- **Codegen materializer** (`tasks/codegen/`, Python): renders `*_types/_functions/_operators/_aggregates.sql` from `tasks/codegen/types/int4.toml`. The whole manifest:

  ```toml
  # tasks/codegen/types/int4.toml — filename supplies the type token.
  [domain]
  int4         = []        # storage only
  int4_eq      = ["hm"]    # equality
  int4_ord_ore = ["ore"]   # equality + ordering
  int4_ord     = ["ore"]   # recommended name
  ```

  Term capabilities are fixed in `terms.py` (`hm`→equality, `ore`→equality+ordering). Output is byte-identical and headed `AUTO-GENERATED — DO NOT EDIT`. `mise run build` regenerates all domains every build; `mise run codegen:domain int4` refreshes one.
- **Per-domain `MIN`/`MAX` aggregates** on ord-capable variants (`eql_v2.min/max(eql_v2_int4_ord)`). Composite-type `eql_v2.min/max(eql_v2_encrypted)` are **retained** and coexist.
- **Test matrix** — the entire SQLx suite for a scalar is one `impl ScalarType` + one macro invocation. int4 in full:

  ```rust
  ordered_numeric_matrix! {
      suite = int4,
      scalar = i32,
      eql_type = "eql_v2_int4",
  }
  ```

- **Domain lint rules** (`src/lint/lints.sql`): `blocker_language`, `blocker_strict`, `domain_over_domain`, `domain_opclass`. `pin_search_path` + splinter allowlist recognise the converged extractor/wrapper names.

- **Committed reference baseline** under `tests/codegen/reference/int4/` (`-- REFERENCE:` marker) guards byte parity.
- **Docs**: `docs/reference/encrypted-domain-implementation-spec.md`, `encrypted-domain-generator.md`; `eql-functions.md` / `sql-support.md` document both aggregate paths; CHANGELOG `Added` entry.
- **CI**: new `codegen` job gating the PG matrix; `permissions: contents: read` + `persist-credentials: false` on checkouts; `CS_CLIENT_ID`/`CS_CLIENT_KEY` env added.

## Test plan

- [ ] `mise exec python -- python -m pytest tasks/codegen -q`
- [ ] `mise run codegen:domain int4 && git diff -- src/encrypted_domain/int4` — empty (byte parity)
- [ ] `mise run clean && mise run build`
- [ ] `mise run test --postgres 17`
- [ ] `mise run --output prefix test:splinter --postgres 17` — `unmatched=0`
- [ ] `cargo test encrypted_int4* encrypted_domain_inlinability_tests`

## Summary by CodeRabbit

* **New Features**
  * Added encrypted-domain scalar types with automatic code generation from type manifests.
  * Introduced per-domain `MIN`/`MAX` aggregates for ordered encrypted-domain types, routing comparisons through variant operators without decryption.

* **Documentation**
  * Added encrypted-domain implementation specification and generator reference.
  * Updated SQL support matrix and function documentation for new scalar-domain aggregates and extractors.

* **Tests**
  * Enhanced encrypted-domain test coverage for correctness and invariant validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cipherstash/encrypt-query-language/pull/239?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tasks/codegen/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""Encrypted-domain SQL code generator for EQL scalar domain families."""
```
tobyhede, line 1: **Full disclosure:** 
I have not looked at the Python. 

**Assumptions:** 
"Working" is sufficient verification.

**Supported By:**
- Code is basic 
- Problem is basic
- Many and extensive agent reviews 
- Test confirms that the generated code == reference source fixture.

tests/sqlx/src/matrix.rs
```diff
@@ -0,0 +1,2318 @@
+//! Type-generic test matrix for encrypted scalar domains.
```
tobyhede, line 1: And also ... not sure how I feel about the macro tests.
I am adding a test for them because ... jfc.
And I would like to loop back and do a pass to make them more human readable.

tests/codegen/reference/int4/int4_eq_functions.sql
```diff
@@ -0,0 +1,406 @@
...
+AS $$ SELECT eql_v2.eq_term(a::eql_v2_int4_eq) <> eql_v2.eq_term(b) $$;
+
+--! @brief Blocker for < on eql_v2_int4_eq.
+--! @param a eql_v2_int4_eq
+--! @param b eql_v2_int4_eq
+--! @return boolean (never returns; always raises)
+CREATE FUNCTION eql_v2.lt(a eql_v2_int4_eq, b eql_v2_int4_eq)
+RETURNS boolean IMMUTABLE PARALLEL SAFE
+AS $$ BEGIN RETURN eql_v2.encrypted_domain_unsupported_bool('eql_v2_int4_eq', '<'); END; $$
+LANGUAGE plpgsql;
```
coderdan, line 73: Nice. Polymorphism FTW

tests/codegen/reference/int4/int4_eq_operators.sql
```diff
@@ -0,0 +1,233 @@
...
+CREATE OPERATOR <> (
+  FUNCTION = eql_v2.neq,
+  LEFTARG = jsonb, RIGHTARG = eql_v2_int4_eq,
+  COMMUTATOR = <>, NEGATOR = =, RESTRICT = neqsel, JOIN = neqjoinsel
+);
+
+CREATE OPERATOR < (
+  FUNCTION = eql_v2.lt,
+  LEFTARG = eql_v2_int4_eq, RIGHTARG = eql_v2_int4_eq
+);
```
coderdan, line 49: Should this (and those below) have `COMMUTATOR`, `RESTRICT`, `JOIN` and `NEGATOR`? I suspect yes or else the planner won't be able to reason about index usage at query time.

coderdan: Oh wait. Its because this type only supports `equality` these operators are just placeholders that throw errors.

Maybe just a comment explaining that in case future reviewers wonder the same.

tests/codegen/reference/int4/int4_ord_aggregates.sql
```diff
@@ -0,0 +1,68 @@
...
+LANGUAGE plpgsql IMMUTABLE STRICT
+SET search_path = pg_catalog, extensions, public
+AS $$
+BEGIN
+  IF value < state THEN
+    RETURN value;
+  END IF;
+  RETURN state;
+END;
+$$;
```
coderdan, line 31: Does this need to be plpgsql? Could possibly use a `CASE` in `SQL` though I'm not sure if the performance impact here is significant.

tests/codegen/reference/int4/int4_ord_functions.sql
```diff
@@ -0,0 +1,395 @@
...
+-- REQUIRE: src/ore_block_u64_8_256/functions.sql
+-- REQUIRE: src/ore_block_u64_8_256/operators.sql
+
+--! @file encrypted_domain/int4/int4_ord_functions.sql
+--! @brief Ordered domain of the int4 encrypted-domain family — comparison/path functions.
+
+--! @brief Index extractor for the eql_v2_int4_ord variant.
+--! @param a eql_v2_int4_ord
+--! @return eql_v2.ore_block_u64_8_256
+CREATE FUNCTION eql_v2.ord_term(a eql_v2_int4_ord)
```
coderdan, line 14: While I think of it, everything is still in the `eql_v2` schema, not `eql_v3` - what's the plan there? Keep as is or move everything over?

tests/codegen/reference/int4/int4_types.sql
```diff
@@ -0,0 +1,68 @@
...
+
+  --! @brief Equality-only encrypted int4 domain.
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_type
+    WHERE typname = 'eql_v2_int4_eq' AND typnamespace = 'public'::regnamespace
+  ) THEN
+    CREATE DOMAIN public.eql_v2_int4_eq AS jsonb
+      CHECK (
+        jsonb_typeof(VALUE) = 'object'
+        AND VALUE ? 'v'
```
coderdan, line 32: Should we check that `v == 2` (or 3)?

tests/codegen/reference/int4/int4_ord_functions.sql
tobyhede: Was attempting to defer the decision.
But I think it would actually be clearer to change to `eql_V3`
Will do that.

tests/sqlx/src/matrix.rs
```diff
@@ -0,0 +1,2403 @@
...
+                sqlx::query(&format!(
+                    "INSERT INTO {table}(plaintext, value) \
+                     SELECT plaintext, payload::{d} FROM {fixture}", d = &spec.sql_domain, fixture = fixture_table,
+                )).execute(&mut *tx).await?;
+                sqlx::query(&format!(
+                    "CREATE INDEX {index} ON {table} USING {using} ({extractor}(value))", using = $using, extractor = $extractor,
+                )).execute(&mut *tx).await?;
+                sqlx::query(&format!("ANALYZE {table}"))
+                    .execute(&mut *tx).await?;
+                sqlx::query("SET LOCAL enable_seqscan = off").execute(&mut *tx).await?;
```
coderdan, line 1671: **Index engagement here proves the index is *usable*, not *preferred*.** With `enable_seqscan = off` on a ~17-row table the planner will pick almost any index to avoid the forced-expensive seqscan, so an index it would never choose under realistic costs still passes — and `plan_text.contains(index)` is a substring match that doesn't distinguish an `Index Scan` from a bitmap recheck that still reads every row. The real cost-preference proof (`__scalar_matrix_scale_case`) is behind `#[cfg(feature = "scale")]`, which is off in PR CI, so the headline "operators engage the functional indexes" claim is only validity-checked by default.

Suggest one of: run a single scale-preference case in the standard PG matrix, assert on the EXPLAIN *node type* (`Index Scan`/`Index Only Scan`/`Bitmap Index Scan`) rather than a name substring, or at minimum document loudly that default CI verifies index validity only.

tests/sqlx/src/matrix.rs
```diff
@@ -0,0 +1,2403 @@
...
+
+                let pivot_payload: String = sqlx::query_scalar(&format!(
+                    "SELECT (payload - 'hm')::text FROM {fixture} WHERE plaintext = {lit}",
+                    fixture = fixture_table, lit = pivot_lit,
+                )).fetch_one(&mut *tx).await?;
+
+                let eq_count: i64 = sqlx::query_scalar(&format!(
+                    "SELECT count(*) FROM {table} WHERE value = $1::jsonb::{d}",
+                )).bind(&pivot_payload).fetch_one(&mut *tx).await?;
+                anyhow::ensure!(eq_count >= 1,
```
coderdan: `eq_count >= 1` is weak, and it isn't independent of the `<>` check below: `expected_neq` is computed as `len - eq_count`, so an `=` that over-matches inflates `eq_count` and deflates `expected_neq` in lockstep — both assertions still pass. The fixture plaintexts are distinct, so pin `eq_count == 1` to make this (and the derived `<>` count) actually load-bearing.

tasks/codegen/templates.py
```diff
@@ -0,0 +1,368 @@
...
+        return ""
+    return f" ({_arg_label(dom, arg_a)}, {_arg_label(dom, arg_b)})"
+
+
+def render_domain_block(domain: DomainSpec, token: str) -> str:
+    """One idempotent IF NOT EXISTS CREATE DOMAIN block, prefixed by a
+    per-domain --! @brief derived from role + token."""
+    dom = domain_name(domain.name)
+    keys = ENVELOPE_KEYS + [CIPHERTEXT_KEY] + term_json_keys(domain.terms)
+    checks = "\n        AND ".join(f"VALUE ? '{key}'" for key in keys)
```
coderdan: This interpolates `key` straight into SQL text (`VALUE ? '{key}'`), and the same pattern is used for `op`/`dom`/`backing` throughout the `render_*` helpers — none of them escape single quotes. It's safe *today* only because every string in `TERM_CATALOG`/`OPERATORS`/`SCALAR_KINDS` happens to be quote-free, but nothing enforces that invariant. The first catalog entry containing a `'` produces broken DDL (or worse). Cheap hardening: a `_sql_str()` quote-doubler at the interpolation boundary, or an `assert "'" not in key` guard.

tasks/codegen/spec.py
```diff
@@ -0,0 +1,118 @@
...
+    if not isinstance(values, list):
+        raise SpecError("[fixture] values: must be a list of value tokens")
+    if not values:
+        raise SpecError("[fixture] values: must not be empty")
+    if any(not isinstance(v, str) for v in values):
+        raise SpecError("[fixture] values: must be strings")
+
+    try:
+        kind = require_scalar(token)
+        numbers = {kind.numeric_value(v) for v in values}
```
coderdan: Fixture values flow into a `set` for range-checking, so duplicate values pass validation silently and emit a duplicated Rust literal. The matrix oracle documents a "distinct plaintext values" contract — a copy-paste dup in a manifest would violate it with no guard. Worth a `len(set(values)) == len(values)` check (or dedupe-with-error) alongside the MIN/MAX/zero check below.

tasks/codegen/operator_surface.py
```diff
@@ -0,0 +1,56 @@
...
+    backing: str  # eql_v2 backing function name (bare or quoted)
+    kind: Literal["symmetric", "path", "blocker_only"]
+    restrict: str | None  # selectivity estimator, symmetric ops only
+    join: str | None  # join selectivity estimator, symmetric ops only
+    commutator: str | None
+    negator: str | None
+
+
+SYMMETRIC_OPERATORS = ["=", "<>", "<", "<=", ">", ">=", "@>", "<@"]
+PATH_OPERATORS = ["->", "->>"]
```
coderdan, line 26: These three lists are the basis of the storage-only "every operator blocked" guarantee, and together they are exactly the full jsonb operator surface for PG 14–17 (verified) — so no hole today. But it's an enumerated list over a `jsonb` base type, not a structural guarantee: any future-PG jsonb operator not added here silently falls through to native plaintext-jsonb semantics on an encrypted column. Cheap insurance: a test that queries `pg_operator`/`pg_type` for jsonb operators absent from this set and fails when one appears.

tests/codegen/reference/int4/int4_ord_aggregates.sql
```diff
@@ -0,0 +1,68 @@
...
+
+--! @brief Find the minimum encrypted value in a group of eql_v2_int4_ord values.
+--!
+--! Comparison routes through the domain's `<` operator, which uses the ORE block term — no decryption.
+--!
+--! @param input eql_v2_int4_ord encrypted values to aggregate
+--! @return eql_v2_int4_ord minimum of the group, or NULL if all inputs are NULL
+CREATE AGGREGATE eql_v2.min(eql_v2_int4_ord) (
```
coderdan, line 42: Decision point, not a defect: this aggregate declares no `COMBINEFUNC` and no `PARALLEL = SAFE` (and the composite `min/max(eql_v2_encrypted)` family is the same), so PostgreSQL will never use partial/parallel aggregation. That's the safe default, but it forecloses parallel aggregation on exactly the large `GROUP BY` workloads these ORE aggregates exist to serve. Worth a deliberate "not now, because…" rather than a silent omission — the sfunc is `IMMUTABLE STRICT`, so `PARALLEL SAFE` + a combinefunc would be straightforward if wanted.

tests/codegen/reference/int4/int4_types.sql
```diff
@@ -0,0 +1,68 @@
...
+        AND VALUE ? 'i'
+        AND VALUE ? 'c'
+        AND VALUE ? 'ob'
+      );
+  END IF;
+
+  --! @brief Ordered encrypted int4 domain.
```
coderdan: Nit: `eql_v2_int4_ord` and `eql_v2_int4_ord_ore` (L38) share an identical `@brief` ("Ordered encrypted int4 domain.") even though they're distinct domains. Since `int4_ord` is the recommended converged name and `int4_ord_ore` is the scheme-explicit twin, the briefs should say so.

coderdan (approved):
Epic. I don't mind the macro approach. Its a bit hairy but its all autogen so no big deal.

PR #255, opened 2026-06-03
Self-contained eql_v3 schema + standalone v3 installer

tobyhede:
## Summary

Severs `eql_v3`'s runtime dependency on `eql_v2`: gives it its own SEM index-term types, relocates the whole v3 surface under `src/v3/`, and ships a CI-gated v3-only installer. Implements `docs/plans/eql-v3-self-contained-plan.md` (D1–D13). `eql_v2` is unchanged.

- Hand-written SEM types `eql_v3.hmac_256` and `eql_v3.ore_block_u64_8_256` (jsonb-only ports). No `eql_v2.<symbol>` appears anywhere in v3 SQL.
- Codegen `SCHEMA` → `eql_v3`; scalar output paths + REQUIRE edges repointed to `src/v3/scalars/`; golden regenerated.
- Standalone `release/cipherstash-encrypt-v3.sql`, globbed from `src/v3` alone, installable with no `eql_v2` present.
- Self-containment gate `mise run test:self_contained_v3` (symbol + dependency-closure + artifact greps), wired into CI.
- Combined-build inlining preserved: new SEM functions added to `pin_search_path.sql`'s inline-critical allowlist + splinter rows + regression guard.

## Directory layout (`src/v3/`)

```
src/v3/
├── schema.sql              # CREATE SCHEMA eql_v3
├── crypto.sql              # forked from eql_v2 (D7)
├── common.sql              # forked from eql_v2 (D8)
├── sem/                    # hand-written, committed SEM index-term types
│   ├── hmac_256/           #   types + functions
│   └── ore_block_u64_8_256/#   types, functions, operators, operator_class
└── scalars/
    ├── functions.sql       # shared jsonb-operator blocker (committed)
    └── <T>/                # generated per scalar (gitignored): int4, int2, …
                            #   <T>_types/functions/operators + _eq/_ord/_ord_ore variants
```

## Test Plan

- [x] `mise run clean && mise run build` → all four variants (Main, Supabase, Protect, v3)
- [x] `cargo test -p eql-scalars -p eql-codegen` + `mise run codegen:parity` (byte-for-byte golden)
- [x] `mise run test:self_contained_v3` → no eql_v2 symbol/file leak
- [x] `mise run test:splinter` → 184/184 allowlisted, 0 unmatched
- [x] `mise run test:clean_install_v3` → installs with no eql_v2; ORE opclass functional index works (D4); blocker raises (D11)
- [x] `mise run test:matrix:inventory` → reconciled (int4, int2)
- [x] `mise run test` (full SQLx suite, PG17) → ALL TESTS PASSED

## Summary by CodeRabbit

* **New Features**
  * Added standalone v3 schema installer artifact with self-contained encrypted-domain support
  * Introduced searchable encryption index-term functions for HMAC-256 and Order-Revealing Encryption primitives
  * Implemented v3 schema independence from v2 infrastructure

* **Documentation**
  * Updated reference guides for encrypted-domain type definitions and schema architecture
  * Added guidance for v3-only configuration and setup

* **Tests**
  * Added v3 self-containment validation tests
  * Implemented clean installation smoke tests for v3 schema
  * Added searchable encryption semantic function behavioral tests

* **Chores**
  * Enhanced CI pipeline with v3 build verification
  * Updated build system for v3 artifact generation and cleanup


src/v3/schema.sql
```diff
@@ -0,0 +1,23 @@
...
+--! @warning DROP SCHEMA CASCADE will remove all objects in the schema
+--! @note eql_v3 is a new, additional schema for the encrypted-domain families.
+
+--! @brief Drop existing EQL v3 schema
+--! @warning CASCADE will drop all dependent objects
+DROP SCHEMA IF EXISTS eql_v3 CASCADE;
+
+--! @brief Create EQL v3 schema
+--! @note Houses the encrypted-domain type families
+CREATE SCHEMA eql_v3;
```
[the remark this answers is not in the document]
tobyhede: Intentional — `src/v3/schema.sql` is a faithful mirror of the shipped `eql_v2` `src/schema.sql`, which uses the identical `DROP SCHEMA IF EXISTS eql_v2 CASCADE; CREATE SCHEMA eql_v2;` with the same `@warning`. Drop-and-recreate for a clean, idempotent reinstall is the established EQL install model for both schemas, and a dedicated destructive teardown already ships separately as `tasks/uninstall-v3.sql` (→ `release/cipherstash-encrypt-v3-uninstall.sql`). Removing `CASCADE` from v3 only would make it inconsistent with the rest of the product; revisiting the install-time CASCADE policy would be an eql_v2-wide design decision, out of scope for this self-contained-v3 PR (which deliberately leaves eql_v2 unchanged). Resolving as intentional.

src/v3/sem/hmac_256/types.sql
```diff
@@ -0,0 +1,12 @@
...
+--! @file v3/sem/hmac_256/types.sql
+--! @brief HMAC-SHA256 index term type (eql_v3 SEM)
+--!
+--! Domain type representing HMAC-SHA256 hash values. Used for exact-match
+--! encrypted searches. The hash is stored in the 'hm' field of encrypted data
+--! payloads. Self-contained eql_v3 copy (design D1/D3); the eql_v2 original is
+--! unchanged.
+--!
+--! @note Transient type used only during query execution.
+CREATE DOMAIN eql_v3.hmac_256 AS text;
```
coderdan, line 12: Minor thing but HMACs are always a fixed length so I wonder if a varchar is actually preferred here? I suspect `text` is perfectly fine but thought I'd check.

src/v3/sem/ore_block_u64_8_256/functions.sql
```diff
@@ -0,0 +1,206 @@
...
+    RETURN NULL;
+  END IF;
+
+  SELECT array_agg(ROW(b)::eql_v3.ore_block_u64_8_256_term)
+  INTO terms
+  FROM unnest(eql_v3.jsonb_array_to_bytea_array(val)) AS b;
+
+  RETURN ROW(terms)::eql_v3.ore_block_u64_8_256;
+END;
+$$ LANGUAGE plpgsql;
```
coderdan: This should be possible with an `SQL` function. I realise that it might just be copy-paste from the old functions but might be worth updating while we are at it.

src/v3/common.sql
```diff
@@ -0,0 +1,38 @@
...
+--!
+--! Converts a JSONB array of hex-encoded strings into a PostgreSQL bytea array.
+--! Used for deserializing binary data (like ORE terms) from JSONB storage.
+--!
+--! @param val jsonb JSONB array of hex-encoded strings
+--! @return bytea[] Array of decoded binary values
+--!
+--! @note Returns NULL if input is JSON null
+--! @note Each array element is hex-decoded to bytea
+CREATE FUNCTION eql_v3.jsonb_array_to_bytea_array(val jsonb)
```
coderdan, line 30: I think this can be `SQL` now, too.

coderdan (approved):

src/v3/sem/hmac_256/types.sql
tobyhede: Pretty sure is the same underlying implementation in Postgres so the benefit is semantic clarity - no performance implication. Unless you really care, I think this is ok?

src/v3/sem/ore_block_u64_8_256/functions.sql
tobyhede: Done in `904661c` — now inlinable `LANGUAGE sql IMMUTABLE`. Used a `CASE` over a scalar subquery (a `FROM`-clause `jsonb_array_elements_text` + `WHERE` would raise on JSON `null` instead of returning `NULL`), and added the `eql-inline-critical` COMMENT marker so it actually inlines (bare-`jsonb` arg dodges the structural skip). v2 left as plpgsql.

src/v3/common.sql
tobyhede: Done in `904661c`, same treatment as `jsonb_array_to_ore_block_u64_8_256` above. v2 left as plpgsql.

PR #256, opened 2026-06-03
feat(scalars): add eql_v3.date encrypted-domain type + temporal wiring

tobyhede:
## Summary

**PR 2 of 2 (stacked on #258).** Adds the `eql_v3.date` encrypted-domain scalar — the first **non-integer ordered** type. The integer-agnostic harness refactor that made this possible now lives in its parent, **#258** (`refactor(tests): generalise scalar matrix harness off integer-inherent consts`); this PR is the catalog row plus the temporal wiring that refactor enabled.

> Review/merge **#258 first.** This branch's diff is only the date-addition delta on top of it.

## What changed

**Catalog (`eql-scalars`)** — `ScalarKind::Date` (`chrono::NaiveDate`), `Fixture::Date(&str)` (zero-dep ISO strings), `DATE_FIXTURES` (16 dates incl. the three matrix pivots), `pub const DATE` appended to `CATALOG`, plus mirrored panic / pivot-presence / token-order tests. The catalog stays zero-dependency: it stores ISO strings; chrono parsing lives in the harness.

**Harness** — an explicit `[temporal]` marker in the `scalar_types!` dispatch list (`date => chrono::NaiveDate [temporal]`) drives the two divergences from the integer path:
1. the `impl ScalarType` for a temporal scalar is **hand-written** in `scalar_domains.rs` (chrono values can't be a `const` slice; pivots are explicit sentinels), so the macro skips it;
2. `scalar_fixture!` stamps a **pivot-presence** assert instead of the integer signed-extreme asserts.

Adds `impl ScalarType for NaiveDate` (LazyLock-parsed values, `1900-01-01`/`2099-12-31` pivot sentinels, quoted `to_sql_literal`), `EqlPlaintext for NaiveDate` (`Cast::DATE`), the sqlx `chrono` feature + direct `chrono` dep, the CHANGELOG entry, and a temporal-kinds note in the adding-a-scalar reference.

## Verification

- `cargo test -p eql-scalars -p eql-codegen -p eql-tests-macros` — green
- `cargo check -p eql_tests --tests` — clean
- `cargo fmt --check`, `clippy -D warnings` — clean
- `mise run test:matrix:inventory` — catalog reconciled (token-normalized snapshot unchanged; `list-types` → `int4 int2 date`)
- Live PG17 `mise run test:sqlx` — full suite passed incl. the `scalars::date::*` matrix tests and a generated date fixture (16 rows, real HMAC + ORE, `v:2`)

CI covers the full PG 14–17 matrix.


crates/eql-scalars/src/lib.rs
```diff
@@ -502,6 +543,46 @@ mod rust_tests {
...
+        assert!(!ScalarKind::Date.is_int());
+    }
+
+    // The bounded-numeric accessors panic on Date exactly as on the other
+    // non-integer kinds — Date is not an integer kind.
+    #[test]
+    #[should_panic(expected = "min_symbol is only defined for integer kinds")]
+    fn min_symbol_panics_on_date() {
+        ScalarKind::Date.min_symbol();
+    }
```
coderdan, line 581: Not a big deal but this might imply that `min_symbol` and `max_symbol` should be defined in a separate subtrait (can be defined by Int2 etc but not date).

coderdan (approved):

PR #257, opened 2026-06-03
feat(scalars): add eql_v3.timestamptz encrypted-domain type (equality-only)

tobyhede:
## What

Add `eql_v3.timestamptz` as an **equality-only** encrypted-domain scalar (storage + `_eq` via HMAC; `=` / `<>`). Ordering is deferred: cipherstash encrypts `Plaintext::Timestamp` at 12-block ORE width but EQL's only ORE comparator is 8-block, so an ordered domain would silently mis-order.

## Rides the generalized harness (#261)

Rebased onto `v3-scalar-harness-reduction`, the per-type surface is **+259 / 5 files** (was +671 / 10). The whole type is:

- **catalog row** (`eql-scalars`): `ScalarKind::Timestamptz`, `Fixture::Timestamptz`, `EQ_ONLY_DOMAINS`, the `TIMESTAMPTZ` spec + `CATALOG` append.
- **one `temporal_values!` call** (`scalar_domains.rs`):

```rust
temporal_values! {
    cell = TIMESTAMPTZ_VALUES_CELL, accessor = timestamptz_values,
    rust_type = chrono::DateTime<chrono::Utc>, spec = eql_scalars::TIMESTAMPTZ,
    variant = Timestamptz, pg_type = "timestamptz",
    parse = |s| DateTime::parse_from_rfc3339(s).unwrap().with_timezone(&Utc),
    min_pivot = .., max_pivot = .., sql_lit = |v| format!("'{}'", v.to_rfc3339()),
}
```

- **one bare dispatch line** (`scalar_types.rs`): `timestamptz => chrono::DateTime<chrono::Utc>,`
- cast wiring (`eql_plaintext.rs`) + CHANGELOG.

No marker, no second snapshot, no macro change — the catalog's eq-only domain shape drives everything: the emitter picks `scalar_matrix! { caps = [eq] }`, and the inventory derives its eq-only name subset from the single baseline.

## Status / verification

- `cargo test -p eql-scalars` → **43 passed** (incl. `timestamptz_*`)
- matrix inventory → `timestamptz: eq_only` (derived, no second snapshot)
- full matrix (target **895 passed**) + `release/*.sql` identical to the pre-rebase commit — **verification in progress**
- base retarget to `v3-scalar-harness-reduction` + force-push (`--force-with-lease`) — **pending** (depends on #261)


auxesis (approved):
Nice works, thanks @tobyhede.

tests/sqlx/src/scalar_domains.rs
```diff
@@ -205,6 +205,31 @@ temporal_values! {
...
+temporal_values! {
+    cell      = TIMESTAMPTZ_VALUES_CELL,
+    accessor  = timestamptz_values,
+    rust_type = chrono::DateTime<chrono::Utc>,
+    spec      = eql_scalars::TIMESTAMPTZ,
+    variant   = Timestamptz,
+    pg_type   = "timestamptz",
+    parse     = |s| chrono::DateTime::parse_from_rfc3339(s)
+        .expect("catalog timestamptz fixture must be RFC3339")
+        .with_timezone(&chrono::Utc),
```
auxesis, line 223: **Gap: the UTC-normalization contract is never exercised with a non-UTC input.** This `.with_timezone(&Utc)` is what makes the type's headline guarantee (“Values are UTC-normalized”, per the CHANGELOG) true — but every catalog fixture is already `…Z`, and the auto-generated `timestamptz_values::values_match_catalog_fixtures` test re-runs this *exact* closure, so it's tautological: it cannot catch a regression that drops the conversion (e.g. a switch to `.naive_utc()` or constructing the `DateTime<Utc>` from the naive local time). This is the temporal analogue of “the happy path is tested but the transform it claims to perform is not.”

Add a focused test that feeds an offset-bearing RFC3339 string and asserts the resulting instant:

```rust
#[test]
fn rfc3339_offset_is_normalized_to_utc() {
    use chrono::{Datelike, Timelike};
    let parse = |s: &str| chrono::DateTime::parse_from_rfc3339(s)
        .unwrap()
        .with_timezone(&chrono::Utc);
    // 05:00 at +05:00 is midnight UTC — same instant as the Z form.
    assert_eq!(
        parse("2000-01-01T05:00:00+05:00"),
        parse("2000-01-01T00:00:00Z"),
    );
    // …and it lands on the UTC wall-clock, not the offset-local one.
    let utc = parse("2000-01-01T05:00:00+05:00");
    assert_eq!((utc.hour(), utc.day()), (0, 1));
}
```

Expected: passes today; fails the moment the parse path stops converting offsets to UTC.

crates/eql-scalars/src/lib.rs
```diff
@@ -453,6 +496,21 @@ const DATE_FIXTURES: &[Fixture] = fixtures!(date;
...
 
+/// timestamptz fixture plaintexts — RFC3339 UTC strings, parsed into
+/// `chrono::DateTime<Utc>` in the SQLx harness (the catalog stays zero-dep).
+/// The three temporal pivots MUST be present verbatim: `"1900-01-01T00:00:00Z"`
+/// (min_pivot), `"1970-01-01T00:00:00Z"` (zero = `DateTime::<Utc>::default()`,
+/// the Unix epoch), and `"2099-12-31T23:59:59Z"` (max_pivot) — the matrix
+/// fetches each one's ciphertext via `fetch_fixture_payload`, which fails loudly
+/// if a row is absent. The interior timestamps span varied dates AND times of
+/// day so range operators yield distinguishable counts. All distinct.
+const TIMESTAMPTZ_FIXTURES: &[Fixture] = fixtures!(timestamptz;
```
auxesis, line 237: **Gap: timestamptz fixture distinctness is checked by string, but two RFC3339 strings can denote the same UTC instant.** `invariant_tests::fixture_values_are_distinct_by_resolved_number` keys `Fixture::Timestamptz` by its literal (`DistinctKey::Str`), so `"1970-01-01T00:00:00Z"` and `"1970-01-01T01:00:00+01:00"` would pass as “distinct” — yet the fixture *table* keys on the parsed `DateTime<Utc>`, so an aliasing pair would silently insert duplicate `plaintext` rows and break `fetch_fixture_payload`'s `fetch_one`. The string-level distinctness invariant that's sufficient for `int*`/`date` doesn't fully transfer to a tz-bearing type. (Today's fixtures are all `Z` and genuinely distinct — this is a guard against a future edit, e.g. when more values are added.)

`eql-scalars` is deliberately zero-dep (no chrono), so the guard can't live here. Put it in the SQLx harness, where `timestamptz_values()` already gives the parsed slice:

```rust
// tests/sqlx/src/scalar_domains.rs, alongside the temporal_values! call
#[test]
fn timestamptz_fixtures_are_distinct_by_instant() {
    use std::collections::HashSet;
    let vals = timestamptz_values(); // &[DateTime<Utc>], parsed from the catalog
    let unique: HashSet<_> = vals.iter().collect();
    assert_eq!(
        unique.len(), vals.len(),
        "two timestamptz fixtures alias to the same UTC instant",
    );
}
```

Expected: passes today; fails if a fixture is later added whose offset form collides with an existing instant.

auxesis (commented):
Claude identified two gaps, which are posted inline. 

There are two other gaps not posted inline: 

> - **`to_sql_literal` RFC3339 rendering is not unit-pinned.** `tests/sqlx/src/scalar_domains.rs:230` uses `format!("'{}'", v.to_rfc3339())` — a new, timestamptz-specific path distinct from `date`'s `Display`-based `sql_lit`. It's exercised transitively by the matrix (via `fetch_fixture_payload`), but no focused test pins the literal form. A one-liner — `assert_eq!(<chrono::DateTime<chrono::Utc> as ScalarType>::to_sql_literal(Default::default()), "'1970-01-01T00:00:00+00:00'")` — would lock it and catch a stray format change.
> - **`to_plaintext` roundtrip only covers the epoch.** `datetime_utc_to_plaintext_wraps_in_timestamp_variant` (eql_plaintext.rs) uses `DateTime::<Utc>::default()`; a value carrying a time-of-day (e.g. one of the `…T23:20:50Z` fixtures) would be a marginally stronger roundtrip, though the existing `assert_eq!(value, ts)` is already value-generic. Low priority.

auxesis (changes requested):
Missing tests need addressing.

tests/sqlx/src/scalar_domains.rs
tobyhede: Addressed in efe4efb. Added `rfc3339_offset_is_normalized_to_utc` (in `mod timestamptz_value_guards`, alongside the `temporal_values!` call) — it feeds `2000-01-01T05:00:00+05:00` and asserts it resolves to the same instant as the `…Z` form and lands on the UTC wall-clock, so a switch to `.naive_utc()` fails it. The parse closure is mirrored independently of the macro so the regression is caught here rather than re-running the all-UTC catalog fixtures.

crates/eql-scalars/src/lib.rs
tobyhede: Addressed in efe4efb. Added `fixtures_are_distinct_by_instant` in the SQLx harness (zero-dep `eql-scalars` can stay string-keyed): it collects `timestamptz_values()` (parsed `DateTime<Utc>`) into a `HashSet` and asserts `unique.len() == vals.len()`, guarding the by-instant property the fixture table keys on against a future aliasing fixture.

auxesis (approved):

PR #260, opened 2026-06-05
feat(scalars): add eql_v3.text encrypted-domain family (eq / match / ord)

tobyhede:
Adds the `eql_v3.text` scalar encrypted-domain family at parity with EQL v2 encrypted text: equality (HMAC), match (a new self-contained `eql_v3.bloom_filter` SEM index term), and ORE ordering. First scalar to add a new index `Term` (`Bloom`) and the first non-integer, unbounded ordered kind.

## Supported text domains

| Domain | Operators | Index term |
|---|---|---|
| `eql_v3.text` | — (storage only) | — |
| `eql_v3.text_eq` | `=` `<>` | HMAC (`hm`) |
| `eql_v3.text_match` | `@>` `<@` | Bloom filter (`bf`) |
| `eql_v3.text_ord` | `=` `<>` `<` `<=` `>` `>=`, `min`/`max` | ORE block (`ob`) |
| `eql_v3.text_ord_ore` | same as `text_ord` | ORE block (`ob`) |

A real encrypted text payload carries `hm` + `bf` + `ob`; callers cast per predicate. Match is bloom-filter containment on `text_match` — deliberately **not** SQL `LIKE` — and never backs equality (that always routes through `Hm`).

## Notes
- `@>`/`<@` flip from blocker → inlinable wrappers **only** on `Bloom` domains, so the int4 golden is byte-identical (`codegen:parity` green).
- New `eql_v3.bloom_filter` SEM type is self-contained (no `eql_v2` dependency; `test:self_contained_v3` green).
- SQLx matrix generalised for non-`Copy` (`String`) plaintext (`Copy`→`Clone`, `to_sql_literal(&Self)`); `[text]` harness marker mirrors `[temporal]`.

Stacked on `eql_v3`.

## Summary by CodeRabbit

* **New Features**
  * Added support for encrypted text domain (`eql_v3.text`) with equality comparison, bloom-filter-based text matching via containment operators (`@>`/`<@`), and ordering operations (`<`, `<=`, `>`, `>=`, `min`, `max`)


src/v3/sem/bloom_filter/functions.sql
```diff
@@ -0,0 +1,59 @@
...
+--! the extractor below, which gates on value-shape inline, nor by the generated
+--! domain CHECK, which tests `bf` presence via the envelope-key skeleton. Kept
+--! as the canonical presence test for callers that need one.
+CREATE FUNCTION eql_v3.has_bloom_filter(val jsonb)
+  RETURNS boolean
+  IMMUTABLE STRICT PARALLEL SAFE
+  SET search_path = pg_catalog, extensions, public
+AS $$
+  BEGIN
+    RETURN val ? 'bf' AND val ->> 'bf' IS NOT NULL;
```
auxesis, line 29: **Gap: the new public `eql_v3.has_bloom_filter(jsonb)` predicate has no tests.** Its sibling extractor `eql_v3.bloom_filter(jsonb)` gets three direct tests in `family/sem.rs`, but this presence predicate is never exercised — and its own doc comment notes it isn't called by the extractor or the domain CHECK, so nothing covers it transitively either. The two-part guard (`val ? 'bf'` **and** `val ->> 'bf' IS NOT NULL`) has an untested second half: `{"bf":null}` should be `false`.

Sketch — add to `tests/sqlx/tests/encrypted_domain/family/sem.rs` next to the `bloom_filter_extractor_*` tests:
```rust
#[sqlx::test]
async fn has_bloom_filter_detects_bf_presence(pool: PgPool) -> Result<()> {
    // present + non-null array -> true
    let present: bool = sqlx::query_scalar(
        "SELECT eql_v3.has_bloom_filter('{\"bf\":[1,2,3]}'::jsonb)")
        .fetch_one(&pool).await?;
    assert!(present, "present bf array must be true");
    // key absent -> false
    let absent: bool = sqlx::query_scalar(
        "SELECT eql_v3.has_bloom_filter('{\"hm\":\"x\"}'::jsonb)")
        .fetch_one(&pool).await?;
    assert!(!absent, "absent bf must be false");
    // key present but null value -> false (the `->> ... IS NOT NULL` half)
    let null_val: bool = sqlx::query_scalar(
        "SELECT eql_v3.has_bloom_filter('{\"bf\":null}'::jsonb)")
        .fetch_one(&pool).await?;
    assert!(!null_val, "bf:null must be false");
    Ok(())
}
```
Expected: passes as written. The `{"bf":null}` case is the one that distinguishes the `IS NOT NULL` half — it would fail if that half were ever dropped.

auxesis (commented):
Claude caught four gaps in test coverage of new or changed code.

src/v3/sem/bloom_filter/functions.sql
```diff
@@ -0,0 +1,59 @@
...
+--!
+--! @param val jsonb The encrypted payload.
+--! @return eql_v3.bloom_filter The `bf` array as a smallint[] domain value, or
+--!   NULL when `bf` is absent or not a json array.
+CREATE FUNCTION eql_v3.bloom_filter(val jsonb)
+  RETURNS eql_v3.bloom_filter
+  LANGUAGE sql
+  IMMUTABLE STRICT PARALLEL SAFE
+AS $$
+  SELECT CASE WHEN jsonb_typeof(val -> 'bf') = 'array'
```
auxesis, line 56: **Gap: the empty-array branch of `eql_v3.bloom_filter(jsonb)` is untested.** `family/sem.rs` covers `[1,2,3]` → array, missing `bf` → NULL, and `{"bf":null}` (non-array) → NULL — but not `{"bf":[]}`, which hits the `jsonb_typeof(...) = 'array'` branch with an empty array and must return an **empty** `smallint[]`, *not* NULL. The doc comment calls this case out explicitly ("An empty `bf` array yields an empty filter… contained by everything"), and it is the extractor-level basis for the empty-set semantics the smoke tests only assert via literal `text_match` casts.

Sketch — add alongside the other `bloom_filter_extractor_*` tests:
```rust
#[sqlx::test]
async fn bloom_filter_extractor_empty_array_is_empty_not_null(pool: PgPool) -> Result<()> {
    let got: Option<Vec<i16>> = sqlx::query_scalar(
        "SELECT eql_v3.bloom_filter('{\"bf\":[]}'::jsonb)::smallint[]")
        .fetch_one(&pool).await?;
    assert_eq!(got, Some(vec![]), "empty bf array must extract as empty array, not NULL");
    Ok(())
}
```
Expected: passes; distinguishes the empty-array branch (`Some([])`) from the absent/non-array NULL branches already covered.

tests/sqlx/src/scalar_domains.rs
```diff
@@ -205,6 +255,108 @@ temporal_values! {
...
+    const PG_TYPE: &'static str = "text";
+
+    fn fixture_values() -> &'static [Self] {
+        text_values()
+    }
+
+    /// `Display` for a `String` is the unquoted text, which is not a valid SQL
+    /// literal; quote it and double any embedded single quotes.
+    fn to_sql_literal(value: &Self) -> String {
+        format!("'{}'", value.replace('\'', "''"))
```
auxesis, line 374: **Gap: the single-quote escaping branch of `String::to_sql_literal` is never exercised.** This new override doubles embedded single quotes (`value.replace('\'', "''")`), but every `TEXT_FIXTURES` value (`aard`, `frank`, `zzzz`, …) is quote-free, so no test — unit or DB-backed — ever hits the `.replace`. A literal-building escape that no test covers is exactly where a quoting/injection regression slips through unnoticed.

Sketch — a unit test alongside `text_value_tests` in this file:
```rust
#[test]
fn text_to_sql_literal_escapes_single_quotes() {
    let lit = <String as ScalarType>::to_sql_literal(&"O'Brien".to_string());
    assert_eq!(lit, "'O''Brien'");
    // a quote-free value is wrapped but otherwise untouched
    assert_eq!(<String as ScalarType>::to_sql_literal(&"frank".to_string()), "'frank'");
}
```
Expected: passes as written; if the `.replace` were removed it would yield `'O'Brien'` and fail.

crates/eql-scalars/src/lib.rs
```diff
@@ -303,7 +310,10 @@ impl Fixture {
                 Some(k) => Some(k.max_value()),
                 None => None,
             },
-            Fixture::Zero => Some(0),
+            Fixture::Zero => match kind.as_bounded_int() {
```
auxesis: **Gap: the changed `Zero/Min/Max → None` arm for non-integer kinds has no direct test.** This PR changed `Fixture::Zero` from an unconditional `Some(0)` to `match kind.as_bounded_int() { … None => None }`. The existing `numeric_value` tests only assert `Zero → Some(0)` with **integer** kinds, and `numeric_value_is_none_for_string_variants` covers the string *variants* (`Text`/`Numeric`/`Jsonb`/`Date`), not a *sentinel* paired with a non-integer kind. So the new `None` path for `Zero`/`Min`/`Max` on `Date`/`Text` is only guarded indirectly by the `pivot_sentinels_only_appear_with_integer_kinds` catalog invariant — a refactor could silently restore `Some(0)` and stay green.

Sketch — add to the `rust_tests`/`values_tests` module:
```rust
#[test]
fn sentinel_value_is_none_on_non_integer_kinds() {
    // Previously Fixture::Zero returned Some(0) unconditionally.
    assert_eq!(Fixture::Zero.numeric_value(ScalarKind::Date), None);
    assert_eq!(Fixture::Zero.numeric_value(ScalarKind::Text), None);
    assert_eq!(Fixture::Min.numeric_value(ScalarKind::Text), None);
    assert_eq!(Fixture::Max.numeric_value(ScalarKind::Date), None);
}
```
Expected: passes now; on the pre-PR `Fixture::Zero => Some(0)` the two `Zero` cases would have returned `Some(0)` and failed.

auxesis (changes requested):
Nice work @tobyhede. 

Claude has left comments on four test coverage issues that need to be addressed.

src/v3/sem/bloom_filter/functions.sql
tobyhede: Addressed in 71b3c8b — added `has_bloom_filter_detects_bf_presence` to `family/sem.rs`, covering present `{"bf":[1,2,3]}` → true, absent → false, and the `{"bf":null}` → false case that pins the `->> 'bf' IS NOT NULL` half. Passes. Thanks for the catch.

src/v3/sem/bloom_filter/functions.sql
tobyhede: Addressed in 71b3c8b — added `bloom_filter_extractor_empty_array_is_empty_not_null`, asserting `{"bf":[]}` extracts to `Some(vec![])` (empty, not NULL), distinct from the absent/non-array NULL branches. Passes.

tests/sqlx/src/scalar_domains.rs
tobyhede: Addressed in 71b3c8b — added `text_to_sql_literal_escapes_single_quotes` in `text_value_tests`, exercising the `String::to_sql_literal` override directly: `O'Brien` → `'O''Brien'` (and a quote-free value unchanged). Note this pins the trait method itself; the existing `sql_string_literal` test covers the sibling helper. Passes.

crates/eql-scalars/src/lib.rs
tobyhede: Addressed in 71b3c8b — added `sentinel_value_is_none_on_non_integer_kinds`, directly asserting `Fixture::Zero/Min/Max.numeric_value(Date/Text) == None`. Previously only the `pivot_sentinels_only_appear_with_integer_kinds` catalog invariant guarded this indirectly. Passes.

auxesis (approved):

PR #268, opened 2026-06-10
feat: eql-types — TypeScript bindings (stacked on #236)

coderdan:
Stacked on #236 (now merged). Adds the ts-rs layer to the eql-types crate:

- `TS` derive + `#[ts(export, export_to = "v3/")]` on every v3 domain type, the term newtypes, and `Identifier`
- Checked-in generated bindings: `crates/eql-types/bindings/v3/*.ts` (30 files). Term newtypes export as named TS aliases (`export type Hmac256 = string`) imported by every domain binding
- `mise run types:generate` (clean-regenerate) and `mise run types:check` (regenerate + fail on diff/untracked), with a freshness-gate step in the `rust-crates` CI job

JSON Schemas follow in a further stacked PR.

Closes #287.


crates/eql-types/Cargo.toml
```diff
@@ -2,10 +2,11 @@
 name = "eql-types"
 version = "0.1.0"
 edition = "2021"
-description = "Canonical wire types for EQL payloads — the single Rust source of truth (TypeScript bindings and JSON Schemas are generated from these types in stacked changes)."
+description = "Canonical wire types for EQL payloads — the single Rust source of truth, with generated TypeScript bindings (JSON Schemas are generated from these types in a stacked change)."
 
 [dependencies]
 serde = { version = "1", features = ["derive"] }
+ts-rs = { version = "10", features = ["serde-json-impl"] }
```
freshtonic: **`serde-json-impl` looks unused.** Nothing in `src/` references `serde_json::Value`/`Map`/`Number`, so this ts-rs feature — which only adds `TS` impls for those serde_json types — appears to be dead weight pulling `serde_json` into the build graph. (The `serde_json` *dev*-dependency below is for test assertions, unrelated to this feature.) Suggest dropping it, or adding a one-line comment if it's intentional groundwork for the upcoming schemars / JSON-Schema PR.

crates/eql-types/Cargo.toml
```diff
@@ -2,10 +2,11 @@
 name = "eql-types"
```
freshtonic, line 2: **Naming suggestion: consider `eql-types-postgres`.** The domain types here are prefixed with PostgreSQL type names (`int2` / `int4` / `int8` / `timestamptz` / `text`), so the crate is implicitly Postgres-flavoured. Renaming it now leaves room for future database-vendor-native EQL type crates (e.g. `eql-types-<vendor>`) without a disruptive rename later, while there are still few external consumers.

**This is fine as a separate follow-up — it needn't block or fold into this bindings PR.** A rename ripples beyond these changes: the crate is `path`-referenced by `eql-scalars`' dev-dependencies, and the `test:crates` / `types:generate` / `types:check` mise tasks all hardcode `-p eql-types`. Keeping it as its own small PR keeps that churn out of this focused codegen change.

crates/eql-types/README.md
```diff
@@ -56,11 +56,15 @@ the catalog by the JSON Schema parity test in the stacked schemars change.
...
 ```sh
-cargo test -p eql-types
+mise run types:generate   # clean-regenerate bindings/
+mise run types:check      # regenerate + fail if checked-in bindings are stale
 ```
 
-The crate is also part of the lean `mise run test:crates` set (fmt, clippy,
-test — no database).
+Both wrap `cargo test -p eql-types`, which runs the conformance tests and
+regenerates `bindings/` (TypeScript, via ts-rs). The directory is checked in
```
freshtonic, line 64: FYI (no change strictly required): because ts-rs defaults `TS_RS_EXPORT_DIR` to `./bindings`, a plain `cargo test -p eql-types` (and therefore `mise run test:crates`) writes into the checked-in `bindings/` directory **in place** — it can leave a developer's working tree dirty. The temp-dir swap in `types:generate` is the right mitigation for the canonical path; worth a sentence here noting that `test:crates` regenerates `bindings/` as a side effect so it isn't a surprise.

freshtonic (approved):
Approving — **conditional on the three minor fixes flagged inline below.** None are blockers: the substance of the codegen layer is correct and coverage is complete (30 structs → 30 `#[ts(export)]` attributes → 30 checked-in bindings under `bindings/v3/`), the generated TS is accurate, and the regenerate-and-swap script + CI freshness gate are sound.

Three minor items, posted inline:
1. `serde-json-impl` ts-rs feature appears unused (`Cargo.toml`).
2. Crate-naming suggestion: `eql-types` → `eql-types-postgres` (`Cargo.toml`).
3. FYI on the `cargo test` → `bindings/` write side effect (`README.md`).

One extra nit not tied to a line: the PR description says "28 files" but there are **30** `.ts` bindings under `bindings/v3/` — worth correcting so reviewers' file counts line up.

crates/eql-types/Cargo.toml
coderdan: Deserves a separate discussion. IMHO this crate defines the consumer shapes which should be consistent. If it turns out that XYZ DB needs a different shape I'd prefer to create an internal adapter crate and avoid changes to upstream consumers (stack, proxy etc).

PR #269, opened 2026-06-10
feat: eql-types — JSON Schemas (stacked on #268)

coderdan:
Adds the schemars layer to the eql-types crate. Builds on #236 and #268 (both merged); rebased onto `eql_v3`.

- `JsonSchema` derive on every v3 domain type, the term newtypes, and `Identifier`; the registry regains its `schema` fn
- Checked-in generated schemas: `crates/eql-types/schema/v3/<domain>.json` (24 files, named after the SQL domain — the protocol identity), each with a canonical `$id` (`https://schemas.cipherstash.com/eql/v3/<domain>.json`) injected by `tests/export.rs`. Term newtypes appear as named definitions `$ref`'d by every domain schema
- A third parity gate in `catalog_parity.rs`: each published schema's `required` list must equal envelope + catalog term keys (plus a strictness gate)
- `types:generate` / `types:check` and the CI freshness step extended to cover `schema/` (footgun-safe temp-dir swap, honoring `EQL_TYPES_SCHEMA_DIR`)

Closes #288.

### DIFF
--- a/crates/eql-types/src/lib.rs
+++ b/crates/eql-types/src/lib.rs
@@ -85,6 +85,10 @@ impl schemars::JsonSchema for SchemaVersion {
             instance_type: Some(schemars::schema::InstanceType::Integer.into()),
             const_value: Some(serde_json::json!(EQL_SCHEMA_VERSION)),
             metadata: Some(Box::new(schemars::schema::Metadata {
+                // KEEP IN SYNC with the `SchemaVersion` doc comment above — it
+                // is the canonical text. A derived `JsonSchema` would copy the
+                // doc comment automatically; this manual impl can't, so this
+                // hand-written copy must be updated alongside it.
                 description: Some(
                     "The envelope version field (`v`) — always exactly `2` on the wire.".to_owned(),
                 ),
--- a/crates/eql-types/src/lib.rs
+++ b/crates/eql-types/src/lib.rs
@@ -3,9 +3,9 @@
 //! One Rust definition per EQL payload shape — the single source of truth
 //! for every tool that produces or consumes EQL payloads
 //! (`cipherstash-client`, `protect-ffi`, CipherStash Proxy). TypeScript
-//! bindings are generated from these definitions via `ts-rs` (run
-//! `cargo test`, see `bindings/v3/`); JSON Schemas follow in a stacked
-//! change. The Rust types are the contract.
+//! bindings (`ts-rs`) and JSON Schemas (`schemars`) are generated from
+//! these definitions — run `cargo test`, see `bindings/v3/` and
+//! `schema/v3/`. The Rust types are the contract.
 //!
 //! ts-rs rule (learned from the original spike): ts-rs silently drops a
 //! serde attribute it cannot parse, so keep field-level serde attributes
@@ -20,6 +20,7 @@
 //! Wire rule: **field names ARE wire names** — no `#[serde(rename)]`
 //! anywhere. The struct definition reads exactly like the JSON payload.
 
+use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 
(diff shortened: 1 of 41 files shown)


crates/eql-types/src/lib.rs
```diff
@@ -72,10 +73,33 @@ impl<'de> Deserialize<'de> for SchemaVersion {
...
+        "SchemaVersion".to_owned()
+    }
+
+    fn json_schema(_: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        schemars::schema::SchemaObject {
+            instance_type: Some(schemars::schema::InstanceType::Integer.into()),
+            const_value: Some(serde_json::json!(EQL_SCHEMA_VERSION)),
+            metadata: Some(Box::new(schemars::schema::Metadata {
+                description: Some(
+                    "The envelope version field (`v`) — always exactly `2` on the wire.".to_owned(),
```
freshtonic, line 93: **Question: should the on-wire version be `3`, not `2`, in the `eql_v3` tier?** It looks deliberate — the `EQL_SCHEMA_VERSION` doc comment states it's hard-coded to `2` for every payload *including v3*, mirroring the generated `eql_v3` domain CHECK `VALUE->>'v' = '2'`, so the wire envelope version is intentionally decoupled from the `eql_v3` SQL schema namespace (consistent with the project's stance that the schema name is independent of the release version). Assuming that's right, no code change needed — but it's a genuine footgun for schema consumers: someone validating against `v3/<domain>.json` sees a `v3` path and a `const: 2` and will reasonably suspect a mistake. The Rust doc comment explains the decoupling, but schema consumers never see it. Suggest baking the *why* into this published `description` (e.g. "envelope format version `2`; independent of the `eql_v3` SQL schema namespace") so the artifact is self-explaining.

crates/eql-types/src/v3/terms.rs
```diff
@@ -45,6 +46,51 @@ pub struct OreBlockU64_8_256(pub Vec<String>);
...
+            ..Default::default()
+        };
+        SchemaObject {
+            instance_type: Some(InstanceType::Array.into()),
+            array: Some(Box::new(ArrayValidation {
+                items: Some(Schema::Object(items).into()),
+                ..Default::default()
+            })),
+            metadata: Some(Box::new(Metadata {
+                description: Some(
```
freshtonic, line 83: **Description drift risk.** This manual `description` is a hand-maintained near-duplicate of the `BloomFilter` doc comment above (and `SchemaVersion`'s manual impl in `lib.rs` does the same). Every *derived* `JsonSchema` pulls its description straight from the doc comment, so only these two hand-written copies can silently drift from the doc comment they shadow. Worth a one-line comment marking the doc comment as the canonical source, or otherwise flagging the two must be kept in sync.

crates/eql-types/tests/export.rs
```diff
@@ -0,0 +1,36 @@
...
+    let base = std::env::var("EQL_TYPES_SCHEMA_DIR").unwrap_or_else(|_| "schema".into());
+    let dir = format!("{base}/v3");
+    std::fs::create_dir_all(&dir).unwrap();
+    for entry in v3::all() {
+        let mut schema = serde_json::to_value(entry.schema()).unwrap();
+        // schemars 0.8 emits no $id; inject the canonical one.
+        schema.as_object_mut().unwrap().insert(
+            "$id".into(),
+            format!(
+                "https://schemas.cipherstash.com/eql/v3/{}.json",
```
freshtonic: **The `$id` injection is untested.** This canonical URL is the schema's published identity, but no test asserts the format — a regression (wrong host, missing `v3/`, wrong domain name) would only surface via the freshness diff or manual review, never a red test. Consider a small assertion in `catalog_parity.rs` that each domain's published schema carries `$id == https://schemas.cipherstash.com/eql/v3/<domain>.json`.

freshtonic (approved):
Approving. The schemars layer mirrors the ts-rs layer from #268 cleanly, coverage is complete (every domain type + term newtype + `Identifier` derives `JsonSchema`; `DomainType::schema()` is implemented across the board), and I verified the generated schemas against the new tests — all four JSON pointers in `schemas_are_strict` match the actual output.

Standouts worth calling out: the **manual schemas tighten exactly where the derive is too loose** (`SchemaVersion` → `const: 2`; `BloomFilter` → `minimum`/`maximum` i16 bounds rather than the non-validating draft-07 `format: "int16"`), and the **`schemas_are_strict` test** correctly catches the cases the `required`-keys parity test alone would miss (a struct silently losing `deny_unknown_fields`, or `v` swapped for a bare integer). The `types:generate` temp-dir swap is correctly extended to two output trees, and dropping the explicit `trap - EXIT` is fine now since both `mv`s leave an empty `$tmp` for the EXIT trap to clean.

Minor items posted inline:
1. Description drift risk in the two manual `JsonSchema` impls (`lib.rs`, `terms.rs`).
2. The `$id` injection is untested (`tests/export.rs`).
3. A question about the on-wire version value (`lib.rs`).

One trivial nit not tied to a line: the PR description says "23 files" but there are **24** schema files under `schema/v3/` (one per v3 domain). The asymmetry vs the 30 TS bindings is correct and intentional — schemas inline the terms / `Identifier` / `SchemaVersion` as `definitions` rather than as separate files.

crates/eql-types/src/lib.rs
coderdan: Waiting on @tobyhede to give the go-ahead to move to `v3`.

crates/eql-types/src/v3/terms.rs
coderdan: Addressed in 68a977e — added `KEEP IN SYNC` markers on both manual `JsonSchema` impls (`BloomFilter` here, `SchemaVersion` in `lib.rs`), naming the doc comment as the canonical source and noting why these two cant auto-derive it (the derive copies the doc comment automatically; a manual impl cannot).

crates/eql-types/tests/export.rs
coderdan: Addressed in 68a977e — factored the URL into `DomainType::schema_id` (`{SCHEMA_ID_BASE}{domain}.json`), so `tests/export.rs` injects a single source of truth. Added a `schema_id_is_canonical` parity test that pins the shape with independent literals (e.g. `int4_eq` → `https://schemas.cipherstash.com/eql/v3/int4_eq.json`), so a wrong host / dropped `v3/` / wrong domain now fails a red test rather than only shifting the freshness diff.

PR #270, opened 2026-06-10
refactor(v3): clean up eql-codegen/eql-scalars internals (no SQL change)

tobyhede:
## What

Internal cleanup of the `eql-codegen`/`eql-scalars` Rust crates from a code review. **The generated SQL is byte-identical** — nothing in `release/` or `src/` changes, the parity gate proves it. No user-facing change.

Two things actually change:

1. **The parity test now covers every catalog scalar, not just `int4`.** It discovers the reference dirs dynamically, cross-checks that set against `CATALOG` so a type missing its references fails CI (`reference_dirs_match_catalog_tokens`), and asserts two generator runs are byte-identical. (The references themselves are #271.)
2. **Code cleanup** — no behaviour change, caught by the parity gate:
   - Replace the stringly-typed `"ord"` capability check with a `Role` enum, so a typo can't silently disable aggregate codegen.
   - Delete dead code (`Fixture::render_literal`), collapse three duplicate token+suffix impls into one (`DomainSpec::name_with_token`), move `extractor_terms` into `eql-scalars`.
   - `WriteError` implements `std::error::Error` (via `thiserror`); `ScalarKind::rust_type` panics on the arms that have no SQL surface instead of returning junk.
   - Add unit tests for the error and panic paths; commit `matrix_tests_eq_only.txt`.

## Stack

#271 (SQL references) → **this PR**. Depends on #271's files and on `eql_codegen::repo_root`. **Merge #271 first.**

## Verification

- `cargo test -p eql-codegen -p eql-scalars` → 51 passed.
- `cargo test -p eql-codegen --test parity` → 4 passed (byte-identical SQL across all catalog types; two runs deterministic).


crates/eql-codegen/src/consts.rs
```diff
@@ -1,7 +1,15 @@
 //! AUTO-GENERATED headers, schema constants, and SQL-string escaping.
 
-/// SQL generated-file marker. The SQL templates emit this as their first line;
-/// the writer uses it only to recognise files it owns (overwrite/clean safety).
+/// SQL generated-file marker — the header's first line, with no trailing
+/// newline. The writer uses it to recognise files it owns (overwrite/clean
+/// safety) without re-splitting the header on every call.
+pub(crate) const AUTO_GENERATED_MARKER: &str = "-- AUTOMATICALLY GENERATED FILE.";
```
tobyhede, line 7: Why are there two?

PR #267, opened 2026-06-10
feat(v3): self-contained encrypted-JSONB (SteVec) document type `eql_v3.json`

tobyhede:
## Summary

Adds a self-contained encrypted-JSONB (SteVec) document type to the `eql_v3` schema, with **zero `eql_v2` dependency** (CI-gated).

## Stack

Top of the v3 codegen stack: `eql_v3` → #271 (SQL references) → #270 (codegen review hardening) → **this PR**. Merge bottom-up.

- **Types:** `eql_v3.json` (storage/document domain), `eql_v3.ste_vec_entry` (single sv leaf), `eql_v3.ste_vec_query` (containment needle), plus the self-contained SEM index-term type `eql_v3.ore_cllw` (composite + comparators + btree opclass) under `src/v3/sem/ore_cllw/`.
- **Search surface:** document containment (`@>`, `<@`), field/array access (`->`, `->>`, `jsonb_path_query`/`_exists`/`_query_first`, `jsonb_array_length`/`_elements`/`_elements_text`), entry equality (`=`, `<>`) via HMAC, and entry-level ordered range (`<`, `<=`, `>`, `>=`) via CLLW ORE.
- **Firewall (`blockers.sql`):** every native `jsonb` operator reachable through domain fallback is blocked and raises — `?`, `?|`, `?&`, `@?`, `@@`, `#>`, `#>>`, `-`, `#-`, `||`, **and root-document comparisons** `=`, `<>`, `<`, `<=`, `>`, `>=` (so root comparisons raise rather than falling through to native whole-jsonb compare).
- **Indexing:** functional indexes on `eql_v3.eq_term` / `eql_v3.ore_cllw` extractors; `to_ste_vec_query(col)::jsonb jsonb_path_ops` GIN for containment; default btree opclass on `eql_v3.ore_cllw`.
- **Tests:** parameterized SQLx harness (`v3_jsonb_tests`, 65 tests) covering D1–D14, a signature-aware operator-surface guard (`v3_jsonb_operator_surface_tests`), a committed inventory snapshot + CI guard, an in-DB smoke, and a 10-row fixture migration.

### Design note — typed selectors required
[...]

PR #272, opened 2026-06-11
feat(v3): add eql_v3.ore_cllw SEM index-term type and operators

tobyhede:
## What

Adds the self-contained **`eql_v3.ore_cllw`** CLLW-ORE searchable-encrypted-metadata (SEM) index-term type under `src/v3/sem/ore_cllw/`: the type, extractors/comparators, comparison operators, and a btree operator class. No `eql_v2` dependency.

It is a peer of the existing `eql_v3.hmac_256` / `eql_v3.ore_block_u64_8_256` SEM types and provides **entry-level ordering** for the encrypted-JSONB document surface.

## Why split out

This is a foundational SEM primitive with a hard `-- REQUIRE:` boundary — `src/v3/jsonb/{operators,aggregates}.sql` require `src/v3/sem/ore_cllw/operators.sql`. Landing it as its own PR underneath the `eql_v3.json` document surface (#267) keeps that PR focused on the document type rather than mixing in a new index-term primitive.

## Stack

```
eql-v3-codegen-hardening
  └─ eql-v3-ore-cllw            ← this PR
       └─ eql-v3-jsonb-domain   ← #267 (retargeted onto this branch)
```

## Scope

`src/v3/sem/ore_cllw/{types,functions,operators,operator_class}.sql` — 4 new files, ~389 lines, purely additive.

## Summary by CodeRabbit

* **New Features**
  * Added order-revealing encryption (ORE) support for encrypted range queries, enabling efficient comparison operations on encrypted data without decryption.
  * Introduced comparison operators for encrypted values, supporting equality and ordering (=, <>, <, <=, >, >=).
  * Added indexing capabilities for encrypted data with proper ordering semantics.

* **Tests**
  * Expanded validation to ensure encryption support functions remain optimizable across database schemas.


src/v3/sem/ore_cllw/functions.sql
```diff
@@ -0,0 +1,171 @@
...
+    len_a INT;
+    len_b INT;
+    common_len INT;
+    cmp_result INT;
+BEGIN
+    IF a::text IS NULL OR b::text IS NULL THEN
+      RETURN NULL;
+    END IF;
+
+    IF a.bytes IS NULL OR b.bytes IS NULL THEN
```
yujiyokoo, line 131: Just out of curiosity, does this work differently from `IF a::text IS NULL OR b::text IS NULL THEN` just above?

tests/sqlx/tests/ore_cllw_v3_opclass_tests.rs
```diff
@@ -0,0 +1,349 @@
...
+        "Expected Index Scan via ore_cllw_v3_where_test_idx for WHERE range, got:\n{explain}"
+    );
+    assert!(
+        explain.contains("Index Cond"),
+        "Expected Index Cond clause on the WHERE range, got:\n{explain}"
+    );
+
+    tx.rollback().await?;
+    Ok(())
+}
```
yujiyokoo, line 466: I may have missed it, but is there a test case for different input lengths?

yujiyokoo (approved):
Looks good. Added 2 questions

tests/sqlx/tests/ore_cllw_v3_opclass_tests.rs
tobyhede: Good catch, adding

src/v3/sem/ore_cllw/functions.sql
tobyhede: I had to dig into this one - code is ported from v2

Composite types have row `IS NULL` semantics - true when every field is explicitly `NULL`.
Cast forces empty composite to NULL.

Then bytes check is on the inner type.

PR #275, opened 2026-06-11
test(v3): property-based tests for the eql_v3 encrypted scalar domains (CIP-3141)

tobyhede:
## What

Property-based tests for the `eql_v3` encrypted scalar domains: SQL operator results are asserted to agree with a **plaintext oracle** over a generated input space — reaching operator/oracle disagreements the fixed-pivot matrix can't.

## Property coverage

The harness is three property suites plus an example-based edge-case suite — one unit-level **catalog** suite (no DB) and two integration suites, **fixture** (committed ciphertext) and **e2e** (fresh end-to-end encryption each run):

| Suite | File | Proves | Oracle | Scalars | DB |
|-------|------|--------|--------|---------|----|
| **catalog** | `proptest_invariants.rs` | term/operator/extractor catalog consistency; **every** blocker non-STRICT + `plpgsql`; payload-key set == declared terms | pure catalog (Rust) | all (no SQL) | none |
| **fixture** | `property/fixture_oracle.rs` | `=` `<>` `<` `<=` `>` `>=` + `ord_term` sort, over **every ordered pair** | committed ciphertext (real) | int2/int4/int8/date/text/timestamptz | shared |
| **e2e** | `property/e2e_oracle.rs` (`proptest-e2e`) | same engine **+ equality across distinct ciphertexts of an equal plaintext** | fresh ZeroKMS each run | int2/int4/int8 | shared + creds |
| edge cases | `property/edge_cases.rs` | NULL propagation; blockers raise on unsupported ops (incl. native-jsonb `->`/`@>` domain-fallback); timestamptz ordering deferral; CHECK rejects malformed payload | example-based | int4 + catalog-wide | shared |

The fixture, e2e and edge-case suites share one all-pairs oracle engine; the fixture suite generalises across the whole catalog, and the e2e suite is the only one that can hit "same plaintext, different ciphertext" (fixtures have no duplicates).

## Notes

- The proptest suites drive a manual `TestRunner` — `#[sqlx::test]`'s injected pool isn't usable from a sync `proptest!` body. Evaluation is read-only, so no per-test schema isolation; `connect_pool()` + a race-safe once-per-type fixture load into the shared DB.
- Tests live under `property::`, **not** `scalars::` — the matrix-inventory gate reads every `scalars::<X>::` prefix as a scalar type.
- The e2e suite disables shrinking (each shrink = a ZeroKMS batch, ciphertext won't shrink) and failure persistence (nothing to replay).

## Follow-ups

- #273 — extend the e2e oracle to non-integer scalars (date/text/timestamptz)
- #274 — int4 fixture-oracle case-count asymmetry (48 vs 32)

Origin: CIP-3141.

## Tests
* Added comprehensive property-based test suites for encrypted scalar domains across three suites: catalog invariant validation, fixture-corpus oracle testing, and end-to-end encryption property tests. Tests validate equality and ordering operator behavior including edge cases such as NULL propagation and constraint violations.


yujiyokoo (approved):
LGTM

PR #276, opened 2026-06-11
feat(v3): N-block ORE comparator + ordered numeric & timestamptz

### LINKED ISSUE #241: Generalize ORE block comparator to N blocks (not hardcoded 8) for wide-term ordering (decimal/timestamp)
## Problem

`eql_v2.compare_ore_block_u64_8_256_term` hardcodes an 8-block ORE ciphertext, so it cannot order ORE terms wider than 8 blocks. `numeric`/`Decimal` produces **14-block** ORE (96-bit mantissa + scale → encrypted at native width, not padded to 8 like int/float/date), so decimal range / `ORDER BY` / `MIN`/`MAX` silently misorder. Latent until now — no type with a >8-block ORE term had been exercised. Equality (`hm`) is unaffected; only the ORE comparison path is wrong.

## Root cause

`src/ore_block_u64_8_256/functions.sql`, in `compare_ore_block_u64_8_256_term`:
- L156 `right_offset CONSTANT smallint := 136; -- 8 * 17`
- L176 `FOR block IN 0..7 LOOP`

The wire format scales with block count `N`: left CT `17·N` + right CT `16 + 32·N` = `49·N + 16` bytes (int4 N=8 → 408 B; decimal N=14 → 702 B).

## Fix

Derive `N` from the ciphertext length instead of assuming 8:
- `N := (octet_length(a.bytes) - 16) / 49`
- `right_offset := 17 * N`
- `FOR block IN 0..N-1`
- reject terms where `(len - 16) % 49 != 0`; keep the existing equal-length guard.

No-op for N=8 (`(408-16)/49 = 8`, `17*8 = 136`). The block sizes (16/32), nonce offset, indicator math, and the array-level `compare_ore_block_u64_8_256_terms` already generalize.

## Blast radius

Backs all ORE ordering: `eql_v2_encrypted` (`src/operators/{compare,sort,operator_class}.sql`) plus generated int4/numeric domain wrappers.

## Verification

- int4 matrix + `eql_v2_encrypted` ORE tests stay green (no-op for N=8)
- numeric ORE matrix (range / order-by / min-max, ~75 tests) flips green
- add a direct comparator test asserting a known 14-block pair orders correctly

Also unblocks 12-block `timestamp`.

tobyhede:
Generalizes the `eql_v3` ORE block comparator from hardcoded 8 blocks to any `N`, and ships two new **ordered** scalar families on top. Closes #241.

> **Diff size is misleading.** ~4,200 of ~5,000 added lines are committed generated reference SQL (`tests/codegen/reference/numeric/*`, `timestamptz/*`). Real hand-written surface is ~900 lines.

## What changed

- **N-block comparator (the fix).** Derive block count from term length instead of assuming 8:
  ```
  N = (octet_length - 16) / 49
  ```
  - Parametrizes the left-block base offset (`9 → 1+n`) — the bit #241's sketch missed.
  - Raises on any length that isn't `49·N + 16, N ≥ 1`. The `≤ 16` guard is load-bearing: a 16-byte term → `N=0` → silent all-equal fall-through.
  - No-op for `N=8` (int4/int2/int8/date unchanged).

- **Rename** `eql_v3.ore_block_u64_8_256 → eql_v3.ore_block_256` (v3 SEM only — width-honest now `N` varies). `eql_v2.ore_block_u64_8_256` untouched.

- **`eql_v3.numeric`** — new ordered family, 14-block ORE (terms = 702 bytes). Backed by `rust_decimal::Decimal`; `1` and `1.0` collide like Postgres `numeric`.

- **`eql_v3.timestamptz`** — promoted equality-only → ordered, 12-block ORE (terms = 604 bytes). `EQ_ONLY → ORDERED`, nothing else.


coderdan (changes requested):
This isn't correct. The current version of ore.rs is _always_ 8-blocks output. To handle longer values, 8-block chunks are created in an array.

However, the upcoming changes to ore.rs _will_ support a variable length encoding. I suggest holding off on this work until that lands.

coderdan (approved):

PR #280, opened 2026-06-12
feat(eql_v3): text equality via hm + combined text_search domain

tobyhede:
Closes #279.

Additive, pre-release refinement of the unreleased `eql_v3.text` surface: text equality always routes `=`/`<>` through `hm` (exact HMAC), never ORE; the two ordered text domains gain `Hm`, and a new combined `text_search` domain carries eq + ord + match in one type.

## Text domains (`eql-scalars::CATALOG`)

| Domain | Terms | `=` `<>` | `<` `<=` `>` `>=` / min/max | `@>` `<@` |
|---|---|---|---|---|
| `text` | `[]` | — | — | — |
| `text_eq` | `[Hm]` | `hm` | — | — |
| `text_match` | `[Bloom]` | — | — | `bf` |
| `text_ord` | `[Hm, Ore]` | `hm` | `ob` | — |
| `text_ord_ore` | `[Hm, Ore]` | `hm` | `ob` | — |
| **`text_search`** *(new)* | `[Hm, Ore, Bloom]` | `hm` | `ob` | `bf` |

Invariant: every eq-capable text domain leads with `Hm`, so first-supporting-term-wins resolves `=`/`<>` to `eq_term`. Integer ordered domains keep `[Ore]`-only equality (lossless for ints).

## SEM index terms (`eql_v3`, self-contained)

| Term | Payload key | SEM type | Extractor | Provides |
|---|---|---|---|---|
| `Hm` | `hm` | `eql_v3.hmac_256` | `eq_term` | `=` `<>` |
| `Ore` | `ob` | `eql_v3.ore_block_u64_8_256` | `ord_term` | `=` `<>` `<` `<=` `>` `>=` |
| `Bloom` | `bf` | `eql_v3.bloom_filter` | `match_term` | `@>` `<@` |

## Changes
- **Catalog/codegen:** `Term::provides_ordering()`; `is_ord_capable` gates `min`/`max` on any ordering term (a `[Hm, Ore]` domain now emits aggregates). Regenerated + committed `tests/codegen/reference/text/`.
- **Test harness:** `Variant` term sets derived from `CATALOG` (+ `Search` variant, resolution backstop); `bf` in `PLACEHOLDER_PAYLOAD` + catalog-derived `fixture_shape`; `ord_routes_through_ob` proves text `=` engages the `eq_term` btree (never ORE); `text_search` rides the full matrix (eq + ord + min/max + full bloom `@>`/`<@` correctness/GIN via new `MatchScalar`).
[...]

PR #282, opened 2026-06-16
refactor(v3): rename ore_block_u64_8_256 → ore_block_256 (SEM type, repo-wide)

tobyhede:
## What

Behaviour-preserving rename of the self-contained **`eql_v3`** SEM index-term type `ore_block_u64_8_256` → **`ore_block_256`**, reflecting that the block-ORE wire format is width-agnostic (the comparator derives the block count `N` from the ciphertext length; that generalization lands in the stacked N-block PR). This is split out so the mechanical rename reviews on its own.

The **`eql_v2` public API is deliberately unchanged** — `eql_v2.ore_block_u64_8_256`, `src/ore_block_u64_8_256/`, the v2 operators, and all v2 docs/tests keep the original name.

## Scope (eql_v3 only)

- `src/v3/sem/ore_block_u64_8_256/` → `src/v3/sem/ore_block_256/` (directory + internal symbols)
- `eql-scalars` `Term::Ore` ctor + `REQUIRE` paths; `eql-codegen` doc/assertion
- Committed codegen goldens (`int2`/`int4`/`int8`/`date`/`text` incl. `text_search`)
- `eql-types`: `OreBlockU64_8_256` newtype → `OreBlock256` + SQL-constructor doc
- `tasks/pin_search_path.sql` (v3 block only), `splinter.sh` (v3 rows only), `clean_install_v3.sh`, `drop_operator_classes.sql` (v3 opclass only)
- v3 SQLx family tests (`sem`/`mutations`/`inlinability`) — the `eql_v2.compare_ore_block_u64_8_256_terms` parity reference in `sem.rs` is preserved
- v3 docs (CLAUDE, adding-a-scalar, eql-functions, sql-support, analysis) + CHANGELOG `eql_v3.`-qualified mentions

Covers the new old-name references introduced by #280 (text-search).

**Deliberately not renamed:** `tests/sqlx/src/index_types.rs ORE_BLOCK_U64_8_256` — it's the crypto wire index-term identifier, not the SQL type name.

## Verification

- `codegen:parity` — byte-for-byte ✅
- `test:crates` — fmt + clippy + tests ✅
- clean build ✅; the `release/cipherstash-encrypt-v3.sql` artifact carries no `eql_v2` symbol and no old SEM name
- DB-backed `clean_install_v3` runs in CI (Docker unavailable locally)

## Stacking

The N-block ORE comparator + ordered `numeric`/`timestamptz` work stacks on top of this branch.


coderdan (commented):
The upcoming release of ore.rs introduces a 6-bit block size as the default. An 8-bit will still be available for backwards compatibility and for specific domains where a different leakage profile is needed.

Comparison will be different for 6-bit vs 8-bit blocks and
I can't guarantee that comparison functions will be the same for fixed length vs variable length (still WIP).

PR #295, opened 2026-06-18
feat(eql_v3): add `bool` storage-only encrypted-domain type

tobyhede:
## What

`eql_v3.bool` — the **first storage-only scalar**: encrypted at rest, decrypted by the proxy, **never searchable server-side**.

- One term-less `eql_v3.bool` jsonb domain. No `_eq`, no `_ord`, no SEM term.
- Payload is `{v,i,c}` only — no `hm`/`ob`/`bf`.
- Every comparison/containment/path operator reachable via domain fallback is a `plpgsql` blocker.

**Why no index?** 2-value cardinality → any searchable index (even HMAC eq) leaks the plaintext distribution. So `bool` offers zero search surface by design.

New domain *shape*: strictly below eq-only, first family member with no comparison capability.

## How

One catalog row drives it — codegen already handled a zero-term single-domain type, **no generator change**:

```rust
pub const BOOL: ScalarSpec = ScalarSpec {
    token: "bool",
    kind: ScalarKind::Bool,
    domains: STORAGE_ONLY_DOMAINS, // [] — no _eq/_ord
    fixtures: BOOL_FIXTURES,        // false, true
};
// CATALOG: [..., TEXT, BOOL]
```

Generated SQL (committed reference baseline `tests/codegen/reference/bool/`): 3 files, 44 blockers, no aggregates.

```sql
CREATE DOMAIN eql_v3.bool AS jsonb
  CHECK (jsonb_typeof(VALUE) = 'object'
    AND VALUE ? 'v' AND VALUE ? 'i' AND VALUE ? 'c'
    AND VALUE->>'v' = '2');
```


auxesis (approved):

auxesis (commented):
## Test-coverage review — PR #295 (`eql_v3.bool` storage-only scalar)

This PR is unusually well-tested: nearly every added code path ships with a matching test in the same change.

- **`eql-scalars` catalog** (`ScalarKind::Bool`, `Fixture::Bool`, `BOOL` spec, `is_storage_only()`): covered by `bool_spec_is_storage_only_encryption_only`, `storage_only_is_exclusive_to_bool` (sweeps the whole catalog), the `every_type_uses_a_known_domain_shape` storage-only arm, and the `distinct_key` Bool branch via the `fixture_values_are_distinct_by_resolved_number` catalog sweep.
- **`eql-tests-macros` routing** (`is_storage_only_token`, storage-first ordering in `matrix_suite_for_entry`): covered by `storage_only_is_read_from_catalog`, `storage_only_entry_emits_scalar_matrix_with_storage_caps_only`, and `bool_entry_skips_impl_and_stamps_storage_fixture`.
- **SQLx harness** (`EqlPlaintext for bool`, `storage_only()` on `FixtureSpec`, `scalar_fixture!(storage,...)`, `token_is_storage_only`, the `caps = [storage]` matrix arm): covered by new unit tests plus the committed `matrix_tests_storage_only.txt` inventory snapshot.
- **Generated SQL surface** (`tests/codegen/reference/bool/*.sql`): pinned byte-for-byte by the catalog-driven `crates/eql-codegen/tests/parity.rs`.
- **`eql-types` schema/bindings** (`bool.json`, `Bool.ts`): regenerated by `export.rs`, and the schema-level contract for `bool::Bool` is exercised by the catalog-iterating `catalog_parity.rs` tests (required keys, strictness, `$id`).

I found **one** gap worth closing — see the inline comment.

### Out of scope (noted, not posted inline)
No crypto/security concerns introduced; the storage-only contract (no index term on a 2-value column) is the intended behaviour and is asserted in the matrix `fixture_shape` arm.

crates/eql-types/src/v3/bool.rs
```diff
@@ -0,0 +1,50 @@
...
+use crate::{Identifier, SchemaVersion};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+/// `eql_v3.bool` — storage only / encryption-only; every operator is blocked.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, TS, JsonSchema)]
+#[ts(export, export_to = "v3/")]
+#[serde(deny_unknown_fields)]
+pub struct Bool {
```
auxesis, line 28: **Gap:** `bool::Bool` is the only *hand-written* (non-generated) `DomainType` struct added here, yet it is absent from `tests/v3_conformance.rs::non_int4_tokens_round_trip_every_domain` — the behavioural wire-shape sweep that round-trips every other non-int4 token. That test exists precisely because `catalog_parity.rs` "checks domain *names* only, never the wire shape" (its own comment), so a copy-paste field typo or a dropped `#[serde(deny_unknown_fields)]` in `bool.rs` would ship green. Schema reflection in `catalog_parity` would catch a renamed *required* key, but not a serde round-trip regression or stray-term acceptance — exactly the storage-only invariant (no index term ever) that matters most here.

Add `Bool` to the existing sweep (the `storage` wire builder is already in scope), plus a storage-only negative:

```rust
// in non_int4_tokens_round_trip_every_domain, with the other round_trip! lines:
use eql_types::v3::bool::Bool;
round_trip!(Bool, storage("a"), "eql_v3.bool");

// storage-only contract: the payload must never carry an index term.
let with_term = json!({
    "v": 2, "i": { "t": "a", "c": "x" }, "c": "ct", "hm": "deadbeef"
});
let result: Result<Bool, _> = serde_json::from_value(with_term);
assert!(result.is_err(), "Bool must reject any index term (storage-only)");
```

Expected: both pass against the current `bool.rs` (the round-trip is byte-for-byte; the negative relies on `deny_unknown_fields`); the round-trip fails to compile/assert if a `v/i/c` field drifts, and the negative fails if `deny_unknown_fields` is ever removed.

coderdan (approved):

auxesis (approved):
@tobyhede approved, thank you for this. 

Have identified one test coverage issue that needs addressing before merge.

PR #337, opened 2026-07-02
feat(v3): split internal eql_v3 objects into eql_v3_internal schema

tobyhede:
## Summary

Splits every INTERNAL EQL v3 SQL object (SEM index-term types + support, scalar operator-backing wrappers/blockers/state functions, jsonb containment engine, validators, native-jsonb blockers) out of the public `eql_v3` schema into a new `eql_v3_internal` schema, leaving `eql_v3` as a clean public API surface.

- Creates `eql_v3_internal` (+ uninstall/reset drops).
- Codegen (`eql-codegen`) now emits generated wrappers/blockers/aggregate state functions and index-term constructors into `eql_v3_internal` via a new `INTERNAL_SCHEMA` constant; the scalar SQL surface under `src/v3/scalars/` is regenerated accordingly.
- Hand-written SEM index-term types (`hmac_256`, `bloom_filter`, `ore_block_256`(+`_term`), `ore_cllw`) and support move into `eql_v3_internal`.
- Shared blocker helpers (`encrypted_domain_unsupported_*`) and the jsonb internals (SteVec validators, containment engine, `jsonb_blocked_*`, `jsonb_array_to_bytea_array`) move into `eql_v3_internal`.
- `eql_v3` retains only the public surface: domains, query operators, index extractors (`eq_term`/`ord_term`/`match_term`), `min`/`max` aggregates, the `json → ste_vec_query` cast, `version()`, `lints()`.
- `tasks/pin_search_path_v3.sql`, `tasks/test/splinter.sh`, and `eql_v3.lints()` widened to cover both schemas.
- Bindings prose (`crates/eql-bindings`) and generated TS/JSON regenerated to name `eql_v3_internal` SEM constructors.
- SQLx test suite requalified for the moved symbols, including several structural test files not anticipated by the original plan (`inlinability.rs`, `edge_cases.rs`, `lint_tests.rs`, a shared `property.rs` oracle helper) whose hardcoded `eql_v3`-only schema scans either silently stopped testing anything or actively failed post-move.
- CHANGELOG entry, `docs/upgrading/v3.0.md` (U-001 grant requirement, U-002 reinstall requirement), and CLAUDE.md architecture prose updated.
- Installer does **not** auto-grant `eql_v3_internal` access — mirrors the existing `eql_v3` opt-in grant pattern (documented in README + upgrade note), not the wider `GRANT ... TO PUBLIC` the original plan draft had briefly proposed.

## Why

[...]

PR #340, opened 2026-07-02
feat(eql_v3): CLLW-OPE ordering term, v:3 envelope bump, from_v2 conversion + fixture routing

freshtonic:
Adds the CLLW-OPE (`op`) ordering term across every ordered scalar family, then bumps the `eql_v3` JSON envelope version to `v: 3`.

Linear: [CIP-3340](https://linear.app/cipherstash/issue/CIP-3340)

## ✅ Merge gate satisfied — stack collapsed into this PR

This PR now carries the whole stack: the original OPE + envelope work, **#341** (`eql_bindings::from_v2` — EQL v2.3 → v3 wire conversion, [CIP-3341](https://linear.app/cipherstash/issue/CIP-3341)) and **#343** (sqlx fixture generation routed through `from_v2`, [CIP-3347](https://linear.app/cipherstash/issue/CIP-3347)). #341/#343 show as merged (their head landed in this branch). The original gate — fixtures failing the `v = '3'` CHECKs — is resolved in-PR: the pinned client's `v: 2` output is converted through `from_v2` before fixtures are written, consistent with the no-client-changes architecture.

The branch also merges `eql_v3` post-#337 (the `eql_v3_internal` schema split): the `ope_cllw` SEM fork (bytea domain + jsonb constructor) moved to `eql_v3_internal` to match its `hmac_256`/`ore_cllw` siblings, the `*_ord_ope` codegen surface was regenerated (public `eql_v3.ord_ope_term` unchanged), and the public-surface golden was extended with the ord_ope objects.

## `<T>_ord_ope`: CLLW-OPE ordering domains (9 families)

Every ordered scalar family — `int2`, `int4`, `int8`, `date`, `timestamp`, `numeric`, `text`, `float4`, `float8` — gains an `_ord_ope` domain backed by the new `op` index term (a hex-encoded CLLW-OPE ciphertext; revives the v2.2-era `opf`/`opv` order-preserving terms under the single scalar `op` key cipherstash-client re-emits, CIP-3280). `text_ord_ope` carries `[hm, op]` so `=`/`<>` stay exact via HMAC; the integer/date/numeric/float families carry `[op]` alone. `bool` stays storage-only.

**bytea-domain SEM design.** Unlike the block-ORE terms (custom N-block comparison protocol), the OPE ciphertext is order-preserving under plain byte comparison. The new SEM type `eql_v3.ope_cllw` is a **domain over `bytea`**: the `eql_v3.ord_ope_term(col)` extractor hex-decodes `op` and the comparison chain reduces to native `bytea` operators plus the DEFAULT btree opclass — every function in the chain stays `LANGUAGE SQL` and inlinable, so a plain functional btree index (`CREATE INDEX ... (eql_v3.ord_ope_term(col))`) **engages structurally**, verified by an EXPLAIN-based test that asserts an Index Scan (not a seq scan) for `ORDER BY`/range predicates through the extractor. That functional-index engagement is the reason for the bytea-domain design: a composite SEM type with a custom comparator (the ore_cllw pattern) can't be matched by the planner without an operator class, which managed Postgres (Supabase) installs can't use.

- extractor is `ord_ope_term` (not `ord_term` — `dedupe_terms_by(extractor)` would collapse a mixed `[Ore, Ope]` domain); ctor/SEM type `ope_cllw`; binding newtype `OpeCllw`
- hand-written `src/v3/sem/ope_cllw/{types,functions,operators,operator_class}`, modeled on `ore_cllw` but reduced to hex-decode + native bytea comparison
- `pin_search_path_v3.sql` keeps the ope extractor/operator functions inlinable; `splinter.sh` allowlists the new unpinned names
- regenerated scalar SQL surface, Rust bindings, inventory, TS bindings and JSON Schemas; `capability_label` maps `ord_ope` to "ordering domain"
- literal-payload sqlx smoke suite (`encrypted_domain/ope`) with hand-built `op` hex fixtures: ordering by decoded bytes, hm-routed text equality, blockers, domain CHECK, extractor ORDER BY, `ORDER BY ... USING` rejection, min/max; per-type `<T>_ord_ope` smoke modules wired into catalog-coverage

## Envelope bump to `v: 3`

Every `eql_v3` domain CHECK — the generated scalar families (via `types.sql.j2`) and the hand-written `eql_v3.json` SteVec document domain — now pins `VALUE->>'v' = '3'`, and the canonical bindings reject anything else at the type boundary (`EQL_SCHEMA_VERSION = 3`, the `SchemaVersion` newtype/TS alias, and the JSON Schema `const`, all regenerated). The legacy `eql_v2` wire documentation (`eql-payload-v2.*.schema.json` and its validation tests) is unchanged and stays `v: 2`. **BREAKING** for the v3 tier: payloads written to `eql_v3` domain columns must carry `v: 3` — see the merge gate above. Upgrade note U-001 in `docs/upgrading/v3.0.md`.

## Rebase note

Rebased onto `eql_v3` post-#336 (Shape collapse `af9e5316`, `k` form discriminator `b0dd27bc`). While rebasing, a latent bug in the envelope commit was fixed: the jsonb D9 wrong-version negative had been left at `v: 3` (valid post-bump — the reject test would have failed against the accepts-valid twin); it now uses the legacy `v: 2`.

## Verification

All run at the stacked head (`feat/eql-bindings-from-v2`, which contains these commits) and re-run at this branch head:

- `mise run test:crates` — pass (fmt + clippy + all crate tests, 17 suites)
- `mise run codegen:parity` — pass (committed scalar SQL surface matches the catalog, byte-identical)
- `mise run types:check` — pass (no bindings/schema drift)
- `mise run test:schema` — pass (28 at this head / 32 at stacked head)
- `mise run test:codegen` — pass
- `mise run test:matrix:catalog-coverage` — pass (every CATALOG (type, domain) has matrix tests; run at stacked head)
- `mise run test:self_contained_v3` — pass (no `eql_v2` symbols under `src/v3`)

Not run: full `mise run test:sqlx` fixture-backed arm (needs CS creds — see merge gate).


tobyhede (commented):
> Agent-assisted review. Each finding below was independently re-verified against the head commit `cbc309c5` by re-reading the referenced files and re-deriving counts/line numbers. One sub-claim was corrected during verification (noted inline in the Blocker); one anchor location was corrected in Major 1. Inline comments were not possible: the Blocker's anchor files are not touched by this PR's diff — that absence is the finding.

## Blocker

**The `v: 3` bump leaves the fixture-backed suites unable to pass under any client — two stale `v = '2'` assertions were never swept, and a third pins the old client by design but sits in the same failure path.**

- `tests/sqlx/src/matrix.rs:1959-1964` — the fixture-shape arm asserts every fixture payload declares `v = '2'`, while every domain CHECK now pins `VALUE->>'v' = '3'` (e.g. `src/v3/scalars/int4/int4_types.sql:20`).
- `tests/sqlx/tests/eql_v3_int4_fixture_tests.rs:136-151` — `every_payload_declares_eql_payload_version_v2` is an explicit tripwire whose own comment says a bump to v3 "fails this test loudly, forcing the maintainer to regenerate the fixture and audit consumers." The bump landed without acting on it.
- `tests/sqlx/src/fixtures/cipherstash.rs:324-341` — asserts the live client emits `v = 2`. *Correction from the draft finding:* commit `6d6d0cee`'s message explicitly lists this one as "LEFT AT v: 2 by design" (it pins what the pinned cipherstash-client 0.35 emits, and the live tests are `#[ignore]`d), so it is deliberate rather than missed — but it must flip in the same future sweep, since the fixture suites require exactly the updated client that will make it fail. The matrix.rs and eql_v3_int4_fixture_tests.rs assertions are **not** in the commit's deliberate-exceptions list.

Net effect: the fixture tables themselves are plain `jsonb` so INSERTs succeed, but the fixture-backed tests cast payloads into the domains (e.g. `matrix.rs:2894` — `({col})::{d}`), where a pinned-client `v: 2` payload fails the new `v = '3'` CHECK. With a future v3-emitting client, the two assertions above fail instead. The fixture matrix, `property::fixture_oracle`, `property::cross_ciphertext`, and the e2e suites cannot go green in either world without a further change. The commit message acknowledges the client dependency; the in-repo `v = '2'` assertions also need the sweep (or at minimum a tracked follow-up tying all three together).

## Major

1. **All OPE runtime coverage is synthetic hex — the two load-bearing crypto assumptions are untested.** This is a documented, deliberate, temporary exception (`tests/sqlx/tests/encrypted_domain/ope/support.rs:11-23`, `crates/eql-domains/src/lib.rs:49-60`, CIP-3280): the pinned client doesn't emit `op` yet. What rides on it: integer-family `_ord_ope` domains carry `[Ope]` alone (`crates/eql-domains/src/lib.rs:191-194`), so `=`/`<>` route through `op` — correct only if CLLW-OPE is deterministic. A randomized real term would make equality silently return false negatives while every committed test stays green (all ope tests hand-build hex via `ope_cast`). Nothing forces follow-through: the fixture pipeline is entirely OPE-unaware — `IndexKind` (`tests/sqlx/src/fixtures/index_kind.rs:17-28`: Unique/Ore/Match/SteVec), `FixtureSpec`, and the harness `Variant` enum (`tests/sqlx/src/scalar_domains.rs:1176-1182`: Storage/Eq/Ord/OrdOre/Search) have no OPE members — so when the client lands, no gate fails to remind anyone to add real-ciphertext `ord_ope` coverage. Recommend a tracked issue plus a tripwire test (e.g. assert the generated fixture payloads do NOT carry `op`, so the first client that emits it fails loudly).

2. **The reference doc is stale against the change it now describes.** The `Ope` term-table row in `docs/reference/adding-a-scalar-encrypted-domain-type.md:161` is accurate, but the rest of the doc wasn't swept:
   - envelope version still pinned to `'2'` in four places (lines 547, ~551, 840, 950) while `types.sql.j2` now emits `'3'`;
   - generated-surface arithmetic says eleven files / four domains / 176 operators for int4 (lines ~756-759) and `D = 4, A = 2` (lines ~832-834), where int4 is now 14 files / 5 domains / 220 `CREATE OPERATOR` statements (44 × 5, verified by grep) / `D = 5, A = 3`;
   - the wrapper/blocker count table (lines ~620-627) has no `&[Term::Ope]` or `&[Term::Hm, Term::Ope]` rows;
   - the catalog-row example comment (line 102: "storage, _eq (hm), _ord_ore (ore), _ord (ore)") omits `ord_ope`;
   - the SEM-type enumerations (lines 15, 742, 788) list `hmac_256` / `ore_block_256` / `bloom_filter` but omit `ope_cllw`;
   - the indexing guidance never shows the new `USING btree (eql_v3.ord_ope_term(col))` recipe — the very point of the bytea-domain fix in `9291ce9d` (`ord_ope_term` appears only in the term table at line 161).

## Minor

- **Test-name pinning is per-domain, not per-test, for ope.** The catalog-coverage wiring (`9fa810d3`) works — deleting a whole `<t>_ord_ope` module, or a future type missing its wiring, fails CI — but ord_ore has 83 individually pinned test names in `tests/sqlx/snapshots/matrix_tests.txt` (plus 84 in `matrix_tests_text.txt`) while ope has zero pinned names. Any individual ope test — including the index-engagement regression test from `9291ce9d` (`ope/int4_ord_ope.rs:12-56`) — can be deleted as long as one test per module survives. Follows the `text_match`/`signed` precedent but is materially weaker-pinned than ord_ore. `tests/sqlx/snapshots/README.md` also doesn't mention the nine ope modules.
- **Depth gaps vs ORE** (interim, given the synthetic-payload situation, all verified): NULL propagation through ope operators is untested (no NULL cases in `tests/sqlx/tests/encrypted_domain/ope/`); `<=`/`>=` result semantics are never asserted — the shared smoke covers `<` `>` `=` `<>` only, and int4's five-operator loop (`ope/int4_ord_ope.rs:41-53`) is EXPLAIN-only via `assert_index_scan_uses`; blockers are tested only as `@>` on int4 (`ope/int4_ord_ope.rs:73-87`); there is no SEM-layer test for the `eql_v3.ope_cllw(jsonb)` extractor (missing-`op` → NULL, invalid-hex behavior) while hmac_256/ore_block_256/bloom_filter have sections in `encrypted_domain/family/sem.rs`; wrong-envelope-version rejection is untested at the SQL layer for scalar domains — the matrix `payload_check` tests (`tests/sqlx/src/matrix.rs:1083-1127`) strip keys and test non-object but never mutate the `v` *value*; only jsonb SteVec has an explicit `v: 2` negative (`tests/sqlx/tests/v3_jsonb_tests.rs:~1138`) plus the Rust-layer `v3_conformance.rs`; index engagement is tested on int4 only; text's hm-equality/op-ordering split is proven semantically (`ope/text_ord_ope.rs`) but not at plan level (no EXPLAIN there).
- **Missing-term asymmetry between the two ordering extractors:** `eql_v3.ore_block_256(jsonb)` RAISEs on a missing `ob` (`src/v3/sem/ore_block_256/functions.sql:~64-77`) while `eql_v3.ope_cllw(jsonb)` returns NULL on a missing `op` (`src/v3/sem/ope_cllw/functions.sql:~25-30`). The choice is justified (a RAISE would force plpgsql and kill inlining, defeating the functional-index design) and documented in the file, but raw-jsonb callers get different error semantics from the two ordering extractors — worth a line in the reference doc or changelog.
- **`text` `search` domain not extended with `Ope`** — stays `[Hm, Ore, Bloom]` (`crates/eql-domains/src/lib.rs:329-333`). Presumably deliberate while the client can't emit `op`, but the decision is undocumented: neither the `TEXT_DOMAINS` doc comment (which does explain the Hm-precedence rule) nor the #340 changelog entry says why the combined-capability domain skips OPE ordering.

PR #368, opened 2026-07-06
docs: jsonb domains + per-domain terms/extractors in the manifest (closes #365, #366)

### LINKED ISSUE #365: Manifest: include the hand-written jsonb domains (json / jsonb_entry / jsonb_query)
Follow-up from #364 (the structured `eql-manifest.json`).

The manifest's `domains` array is now sourced from the Rust catalog via `eql-codegen dump-catalog` (authoritative — the source of truth the SQL is generated from). That covers the **48 generated scalar domains**, but the **JSON domains are hand-written SQL** (`src/v3/jsonb/types.sql`: `eql_v3.json`, `eql_v3.jsonb_entry`, `eql_v3.jsonb_query`) and are not in `eql_domains::CATALOG`, so they're missing from the manifest.

**Impact:** the docs V2 variant matrix and the docs-side drift-lint (cipherstash/docs#46) don't cover the JSON / ste_vec domains.

**Options:**
1. Add the JSON domains to the catalog so `dump-catalog` emits them alongside the scalars (most consistent — one source of truth). May not fit if the ste_vec surface is intentionally hand-authored.
2. Have the manifest generator also parse the hand-written `src/v3/jsonb/types.sql` for those 3 domains and merge them in (keeps them hand-authored; small and self-contained).
3. Document the JSON domains only in the hand-written `json.mdx` reference page and exclude them from the generated matrix.

**Recommendation:** (2) as a small immediate step, or (1) if the ste_vec domains should live in the catalog long-term.

Context: PR #364, cipherstash/docs#46.

coderdan:
**Stacked on #364.** Closes #365 and #366 — both achieved through the catalog dump, no SQL autogen.

## #365 — jsonb domains in the manifest

Turns out there's nothing to "move": `json` / `jsonb_entry` / `jsonb_query` are **already** in `eql_domains::CATALOG` as `Shape::SteVec` entries. Their SQL is deliberately hand-written (`src/v3/jsonb/`) with performance-tuned CHECKs (validator functions, inline expressions — #354), and the catalog comment is explicit that "the SQL generator SKIPS these; the catalog drives only their inventory." So **full SQL autogen isn't the design** (and would fight the hand-tuning) — the gap was simply that `scalar_families()` filters them out of `dump-catalog`.

Fix: `dump_catalog()` now emits them as a new **additive `stevec` field** on `CatalogDump`. Existing scalar-only consumers (the fixture-coverage task in `mise.toml`, the dump tests) read only `types[]`, so they're untouched. The manifest now carries **all 51 domains** (48 scalar + 3 jsonb).

## #366 — per-domain terms + extractors

Each `DomainEntry` gains a `terms` field (`key` + `extractor` + `ctor`, straight from `eql_domains::Term`), so the manifest links a domain to its extractor functions:

```jsonc
{ "segment": "ord", "supported_ops": ["=","<>","<","<=",">",">="],
  "terms": [{ "key": "ob", "extractor": "ord_term", "ctor": "ore_block_256" }] }
```

The docs manifest maps this to `termFunctions` (e.g. `integer_ord` → `eql_v3.ord_term`) — authoritative, and it resolves the docs drift-lint false-flags on `eq_term` / `ord_term` / `match_term` (real generated extractors that doxygen didn't capture cleanly).

## Safety

- Both fields are **additive**; nothing existing changes shape.
- `cargo test -p eql-codegen -p eql-domains` — **89 + 91 pass**, incl. new tests (`stevec_jsonb_family_is_dumped`, `ordered_domain_exposes_its_extractor_and_ctor`).
- End-to-end verified: `doxygen → XML` + `dump-catalog` → **984 functions + 51 domains**.

Docs-side consumer updated in cipherstash/docs#46.

https://claude.ai/code/session_01CqDNqLSEEkCi7xAJFq7HJA

### DIFF
--- a/crates/eql-codegen/src/dump.rs
+++ b/crates/eql-codegen/src/dump.rs
@@ -68,6 +68,9 @@ pub struct SteVecEntry {
     pub full_name: String,
     /// The catalog domain name: `json` / `entry` / `query`.
     pub name: &'static str,
+    /// Index terms for this SteVec domain. Non-empty only for `jsonb_entry`
+    /// (the sv element type); the `json` container and `jsonb_query` domains
+    /// carry no term extractors — see `stevec_terms`.
     pub terms: Vec<TermInfo>,
 }
 
@@ -82,16 +85,24 @@ fn term_infos(terms: &[Term]) -> Vec<TermInfo> {
         .collect()
 }
 
-/// Index terms for the `jsonb` (SteVec) family, hardcoded for now.
+/// Index terms for one `jsonb` (SteVec) domain, hardcoded for now.
 ///
 /// The catalog does not model per-SteVec-entry terms — `JSONB_DOMAINS` declare
 /// `terms: &[]` and the `shape_and_terms_are_consistent` invariant fails CI if a
 /// non-`Scalar` domain ever gains one — so `term_infos(d.terms)` is provably
-/// empty here, which would render the searchable SteVec family inert in the
-/// manifest. Until the catalog carries them, source the real hand-written
-/// extractors (`src/v3/jsonb/operators.sql`): every sv entry carries `hm`
-/// (hash-equality, `eql_v3.eq_term`) or `oc` (CLLW-ORE ordering, `eql_v3.ore_cllw`).
-fn stevec_terms() -> Vec<TermInfo> {
+/// empty here. Until the catalog carries them, source the real hand-written
+/// extractors from `src/v3/jsonb/{functions,operators}.sql`.
+///
+/// Terms live on `jsonb_entry` — the sv *element* type — ONLY: `eql_v3.eq_term`
+/// reads `coalesce(hm, oc)` for `=`/`<>`, and `eql_v3.ore_cllw` reads `oc` for
+/// `<`/`<=`/`>`/`>=`. The `json` container and `jsonb_query` domains carry no
+/// term extractors (their surface is containment `@>`/`<@` and path navigation),
+/// so they return no terms. Keyed on the catalog domain name (`json`/`entry`/
+/// `query`).
+fn stevec_terms(name: &str) -> Vec<TermInfo> {
+    if name != "entry" {
+        return Vec::new();
+    }
     vec![
         TermInfo {
             key: "hm",
@@ -145,8 +156,9 @@ pub fn dump_catalog() -> CatalogDump {
         .map(|d| SteVecEntry {
             full_name: d.full_name(eql_domains::JSONB.name),
             name: d.name,
-            // Catalog terms are empty for SteVec (see stevec_terms); hardcode.
-            terms: stevec_terms(),
+            // Catalog terms are empty for SteVec; hardcode per-domain — only
+            // `jsonb_entry` carries extractors (see stevec_terms).
+            terms: stevec_terms(d.name),
         })
         .collect();
 
@@ -210,10 +222,26 @@ mod tests {
         let names: Vec<&str> = dump.stevec.iter().map(|e| e.full_name.as_str()).collect();
         assert_eq!(names, ["json", "jsonb_entry", "jsonb_query"]);
 
-        // The (hardcoded) SteVec extractors are surfaced, not left empty.
-        let entry = &dump.stevec[0];
-        let extractors: Vec<&str> = entry.terms.iter().map(|t| t.extractor).collect();
-        assert_eq!(extractors, ["eq_term", "ore_cllw"]);
+        let by_name = |n: &str| {
+            dump.stevec
+                .iter()
+                .find(|e| e.full_name == n)
+                .unwrap_or_else(|| panic!("{n} present"))
+        };
+
+        // Term extractors live on `jsonb_entry` (the sv element type) ONLY:
+        // `eq_term` (hm/oc equality) + `ore_cllw` (oc ordering).
+        let entry_extractors: Vec<&str> = by_name("jsonb_entry")
+            .terms
+            .iter()
+            .map(|t| t.extractor)
+            .collect();
+        assert_eq!(entry_extractors, ["eq_term", "ore_cllw"]);
+
+        // The `json` container and `jsonb_query` domains carry no term
+        // extractors — their surface is containment (@>, <@) and path nav.
+        assert!(by_name("json").terms.is_empty());
+        assert!(by_name("jsonb_query").terms.is_empty());
     }
 
     /// Pins the hand-re-derived `suffix` wire field — the one channel with no
--- a/crates/eql-codegen/src/dump.rs
+++ b/crates/eql-codegen/src/dump.rs
@@ -93,8 +93,16 @@ fn term_infos(terms: &[Term]) -> Vec<TermInfo> {
 /// (hash-equality, `eql_v3.eq_term`) or `oc` (CLLW-ORE ordering, `eql_v3.ore_cllw`).
 fn stevec_terms() -> Vec<TermInfo> {
     vec![
-        TermInfo { key: "hm", extractor: "eq_term", ctor: "hmac_256" },
-        TermInfo { key: "oc", extractor: "ore_cllw", ctor: "ore_cllw" },
+        TermInfo {
+            key: "hm",
+            extractor: "eq_term",
+            ctor: "hmac_256",
+        },
+        TermInfo {
+            key: "oc",
+            extractor: "ore_cllw",
+            ctor: "ore_cllw",
+        },
     ]
 }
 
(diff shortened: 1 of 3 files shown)


crates/eql-codegen/src/dump.rs
```diff
@@ -8,10 +8,15 @@
 use eql_domains::Term;
 use serde::Serialize;
 
-/// The catalog surface: every scalar type and its domains.
+/// The catalog surface: every scalar type and its domains, plus the non-scalar
+/// SteVec (`jsonb`) family.
 #[derive(Serialize)]
 pub struct CatalogDump {
     pub types: Vec<TypeEntry>,
+    /// The `jsonb` (SteVec) family — `eql_v3.json` / `jsonb_entry` / `jsonb_query`.
```
tobyhede: **Doc references the wrong schema.** These domains are `public`-schema (`public.json` / `public.jsonb_entry` / `public.jsonb_query` — see `eql-domains/src/lib.rs:436-437`, and `xml-to-json.py:122` emits `public.{full_name}`), not `eql_v3`. The extractor _functions_ live in `eql_v3`, but the _domains_ named here live in `public`. Suggest `public.json / …` to avoid a future consumer qualifying as `eql_v3.json` and minting wrong manifest identifiers.

crates/eql-codegen/src/dump.rs
```diff
@@ -36,6 +41,44 @@ pub struct DomainEntry {
...
+    /// SEM index-term constructor (unqualified): `hmac_256` / `ore_block_256` /
+    /// `bloom_filter` / `ope_cllw`.
+    pub ctor: &'static str,
+}
+
+/// One `jsonb` (SteVec) domain: catalog inventory only — its SQL surface
+/// (CHECK, operators) is hand-written and not derivable from the catalog.
+#[derive(Serialize)]
+pub struct SteVecEntry {
+    /// The `eql_v3`-relative domain name: `json` / `jsonb_entry` / `jsonb_query`.
```
tobyhede: Same schema slip: `full_name` returns a bare, schema-agnostic name (identical in kind to a scalar's `integer_eq`, which resolves under `public`). "`eql_v3`-relative" is inaccurate — the domain resolves under `public`. Consider "bare domain name (resolved under `public`)".

crates/eql-codegen/src/dump.rs
```diff
@@ -36,6 +41,44 @@ pub struct DomainEntry {
...
+    pub full_name: String,
+    /// The catalog domain name: `json` / `entry` / `query`.
+    pub name: &'static str,
+    pub terms: Vec<TermInfo>,
```
tobyhede, line 74: **`terms` here is provably always empty for jsonb.** All three `JSONB_DOMAINS` declare `terms: &[]`, `jsonb_family_is_non_scalar_and_in_catalog_after_flip` asserts `.all(|d| d.terms.is_empty())`, and the `shape_and_terms_are_consistent` coupling invariant fails CI if a non-`Scalar` domain ever gains a term. So `term_infos(d.terms)` at line 123 → `[]` always → the manifest shows `public.json`/`jsonb_entry`/`jsonb_query` with empty `termFunctions` and `supportedOperators` (a fully-searchable SteVec family rendered inert). Cuts against #366's goal of linking domains to their extractors — worth either dropping this always-empty field or sourcing jsonb's real (hand-written) extractors into the manifest.

tasks/docs/generate/xml-to-json.py
```diff
@@ -105,7 +112,23 @@ def load_domains(catalog_path: Path) -> list:
...
+    for entry in catalog.get("stevec", []):
+        domains.append({
+            "name": f"public.{entry['full_name']}",
+            "type": "jsonb",
+            "variant": "",
+            "base": "jsonb",
+            "shape": "stevec",
+            "capabilities": ["json"],
+            "supportedOperators": [],
+            "termFunctions": _term_functions(entry.get("terms", [])),
```
tobyhede, line 129: `entry.get("terms", [])` is always `[]` here (jsonb domains carry empty catalog `terms` by the `shape_and_terms_are_consistent` invariant), so `termFunctions` is always `[]` — combined with the hardcoded `supportedOperators: []` above, the manifest presents the SteVec family as having no operators/extractors. See the paired comment on `dump.rs:70`.

tobyhede (commented):
Two low-severity issues from review — both confirmed against the code, neither a correctness bug.

crates/eql-codegen/src/dump.rs
coderdan: Fixed in `ab7f84e2`: the doc now reads `public.json / public.jsonb_entry / public.jsonb_query`. The extractor functions stay `eql_v3.*` — only the domain names were wrong.

crates/eql-codegen/src/dump.rs
coderdan: Fixed in `ab7f84e2`: now reads "The bare domain name (resolved under the `public` schema, like a scalar's `integer_eq`)" — dropped the inaccurate `eql_v3`-relative wording.

crates/eql-codegen/src/dump.rs
coderdan: Addressed in `ab7f84e2` per your Slack suggestion: added `stevec_terms()`, which hardcodes the hand-written SteVec extractors (`hm → eq_term`, `oc → ore_cllw`) instead of reading the provably-empty `d.terms`. The manifest now surfaces `eql_v3.eq_term` / `eql_v3.ore_cllw` for the jsonb family, and the doc comment explains the hardcode + points at the `shape_and_terms_are_consistent` invariant. `supportedOperators` is still `[]` — that one genuinely needs `pg_operator` introspection (the jsonb operators aren't in the catalog at all), which I've flagged as a follow-up rather than fake it here.

tasks/docs/generate/xml-to-json.py
coderdan: Same fix as the paired `dump.rs` comment (`ab7f84e2`): `entry.get("terms")` now returns the hardcoded `hm`/`oc` terms, so `termFunctions` resolves to `["eql_v3.eq_term", "eql_v3.ore_cllw"]`. `test_load_domains` asserts exactly that. `supportedOperators` stays `[]` pending the `pg_operator` follow-up.

tasks/docs/generate/test_xml_to_json.py
```diff
@@ -24,11 +24,26 @@
...
+        "terms": [
+          {"key": "hm", "extractor": "eq_term", "ctor": "hmac_256"},
+          {"key": "bf", "extractor": "match_term", "ctor": "bloom_filter"}
+        ] }
     ]}
+  ],
+  "stevec": [
+    { "full_name": "json", "name": "json", "terms": [
+      {"key": "hm", "extractor": "eq_term", "ctor": "hmac_256"},
+      {"key": "oc", "extractor": "ore_cllw", "ctor": "ore_cllw"}
```
coderdan: These are not valid terms for `stevec`. They could be an array of `jsonb_entry` but I think officially for now the SQL functions return `[]`. Please validate and correct.

coderdan: Good catch — validated against `src/v3/jsonb/{functions,operators}.sql` and you're right: hm/oc are **not** container-level terms. Fixed in `e00880a0`.

The extractors are typed on `public.jsonb_entry` (the sv *element* type), never on the `json` container or `jsonb_query`:
- `eql_v3.eq_term(entry public.jsonb_entry)` → `coalesce(hm, oc)`, backs `=` / `<>`
- `eql_v3.ore_cllw(entry public.jsonb_entry)` → `oc`, backs `<` `<=` `>` `>=`
- doc comment on `ore_cllw` even says *"`oc` is only ever present on an sv element, never at a root encrypted value."*

The `json` container and `jsonb_query` domains have no term extractors at all — their surface is containment (`@>`, `<@`) + path navigation. So `stevec_terms()` now keys on the catalog domain name and returns the hm/oc pair for `entry` only; `json`/`query` resolve to `[]`. Verified end-to-end:

```
public.json         termFunctions=[]
public.jsonb_entry  termFunctions=['eql_v3.eq_term', 'eql_v3.ore_cllw']
public.jsonb_query  termFunctions=[]
```

Two notes:
1. `supportedOperators` is still `[]` across all three (the `pg_operator` introspection follow-up) — the operators exist in the SQL but aren't in the catalog, so I've left them empty here rather than hand-maintain a second list.
2. If "officially for now the SQL functions return `[]`" meant you'd rather ship **all three empty** until `jsonb_entry`'s terms are catalog-backed, say the word and I'll collapse it — but since `eq_term`/`ore_cllw` are live (wired into 6 operators on `jsonb_entry`), surfacing them there is the SQL-accurate representation.

PR #373, opened 2026-07-07
feat(eql v3): scalar query-operand surface — domains, operators, bindings (CIP-3432)

freshtonic:
Implements [CIP-3432](https://linear.app/cipherstash/issue/CIP-3432): the complete EQL v3 **scalar query-operand surface** — a public, supported way to query encrypted columns with an **index-terms-only** operand (no decryptable ciphertext in the query).

**Security driver:** query operands travel through PostgREST query strings, URL logs, proxies, and SQL logs. They should carry index terms only, not a full storage envelope with a real `c` ciphertext.

## Design (locked with @freshtonic)

- **Per-capability query domains** (Option B): each term-bearing storage domain gets a paired `public.<name>_query` twin — type-level safety + a 1:1 codegen mirror.
- **Enveloped, term-only operand** `{v, i, <terms>}` — the storage struct minus the ciphertext `c` (no `k`; the `<name>_query` domain + absent `c` disambiguate). The v2 "hoist": the terms already exist in the v2 payload.

## What's here

**Client bindings** (`crates/eql-bindings`, `crates/eql-codegen`):
- Generated `<Name>Query` twins (storage struct − `c`) + catalog-generated `QueryPayload` (a variant per twin + the SteVec needle) + a separate `all_query()` inventory. **38 TypeScript + 38 JSON Schema** query artifacts.
- `from_v2_query` / `from_v2_query_typed` hoist term-bearing scalar targets into `{v:3, i, <terms>}`; storage-only stays `UnsupportedQueryTarget`.

**SQL surface** (generated from the `eql-domains` catalog under `src/v3/scalars/`):
- `public.<name>_query` domains — CHECK asserts `v`/`i` + terms and **forbids `c`**.
- Query extractor overloads + comparison wrappers binding `(storage, query)` and the `(query, storage)` commutator, plus a `CREATE OPERATOR` for each — for supported operators only, reusing the storage templates.

**Tests** — every generated query domain is exercised against real ciphertext:
- `<T>_eq_query` / `<T>_ord_query` / `<T>_ord_ore_query` (all 9 families) — folded into the shared all-pairs `assert_eq_oracle`/`assert_ord_oracle` (drives the term-only operand both directions, in the **fixture** and **e2e fresh-encryption** suites, with no added round trips).
- `<T>_ord_ope_query` — the OPE fixture smoke runs each predicate against the query operand.
- `text_match_query` — bloom containment via a query needle; `text_search_query` — a new `DomainQuery` fn-oracle overload.
- `jsonb_query` — already covered (`v3_jsonb_tests` D4).
- Standalone **fresh-encryption conformance** (`v3_scalar_query_operand_tests`, gated `proptest-e2e`): an independently-encrypted term-only operand matches exactly the equal / correctly-ordered rows; a `c`-bearing operand is rejected by the CHECK.

## Review notes

- The SQL is machine-generated — review the generator (`generate.rs`, `context.rs`, `query_types.sql.j2`) and one concrete output (`src/v3/scalars/integer/integer_eq_query_*.sql`); the rest is 1:1.
- **Public surface is purely additive:** the golden snapshot regenerated to **+423 / −0**, all `*_query` — no existing operator changed (no ambiguity with the `(domain, jsonb)` overloads).
- One self-inflicted note: a first coverage attempt as a *separate* oracle pass doubled per-test DB round trips and made the parallel suite flaky; it's folded into the existing SELECTs now (42/42 stable).

## Validation (all green locally, PG 17)

`test:crates` (fmt + clippy `-D warnings` + tests) · `codegen:parity` · `types:check` · `docs:validate` (244 SQL files) · `clean_install_v3` · public-surface golden · fixture oracle **42/42** · standalone conformance **3/3**. The `proptest-e2e` suite runs in CI's e2e job.

## Open (planning, not implementation)

Whether CIP-3432 is the umbrella for CIP-3383 / CIP-3423 / CIP-3420 or the EQL-core half alongside the protect-ffi work. JSONB (SteVec) query terms shipped earlier; this PR is the **scalar** half.


tobyhede (commented):
Recommended actions (ranked)

  1. Fix the stale doc — real defect this commit introduced.
  crates/eql-bindings/src/v3/mod.rs:79-82 still says QueryPayload is
  "hand-written rather than generated" and points to a "why hand-written"
  rationale that no longer exists — this commit converted it to
  catalog-generated and reversed the original argument. mod.rs now
  contradicts the file it describes. Zero-risk fix.

  2. Give "term-bearing / has-query-operand" a catalog home (best structural
  win). The predicate !d.terms.is_empty() is re-inlined 3× in the generator
  (bindings.rs:295, :353, :531) plus twice in tests — a stand-in for a
  named concept, unlike its siblings scalar_families() /
  stored_payload_domains(). Add query_domains() (or
  Domain::has_query_operand()) and collapse the call sites. This is where
  the query concept should live.

  3. De-duplicate the generator (do after #2). Three near-identical pairs:
  the impl DomainType block (:169-188 vs :247-266, byte-identical bar two
  literals); the doc-line construction (render_query_struct re-implements
  struct_doc_lines inline at :210-222); and ~90 lines of enum rendering
  (render_payload_rs vs render_query_payload_rs). All are refactors of the
  generator with the existing byte-for-byte drift gates as the safety net.

  Nitpicks: derive the SteVec query row from the catalog instead of the
  hardcoded tuple at bindings.rs:543-548 (keep only the "SteVec" variant
  literal); replace the awkward 4-string tuple; add a query-twin analogue of
  envelope_fields_match_catalog_keys so an ENVELOPE_KEYS change can't
  silently desync the twin.

  Test gaps (low-risk, shared code is covered): query-path MissingTerm, bf
  wrap/out-of-range, and omitted-v leniency are only exercised through the
  stored path — a few targeted scalar-query cases would close them.

crates/eql-bindings/src/v3/numeric.rs
```diff
@@ -165,3 +165,127 @@ impl DomainType for NumericOrdOpe {
...
+///
+/// Operators: `=` `<>`. Required keys: `v` `i` `hm`.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, TS, JsonSchema)]
+#[ts(export, export_to = "v3/")]
+#[serde(deny_unknown_fields)]
+pub struct NumericEqQuery {
+    pub v: SchemaVersion,
+    pub i: Identifier,
+    pub hm: Hmac256,
+}
```
coderdan, line 178: Can be a follow-up but if we implement `PartialEq` for these types then its worth implementing `vitaminc::TimeSafeEq` (safe PartialEq replacement) for `Hmac256` and friends. This will mitigate timing side-chains on the values if used in Rust (largely moot until `stack-encrypt` is ready but flagging now).

freshtonic: Good call — filed as a follow-up: [CIP-3440](https://linear.app/cipherstash/issue/CIP-3440) (`vitaminc::TimeSafeEq` for `Hmac256` & friends). Kept out of this PR since it is largely moot until `stack-encrypt` compares these term values in Rust, but captured so the mitigation is not lost.

crates/eql-bindings/src/v3/real.rs
```diff
@@ -165,3 +165,127 @@ impl DomainType for RealOrdOpe {
...
+///
+/// Operators: `=` `<>` `<` `<=` `>` `>=`. Required keys: `v` `i` `ob`.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, TS, JsonSchema)]
+#[ts(export, export_to = "v3/")]
+#[serde(deny_unknown_fields)]
+pub struct RealOrdOreQuery {
+    pub v: SchemaVersion,
+    pub i: Identifier,
+    pub ob: OreBlock256,
+}
```
coderdan, line 209: Is there anything in the Rust layer that specifies what domain type a query type is valid for?

I half expected a `DomainTypeQuery` trait. E.g.

```rust
impl DomainTypeQuery for RealOrdOreQuery {
    type Target = RealOrdOre;

    // etc
}
```

Maybe its overkill or maybe its just a matter of time?

crates/eql-bindings/bindings/v3/BigintEqQuery.ts
```diff
@@ -0,0 +1,11 @@
...
+import type { Hmac256 } from "./Hmac256";
+import type { Identifier } from "./Identifier";
+import type { SchemaVersion } from "./SchemaVersion";
+
+/**
+ * `public.bigint_eq_query` — equality domain query operand.
+ *
+ * Operators: `=` `<>`. Required keys: `v` `i` `hm`.
+ */
+export type BigintEqQuery = { v: SchemaVersion, i: Identifier, hm: Hmac256, };
```
coderdan, line 11: I really like this but one possible footgun is that query values could be stored in the table. The check doesn't discriminate. We might want to consider a way to add checks on the column itself (superset of the domain) as a follow-up.

```sql
CREATE DOMAIN public.numeric AS jsonb
  CHECK (
    jsonb_typeof(VALUE) = 'object'
    AND VALUE ? 'v'
    AND VALUE ? 'i'
    AND VALUE ? 'c'
    AND VALUE->>'v' = '3'
  );
```

crates/eql-bindings/src/v3/mod.rs
```diff
@@ -151,6 +151,6 @@ pub mod text;
 pub mod timestamp;
 
 pub use domain_type::{DomainType, SCHEMA_ID_BASE, SQL_SCHEMA};
-pub use inventory::all;
+pub use inventory::{all, all_query};
```
coderdan, line 154: There doesn't seem to be any query types for `jsonb` in either Rust or TS. Is that intentional (i.e. a follow-up) or was it missed?

tests/sqlx/src/property.rs
```diff
@@ -173,23 +250,31 @@ pub enum Overload {
     DomainDomain,
     DomainJsonb,
     JsonbDomain,
+    /// CIP-3432: the RHS is the term-only query operand (`<domain>_query`, the
```
coderdan, line 253: Prefer not to mention Linear issues in public changes.

coderdan (approved):
Epic! Some non-blocking feedback. I assume the JSON stuff will be a follow-up (i.e. by "scalar" we mean non-JSON)

PR #11, opened 2024-10-11
Go xorm example

fimac:

PR #374, opened 2026-07-07
feat(release): publish EQL v3 language bindings (Rust + TypeScript) with bundled SQL

tobyhede:
## Summary

Closes #331.

Publishes EQL v3 **language bindings as first-class release artifacts**, and consolidates all releasing onto a **Changesets-owned flow with a single workflow entrypoint** (`release.yml`):

- **Rust** — `eql-bindings` crate (crates.io via release-plz, now **publish-only**: versioning is owned by Changesets)
- **TypeScript** — new `@cipherstash/eql` npm package (`packages/eql/`), generated from the crate's `bindings/` + `schema/` outputs
- **Bundled SQL** — each language package ships the **exact** self-contained `eql_v3` installer/uninstaller it was generated against (crate: `eql_bindings::sql`; npm: `./sql` subpath exports), with a sha256 release manifest

The old CI-native alpha coordinator is **removed** (`release-alpha.yml`, `release-eql.yml`, `.github/scripts/*` + their shell test suites, and the `mise run release:*` dispatch wrappers). In its place:

| Path | Trigger | What happens |
|---|---|---|
| **Production** | push to `main` | `changesets/action` maintains a "Version Packages" PR (`changeset version` + `sync-lockstep-versions.mjs` — bumps `package.json`, `Cargo.toml`, and rebuilds/bundles exact-version SQL). Merging it publishes npm, and the push triggers `release-plz.yml` for the crate; `release.yml` then attaches the `eql-V` SQL + docs release and dispatches the Docker image build. |
| **Prerelease** | `chore(release): …` commit on `eql_v3` (Changesets pre-mode, version already pinned) | `release.yml` builds/attaches prerelease SQL + docs, publishes npm (dist-tag = channel), tags `eql-typescript-vV`, and dispatches `release-plz.yml` against the `eql-V` tag for the crate. |

`@cipherstash/eql`'s `package.json` version is the **single source of truth** for the lockstep identity `V`. Full reference: `docs/development/releasing.md`.

## What's in here

- **New TS package** `@cipherstash/eql`: type barrel (`export type * from generated/v3`), `schema` metadata + `schemaId()`, `sql` runtime helpers (`readInstallSql`, `releaseManifest`). pnpm workspace, tsup (ESM+CJS), vitest.
- **Generation pipeline**: `packages/eql/scripts/sync-generated.mjs` copies TS/JSON from `crates/eql-bindings` (node builtins only — no pnpm needed at gate time); `mise run typescript:generate` / `typescript:check` drift-gate it; `types:check` extended to cover the package output; `node = "22"` added to mise tools.
- **Exact-version SQL bundling**: `tasks/release/prepare-bindings-assets.sh` builds `--version <V>` SQL and copies it (+ sha256 manifest) into both packages. DEV placeholders committed so clean clones build.
- **Changesets**: `.changeset/` config + pre-mode (`alpha`), backfilled changesets for the pre-existing `[Unreleased]` entries (through #379). The release changelog is `packages/eql/CHANGELOG.md` (per-package, lockstep = whole release); the root `CHANGELOG.md` is the frozen pre-3.0 archive.
- **Publish safety rails**:
  - `prepublishOnly` gate (`verify-release-assets.mjs`): npm refuses to publish when the bundled SQL manifest doesn't match the package version (covers both `changeset publish` and the prerelease script).
  - `release-plz.yml` refuses to publish a crate whose committed `sql/` bundle wasn't prepared for the crate version (enforced only when the version isn't already on crates.io).
  - `classify` requires exactly `chore(release):`, rejects non-prerelease-shaped versions, and skips identities whose `eql-typescript-vV` tag already exists (idempotent re-runs).
  - Crate publish dispatched against the `eql-V` **tag**, so all artifacts ship from one commit.
  - npm pinned to an exact version in publish jobs; **no caches** in publish workflows (`lint-no-workflow-caching.mjs` covers `pnpm/action-setup`, `setup-node`, `actions/cache`, **and `jdx/mise-action`**, across `release.yml` + `release-plz.yml`).
- **Docs**: `docs/development/releasing.md` (new), `CLAUDE.md`, root `README.md`, `crates/eql-bindings/README.md` (also fixes the stale "envelope is `v: 2`" claim), `.github/workflows/README.md`.

## Testing

- Package surface: vitest (`packages/eql/src/*.test.ts` — 2 files, 5 tests).
- Release tooling: vitest for the cache lint and the lockstep Cargo bump (`scripts/*.test.mjs` — 8 tests), plus `tasks/release/prepare-bindings-assets.test.sh` (version-validation gate; rejects malformed/injection-shaped identities). All wired into `lint-release.yml` alongside actionlint + shellcheck.
- Drift gates: `mise run types:check` / `typescript:check` (regenerate + git diff, pnpm-free).

## Notes for reviewers

- The prerelease access model is push-to-`eql_v3` + the marker convention; branch protection is the control (documented in `releasing.md`).
- Actions are tag-pinned, not SHA-pinned, matching current repo convention — flagged in review as a candidate follow-up for the two publish workflows.
- The live release flow (Version PR → publish) can only be exercised for real on `main`; the prerelease path can be dry-run on a scratch branch containing a marker commit (`gh workflow run release.yml --ref <scratch-branch>`).


freshtonic (changes requested):
*Scope: non-generated changes only (~80 files). Generated TS/JSON snapshots under `packages/eql/src/generated/**` and the `crates/eql-bindings/bindings|schema` copies were skipped.*

## Overview

Replacing the bespoke alpha coordinator (`release-alpha.yml` + `.github/scripts/*` + `mise run release:*` wrappers) with a **changesets-owned release** — one `release.yml` entrypoint with a production path (`main`, Version-Packages PR) and a prerelease path (`eql_v3`, `chore(release):` marker commit) — is a genuine simplification, `docs/development/releasing.md` is excellent, and the lockstep-version design (`package.json` as the single source of truth for `V`) is coherent. The findings below are mostly about the seams.

## Process findings (read these first)

1. **The PR description describes a previous revision of this PR.** The body says the alpha coordinator was *extended* with `rust`/`typescript` targets, names a `release-typescript.yml` workflow, and claims "extended shell unit tests: derive-identity, release-alpha-resolve, release-alpha-pin-bindings" — none of which exist in the diff; those scripts and their four test suites are **deleted**. A reviewer trusting the body would approve claims the diff doesn't contain. Please rewrite the body for the changesets design.
2. **CI has not run on the current head** (only CodeRabbit as of this review; last force-push 2026-07-08 04:29Z). All "verified locally" claims are unvalidated in CI, and one of them provably breaks there (finding 4).
3. **The PR is CONFLICTING with `eql_v3`** — #375, #377 and especially #379 (query-operand domains renamed to `query_<name>` and moved into the `eql_v3` schema) landed after this branch. The rebase is not mechanical: the generated `bindings/`/`schema/` snapshots must be re-synced (`jsonb_query.json` → `query_jsonb.json` etc.), and see finding 5 about the deleted `[Unreleased]` section.

## Correctness

4. **`mise run types:check` now shells out to pnpm, but nothing provisions pnpm.** `types:check` gained `pnpm --filter @cipherstash/eql sync:generated`, yet `node`/`pnpm` are not in mise `[tools]`, and `test-eql.yml`'s "Rust workspace crates" job (which runs `mise run types:check` at ~line 345) sets up neither pnpm nor `node_modules`. That job will fail with `pnpm: command not found` on every PR once this merges. Same problem for the pre-commit hook and any contributor without pnpm. Fix: add `node`/`pnpm` to mise `[tools]` + a `pnpm install --frozen-lockfile` step in that CI job — or keep `types:check` pnpm-free and let `typescript:check` (already a separate task) own the npm-package gate.
5. **The root `CHANGELOG.md` `[Unreleased]` section is deleted, but changesets will not maintain the root changelog.** Changesets writes *per-package* changelogs (`packages/eql/CHANGELOG.md`); as configured, nothing ever writes the root `CHANGELOG.md` again, yet CLAUDE.md now says "`CHANGELOG.md` is generated from the `.changeset/*.md` files". Either add tooling that assembles the root file, or document that the canonical changelog moves to `packages/eql/CHANGELOG.md` and leave a pointer in the root file. Related rebase hazard: the backfilled `.changeset/eql-*.md` files snapshot `[Unreleased]` as of an older base — entries added since (#373's query-operand surface, #375, #377, #379) have no changeset and will be **silently dropped from release notes** when the conflicting `[Unreleased]` deletion is resolved. Please backfill those four during the rebase.
6. **Nothing guards the crate against publishing the DEV placeholder SQL.** `eql_bindings::sql` uses `include_str!` on committed files; release-plz publishes the committed tree verbatim. The npm prerelease job rebuilds exact SQL in-workflow, but the crate path trusts that the operator's release commit ran the full `pnpm run version` (prepare-bindings-assets). If someone hand-pins `package.json`/`Cargo.toml` — or prep fails partway — `eql-bindings@V` ships `-- DEV placeholder` as its "exact SQL". Cheap fix: in `release-plz.yml` (now publish-only), assert `crates/eql-bindings/sql/release-manifest.json .eqlVersion == Cargo.toml version` before running release-plz; similarly, `npm-publish.mjs` could refuse when `releaseManifest.eqlVersion !== pkg.version` — closing the production npm path too.
7. **The prerelease marker is loose and re-triggerable.** `classify` matches any `eql_v3` commit subject starting `release:` or `chore(release):`, then only checks that the *committed* version is prerelease-shaped — which in pre-mode it always is. A doc commit titled "release: fix runbook" re-runs the entire publish pipeline for the currently pinned identity. The npm-publish and tag steps are idempotent (nice), but `_build-sql.yml`'s `gh release create` path isn't obviously so. Consider requiring exactly `chore(release):`, or short-circuiting in `classify` when `eql-typescript-v$VERSION` already exists.
8. **`prerelease-publish-rust` dispatches release-plz against the branch, not the release commit** (`gh workflow run release-plz.yml --ref "$BRANCH"`). If `eql_v3` advances between the marker push and the dispatch, the crate publishes from a *different commit* than the SQL/npm artifacts — breaking the "one identity, one commit" invariant the old coordinator explicitly enforced. The sibling `build-image` job already solves this by dispatching against the `eql-V` tag; do the same here (the tag exists — `prerelease-build-sql` is a `needs`).

## Security

9. **`npm install -g npm@^11.5.1` in publish jobs is an unpinned registry install at publish time.** This PR ships a whole lint (`lint-no-workflow-caching.mjs`) to keep caches out of the publish path, then pulls whatever the latest npm 11.x is, with `id-token: write` in scope. Pin an exact version.
10. **`jdx/mise-action@v3` runs with `cache: true` in `release.yml` and `release-plz.yml`, and the new cache lint doesn't cover it.** The poisoned-cache rationale that bans `setup-node`/`pnpm` caching applies equally to mise's restored toolchain. Either extend the lint and set `cache: false` in publish workflows, or record why mise's cache is out of scope.
11. Actions are tag-pinned (`changesets/action@v1.8.0`, `pnpm/action-setup@v6.0.8`, …), not SHA-pinned, in workflows holding npm/crates OIDC publishing power. Repo-wide convention today, so a suggestion rather than a blocker — but these two workflows are where SHA-pinning pays.
12. Noting a posture change, deliberately accepted or not: the deleted `release-alpha-verify-commit-signature.sh` gate is not replaced — a prerelease publish is now gated solely by push access to `eql_v3` plus the marker convention. Fine if branch protection is the intended control; worth one line in `releasing.md`.

## Test coverage

- The four deleted shell test suites guarded logic that still exists in new form (identity/classification, pinning). Their replacements: vitest for the cache lint only. `classify`'s branching, `prepare-bindings-assets.sh` (version regex, manifest emission), and `sync-lockstep-versions.mjs` (the Cargo.toml first-match regex — correct today but assumption-laden) have **no tests**. The `scripts/vitest.config.mjs` harness is already there; these are cheap to add.
- The npm package tests are sensible; `sql.test.ts`'s three-state design (missing/DEV/real) is honest about what it can assert.

## Nits

- `release.yml`/`release-plz.yml` use `defaults.run.shell: bash {0}`, which *disables* GitHub's default `-eo pipefail`. Every current step compensates with `set -euo pipefail`, but the default is a trap for the next step added. Prefer the implicit `bash` default.
- `copy-assets.mjs`: unused `repoRoot`.
- `sync-generated.mjs --check` mutates the tree before comparing (regenerate-then-diff is the repo convention, so acceptable — but unlike the mise gates it "fixes" the tree while failing, which can confuse a local run).
- `crates/eql-bindings/README.md` fix of the "`v: 2`" error is appreciated (the body flagged it as out of scope, but it's in the diff — another body/diff mismatch, this one in the diff's favour).

## Verdict

The direction is right and the docs are strong, but this needs a revision before merge: **4** (CI breakage for every future PR), **5** (changelog ownership gap + entries silently dropped on rebase), and **6** (DEV-placeholder crate publish) are blocking in my view; **1–3** are process must-dos (accurate body, green CI on the actual head, rebase over #375/#377/#379); the rest are hardening/tests that could land as a fast-follow if the release is urgent.

freshtonic (approved):
Re-reviewed at `792c78d8` — all actionable findings from my earlier review are addressed:

- **#4** fixed in `25a2001b` (gates are pnpm-free; node provisioned via mise).
- **#3/#5a** resolved by the rebase + changeset backfill (`da4039d3`, `b8d796bf`).
- **#1, #5b, #6–#10, #12, tests, nits** addressed in `792c78d8` (see the findings-addressed comment for the mapping).
- **#11** (SHA-pinning the publish workflows' actions) deferred as a follow-up, noted in the PR body.

CI is green on this head (Test EQL + Lint release tooling, the latter now exercising the new shell test and the extended cache lint).

Caveat for the record: `25a2001b`/`792c78d8` are my own commits, so this approval covers the original PR content plus my review; Toby should sanity-check those two commits before merge — in particular the release-plz DEV-placeholder guard's crates.io no-op check and the classify tag-exists short-circuit, since both change publish-path behaviour.

PR #379, opened 2026-07-07
feat(eql v3)!: query-operand domains — query_<name> prefix, homed in the eql_v3 schema (CIP-3442)

freshtonic:
Implements [CIP-3442](https://linear.app/cipherstash/issue/CIP-3442/rename-scalar-query-operand-domains-from-name-query-to-query-name): the query-operand domains change identity in two coordinated ways relative to the 3.0.0 pre-releases — renamed from the `<name>_query` suffix to a `query_<name>` prefix, and moved from `public` into the `eql_v3` schema.

```sql
WHERE col = $1::eql_v3.query_integer_eq;
WHERE doc @> $1::eql_v3.query_jsonb;   -- was public.jsonb_query
```

**Why the prefix:** alphabetical type listings (most visibly Supabase Studio's Table Builder type picker) interleaved never-a-column-type query operands with the actual column types (`integer_eq` next to `integer_eq_query`); the shared `query_` prefix sorts every query operand together.

**Why the schema move:** a query operand is never a valid column type, so it doesn't belong in the column-type namespace. The survive-schema-drop rationale for `public` placement (dropping EQL-owned schemas must never drop application columns) doesn't apply to a type no application column should use. In `eql_v3` the operands are versioned and uninstalled with the rest of the public API surface, and casting requires the same `USAGE ON SCHEMA eql_v3` a querying role already needs for the extractors and comparison wrappers.

## What's here (3 commits)

1. **Rename** (`24bd6f40`): `Domain::query_name` in `crates/eql-domains` is the single source of truth (`query_<full_name>`); all codegen — SQL renderers, generated file names, bindings emitter, `QueryPayload` — derives from it. The jsonb needle follows as a second documented `full_name` exception beside `public.json`.
2. **Non-vacuous operator-equivalents gate** (`c6f5e01d`): test #1 filtered operand types on `nspname = 'eql_v3'` where (then) no types lived, so it matched zero operators and passed vacuously. It now identifies EQL operands by catalog domain name in their home namespaces and asserts a ≥100-operator match floor.
3. **Schema move** (`39aee45f`): `query_domain_name` in the codegen qualifies twins with `SCHEMA`; `DomainType::sql_domain` is `eql_v3.query_<name>` and `DomainType::domain` strips whichever schema qualifies it. **Uninstall semantics pinned**: a column misusing a query-operand domain is dropped with the schema (CASCADE) — the uninstall suite asserts the misuse table survives minus that column, while column-domain tables survive intact. A new public-surface pin (`query_operand_domains_are_eql_v3_jsonb_domains`) mirrors the user-column placement pins.

Docs: CHANGELOG entries (EQL + eql-bindings) and upgrade note **U-002** in `docs/upgrading/v3.0.md` — the old names shipped only in 3.0.0 pre-releases, so protect-ffi and other alpha adopters need the coordinated cast update. `permissions.md` gains the query-operand cast row (needs `eql_v3` USAGE only).

## Reviewing

The SQL is machine-generated — review the naming/qualification sources (`Domain::query_name`, `full_name`'s needle exception, `query_domain_name` + the `query_types` template) and one concrete output (`src/v3/scalars/integer/query_integer_*.sql`); the rest is 1:1. The public-surface golden regenerated as a clean ±426 both times.

## Validation (local, PG 17)

`codegen:parity` · `types:check` · `test:matrix:inventory` · `docs:validate` · `test:self_contained_v3` · `test:splinter` (0 unmatched) · `clean_install_v3` · full SQLx suite including `proptest-e2e` fresh-encryption query-operand conformance and the reworked uninstall suite.

Known-unrelated local failures: 44 tests (jsonb_entry / v3_jsonb / float_special) fail locally because regenerated SteVec fixtures no longer match the pinned selector constants — reproduces on the clean base tree; environmental (local workspace vs the workspace CI's creds point at), tracked separately. CI is green on these suites.


tobyhede (approved):
Suggest clearly separating query types by moving to the `eql_v3` schema. 
Column types need to be public so that the `DROP/CREATE schema` upgrade path does not drop columns and data. 

Additional items to consider as follow-ups

**Risk: customer uses query type as column type.**

1. Add an eql_v3.lints() check
Add a lint that scans pg_attribute for columns whose type is `query_%` 
It would not prevent CREATE TABLE, but it would help make misuse visible in CI/admin checks.

2. Use an event trigger
A ddl_command_end event trigger could reject CREATE TABLE / ALTER TABLE ADD COLUMN when a column uses a query domain. I've validated that this should work on supabase. See: https://linear.app/cipherstash/issue/CIP-3452/prevent-query-operand-domains-from-being-used-as-column-types

coderdan (approved):
Code looks fine but I agree with @tobyhede about moving the query types to the `eql_v3` schema.

PR #382, opened 2026-07-08
build: order the whole eql_v3 surface from one walk in the codegen; retire tsort

tobyhede:
## Build-ordering refactor

Moves SQL dependency-ordering for the whole `src/v3` surface out of the bash `comment-parse → tsort → tac → cat` pipeline and into the Rust codegen. `eql-codegen order` walks `src/v3` **once**, parses the `-- REQUIRE:` edges every file declares (hand-written and generated alike), and topologically sorts them; `tasks/build.sh` concatenates in that order. `tsort` is gone from the build entirely.

**This is a pure reordering + verification change: no SQL body is modified.**

### Why one walk

The intermediate design here ordered the surface as two blocks — hand-written files (shell-globbed, minus the `-- AUTOMATICALLY GENERATED FILE.` marker) plus a codegen-emitted manifest of `render_type` output — with `tsort` demoted to a verifier over the union. Two enumerations means two predicates, and a file matching **neither** (a generated file rendered outside `render_type`, say) is silently dropped from the installer while every check stays green.

Ordering exactly the set you walk makes that unrepresentable, and it removes the marker classifier, the codegen manifest, and the shell `tsort` wrapper along with it. `install_order_contains_every_v3_sql_file` (parity tests) pins the order against an independent walk — set equality, not a count.

### What changed

- **`crates/eql-codegen/src/ordering.rs` (new).** `walk_v3_surface` + `surface_order`: Kahn with a name-keyed min-heap ⇒ byte-reproducible order. Dangling `-- REQUIRE:` targets (`UnknownTargets`), edges leaving `src/v3` (`OutsideSurface`), self-requires (`SelfEdge`), and cycles (`Cycle`) are hard errors that name every offender. These subsume the old `verify_deps_exist` and `verify_v3_self_contained` shell gates — the invariants now travel with the sort instead of with whoever remembers to call the checker. 18 unit tests.
- **`eql-codegen order` subcommand.** Prints the install order, one repo-relative path per line. Stdout contract pinned in `tests/cli.rs`.
- **`tasks/build.sh`.** Concatenates from `eql-codegen order`; no dep-list construction, no `tsort`, no `| tac` (absent on stock macOS anyway). The order is written via a temp file so an aborted run leaves no truncated list for downstream tasks to read.
- **`tasks/build/ordering.sh` (new).** Only `strip_require_lines` — an anchored `-- REQUIRE:` strip replacing the unanchored `grep -v REQUIRE`, which would eat any body line containing the substring. Propagates real grep failures (exit ≥ 2) rather than `|| true`-ing them into a truncated monolith.
- **`verify_symbol_order_v3.sh` (new).** DB-free pre-flight: an `eql_v3.` / `eql_v3_internal.` / `public.<domain>` reference must be defined by a file ordered earlier. Comment-stripped; recognises the `CREATE DOMAIN` / `CREATE TYPE` / `CREATE OPERATOR CLASS|FAMILY` definition forms. Deliberately stricter than Postgres for `LANGUAGE plpgsql` forward references — the allowlist is the documented escape hatch, and is currently **empty** (no false positives on the real surface).

  **Scope, stated honestly:** it is sound for **singleton** symbols (`hmac_256`, the ~39 `eql_v3.query_*` domains, the opclasses, `version()`) and for plpgsql forward references. It does **not** resolve overloads — a call site is a bare `eql_v3.eq(a, b)` carrying no types, and `CREATE OPERATOR` supplies `LEFTARG`/`RIGHTARG` on other lines, so a line-oriented scan cannot pick the right one. `eql_v3.eq` alone has 186 definitions spanning files #55–#242. Rather than imply otherwise, the gate reports the count it cannot resolve (41 names today) and names the authority. It still rejects a reference that precedes *every* overload, and still checks overloads whose definitions all precede the use.

- **Overload define-before-use is proven exactly by `test:clean_install_v3`** — Postgres resolves at `CREATE` time when the monolith installs, across PG 14–17, on every relevant PR, with no CipherStash credentials and no fork gating. Verified both ways: swapping `text_eq_operators.sql` ahead of `text_eq_functions.sql` in the real order passes the DB-free gate and fails the install with `function eql_v3.eq(text_eq, text_eq) does not exist` (psql exit 3).
- **`verify_installer_complete.sh` (new).** Closes the layer below the order: that the concat loop actually emitted each ordered file's body. 93 of ~244 v3 files are leaves, so a dropped one still installs cleanly and still passes the symbol checker — this gate does arithmetic on lines, which no leaf can hide from.
- **CI + docs.** `test:symbol_order_v3`, `test:installer_complete`, `test:build_ordering_helpers` wired into `test-eql.yml`; `CLAUDE.md`, `DEVELOPMENT.md`, and the scalar-adding reference updated to describe codegen-emitted ordering.

### Verification

- `cargo test -p eql-codegen`, `mise run clean && mise run build`, `codegen:parity`, `types:check`, `test:self_contained_v3`, `test:symbol_order_v3`, `test:installer_complete`, `test:build_ordering_helpers` — all green.
- Cycle and dangling-`REQUIRE:` gates driven end-to-end (injected cycle / typo'd target fails the build).
- Reorder-only was checked during development with a scratch script that `LC_ALL=C`-sorted the pre-refactor baseline monolith against the current one and required byte-identity — it passed. That script was a one-shot and is not committed (`53db3eab`), so treat the claim as a development observation, not a standing gate. The standing guarantees are `install_order_contains_every_v3_sql_file` (order ⊇ disk) and `verify_installer_complete.sh` (installer == Σ ordered bodies + pin script).
- CodeRabbit threads addressed (`e879fa70`): build-cache inputs widened to the Cargo manifests + lockfile (a dep bump changes rendered output with no `.rs` touched); self-`-- REQUIRE:` now rejected via a dedicated `SelfEdge` error rather than misreported as a cycle — its tolerance was justified by the old `tsort` self-edge-per-file workaround, which died with the shell build; symbol-gate overload blindness made visible rather than silent.
- CodeRabbit clean. Follow-up review fixed: `#MISE sources` now lists the build's shell inputs (an edit to `ordering.sh` or either gate previously left mise serving a cached installer built by the old logic), `count_lines` fails loud instead of yielding a silent zero, and the symbol checker documents its cross-file-only scope.

### Note

The first commit (`c8d50efb`, jsonb docs) is pre-existing work tracked on `jsonb-hash-merge-clarification`; it rides along here and drops out of the diff if that merges first. The refactor commits are `58ca3188..e879fa70`.

Executed from `.work/eql-v3-build-ordering-refactor-plan.md`.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic ordering for v3 SQL files based on declared dependencies.
  * Added validation for installer completeness, symbol ordering, missing dependencies, cycles, and invalid references.
  * Added a command to display the calculated v3 installation order.

* **Bug Fixes**
  * Improved removal of dependency directives without affecting unrelated SQL content.
  * Builds now fail safely when ordering or installer validation detects errors.

* **Documentation**
  * Updated development and contributor guidance for v3 dependency ordering and validation.

* **Tests**
  * Added coverage for ordering, installer completeness, symbol dependencies, and build helper behavior.


freshtonic (commented):
## Review — build-ordering refactor (deterministic topo-sort)

Reviewed the whole refactor: `ordering.rs`, `build.sh`, `build/ordering.sh`, the `order` CLI (`main.rs`), `generate.rs`/`lib.rs`, `verify_symbol_order_v3.sh`, `verify_installer_complete.sh`, the parity/cli tests, and the doc/gitignore/CI edits. Scoped out the pre-existing `c8d50efb` jsonb-docs commit (its `src/v3/jsonb/operators.sql` is the only SQL-body change on the branch, and it rides along).

**Overall: excellent — this is exactly the right fix, and it's approve-worthy.** It replaces the non-deterministic, macOS-hostile `comment-parse → tsort → tac → cat` bash pipeline with a deterministic Rust topo-sort, and hardens completeness with two DB-free gates. (For context: I personally hit this class of bug twice last week — BSD-vs-GNU `tsort` producing different valid orders, and `| tac` being absent on stock macOS — so I can vouch that this eliminates a real, recurring footgun.) One thing to fix before merge: the **title/description are stale** (details below); the code is not blocked.

### Design — verified correct
- **`ordering.rs` (Kahn + min-heap keyed by path) is byte-reproducible.** Name-sorted tie-breaking gives a *canonical* order, not just *a* valid one — the property `tsort` never had. Determinism, cycle detection, self-edge tolerance, external-edge tolerance are all unit-tested.
- **One walk, not two enumerations.** `walk_v3_surface` + `surface_order` order exactly the set they walk, with no generated/hand-written classifier — so "a file matching neither predicate is silently dropped from the installer" (the old two-block risk) is *unrepresentable*. `install_order_contains_every_v3_sql_file` pins it against an independent walk. This is the standout design decision.
- **Invariants moved into the sort.** `OutsideSurface` (self-containment) and `UnknownTargets` (dangling REQUIRE) subsume the old `verify_v3_self_contained` / `verify_deps_exist` shell gates; the prefix check is path-segment-exact (`src/v3suffix/` correctly rejected). Cycles are a hard `CycleError` naming the stuck files.
- **Reorder-only confirmed.** The refactor commits touch **no** `src/v3/*.sql` body — the only SQL change on the branch is in the out-of-scope jsonb-docs commit. Combined with `codegen:parity` (generated files committed-in-place) and `strip_require_lines` (fail-loud, anchored), the "no SQL body modified" claim holds.

### Defense-in-depth gates — exemplary
- `verify_symbol_order_v3.sh`: define-before-use across the ordered list, deliberately stricter than PG (treats plpgsql body callees as references), with the allowlist escape hatch documented. Fail-safe on empty order / unreadable allowlist / unreadable path. Correctly handles the SEM types' split DDL forms (DOMAIN/TYPE/OPERATOR CLASS) and the CIP-3442 `eql_v3.query_*` domains.
- `verify_installer_complete.sh`: the reasoning here is the best part of the PR. Because 93/244 files are leaves, a dropped leaf body installs cleanly *and* passes the symbol checker (no dangling reference), surfacing only in a creds-gated DB test. Gate 3's **line-count identity** (`Σ(lines − REQUIRE lines) + pin == installer`) is pure arithmetic no leaf can hide from, and Gate 2 (trailing-newline) closes the `>>`-concat statement-gluing hole nothing else checks.

### To fix before merge — stale title/description
The PR title ("order the generated surface **from the codegen manifest**; **tsort as whole-surface verifier**") and the body's Task 1/Task 3 describe an intermediate design — a two-block scheme (shell-globbed hand-written + codegen manifest of generated) with `tsort` demoted to a verifier. Commit `8b9a19a7` ("order the whole v3 surface from one walk, not two enumerations") **superseded** that: the final state has **no `tsort` at all** (fully replaced by the Rust Kahn sort — only 3 comment references remain, all explaining its removal) and **no separate codegen manifest** (a single walk orders everything). Please retitle to describe the final single-walk design so the squash-merge message isn't misleading to a future `git log` reader. Code-wise this is purely cosmetic.

### Nits (non-blocking)
- `verify_symbol_order_v3.sh` opens each ordered file 3× (probe + definitions + references) — negligible at ~244 files, just noting.
- Gate 3's REQUIRE regex must stay in lockstep with `strip_require_lines`; the comment flags this, and a divergence fails the build loudly rather than silently, so it's fail-safe.

Great work — the single-walk invariant and the leaf-completeness arithmetic are the parts I'd hold up as the model for this kind of build-infra change.

freshtonic (approved):
## Re-review — approving

Re-reviewed after the retitle and `6fbc9079`. All feedback from my earlier comment is addressed, and the new commit fixes a real issue I'd missed. No remaining concerns — this is excellent, ship it.

**Resolved:**
- **Stale title/description → fixed.** Now "order the whole eql_v3 surface from one walk in the codegen; retire tsort" — accurately describes the final single-walk design (no tsort, no separate manifest). And, re my tsort comment on the thread: yep, you'd already retired it.

**New commit `6fbc9079` — correct and well-judged:**
- **The `#MISE sources` cache-invalidation fix is the important one, and I missed it.** `ordering.sh` (shaping the installer via `strip_require_lines`) and the two gate scripts + allowlist weren't in the task's sources, so an edit to any of them left mise considering the build fresh — re-serving a stale installer *and* skipping the gates (a cache hit skips the whole script). Adding all four is exactly right, and verifying the cache "discriminates rather than always rebuilds" is the right way to confirm it.
- **`count_lines` fail-loud** mirrors the `strip_require_lines` `rc <= 1` idiom — a swallowed grep error no longer reads as `0` and silently drops a file from the line-count identity. The note on keeping callers as bare `var=$(...)` (so `set -e` sees the failed assignment) is the correct subtlety.
- **Dropping the redundant `sort_unstable` in `topo_order`** — good call; the min-heap already owns ordering, so removing it makes the determinism source unambiguous (I'd noted the same privately).
- Documenting the symbol checker's cross-file-only scope, and making the two gate scripts executable, are both worthwhile.
- Confirmed the commit touches no SQL body — the reorder-only guarantee holds.

The single-walk invariant and the leaf-completeness line-count arithmetic remain the standout parts. Nice work.

PR #388, opened 2026-07-09
feat(install): disable ORE-backed domains loudly on non-superuser installs (CIP-3468)

freshtonic:
## Summary

`CREATE OPERATOR CLASS` requires superuser, so on cloud-hosted Supabase and most managed Postgres the installer has always attempted the ORE operator class and skipped it on `insufficient_privilege` — but the ORE-carrying domains still installed half-working: `<` / `>` comparisons ran as unindexable seq scans, while `CREATE INDEX ... (eql_v3.ord_term(col))` and bare `ORDER BY` failed with opaque Postgres errors.

The installer now capability-detects the skip (checking `pg_opclass` after the attempt) and **poisons all 38 ORE-carrying domains** — `_ord` / `_ord_ore` on every ordered scalar, `text_search`, and their `eql_v3.query_*` twins — with an always-raising `CHECK` constraint: the first value cast or inserted (including `NULL`) raises `feature_not_supported` (SQLSTATE `0A000`) naming the domain, with a `HINT` listing the same family's platform-supported alternatives (`_ord_ope` for indexed CLLW-OPE ordering, `_eq` for equality, `text_match` for pattern match).

Superuser installs are unchanged: the operator class is created and nothing is poisoned. Detection is install-time, so a superuser install keeps full ORE support regardless of which role queries later.

## Implementation

- **`src/v3/scalars/ore_fallback.sql`** — generated cross-family capability-detection file: a `DO` block that no-ops when the default btree opclass for `eql_v3_internal.ore_block_256` exists, otherwise creates the poison function and adds the `eql_ore_unavailable` constraint to every ORE-carrying domain.
- **`eql-codegen`** — new `ore_fallback.sql.j2` template + `render_ore_fallback_file()`, driven by the catalog (`scalar_families()` filtered on `Term::Ore`), so a new ORE-carrying family is covered by construction. Includes a depth-1 orphan sweep for cross-family generated files.
- **Footguns honoured** (per the encrypted-domain footgun list): the poison function is `LANGUAGE plpgsql` (never inlined — the `RAISE` cannot be planned away) and **not `STRICT`** (a `STRICT` function is skipped for `NULL`, which would let `NULL`s through silently).
- **Docs** — `docs/reference/database-indexes.md` (ORE-requires-superuser-install note + the `ord_ope_term` recipe) and upgrade note **U-003** in `docs/upgrading/v3.0.md`.
- **Changeset** — minor bump.

## Testing

`tests/sqlx/tests/v3_ore_fallback_tests.rs` runs the **actual shipped installer** (`release/cipherstash-encrypt.sql`) under `SET ROLE` to a `NOSUPERUSER` role against a fresh database:

- `nosuper_install_poisons_ore_domains` — every ORE-carrying domain (and query twin) raises `0A000` on first use, while every non-ORE domain in the same families stays fully functional. The poisoned/functional split is derived from the same catalog the generator renders from.
- `superuser_install_creates_opclass_and_poisons_nothing` — superuser installs get the opclass and zero `eql_ore_unavailable` constraints.

Also green locally: `mise run test:crates`, `mise run codegen:parity`, `mise run types:check`, `mise run build` (no regeneration drift).

Closes [CIP-3468](https://linear.app/cipherstash/issue/CIP-3468/do-not-install-ore-types-on-cloud-hosted-supabase)


tobyhede (commented):
Reviewed the diff, read the surrounding generator/installer code, and verified the load-bearing claims empirically against PG 17.

The design is right, the catalog-driven generation means a new ORE-carrying family is covered by construction, and the footguns (`LANGUAGE plpgsql`, non-`STRICT`) are both honoured and pinned by unit tests. Running the *actual shipped installer* under `SET ROLE` to a `NOSUPERUSER` role is the correct level to test this at. One blocking issue, then polish.

---

## Blocking: `ALTER DOMAIN … ADD CONSTRAINT` validates existing rows

`src/v3/scalars/ore_fallback.sql` adds the poison constraint without `NOT VALID`. PostgreSQL validates a newly added domain constraint against every stored value of that domain — and the poison function raises unconditionally, so the scan is guaranteed to fail. Reproduced on PG 17:

```sql
CREATE DOMAIN probe.d AS jsonb;
CREATE TABLE probe.t (x probe.d);
INSERT INTO probe.t VALUES ('{"a":1}');          -- one pre-existing row

ALTER DOMAIN probe.d ADD CONSTRAINT p CHECK (probe.poison(VALUE, 'probe.d'));
-- ERROR:  EQL: probe.d unavailable
-- CONTEXT:  PL/pgSQL function probe.poison(jsonb,text) line 2 at RAISE
```

This is reachable on a plain re-run of the installer — no uninstall needed. `src/v3/schema.sql` drops `eql_v3` / `eql_v3_internal` `CASCADE` at the top, but the user-column domains live in `public` precisely so they survive that, and `<T>_types.sql` guards them with `IF NOT EXISTS`. So a non-superuser install over a database with a single row in any `_ord` / `_ord_ore` / `text_search` column aborts inside the `DO` block. Because `sqlx::raw_sql` sends the script over the simple query protocol as one implicit transaction, the whole install rolls back — the user is left with no EQL rather than half of one, but it fails either way, and it fails with a message that reads like a query-time capability error:

```
ERROR:  EQL: public.integer_ord cannot be used on this platform: the EQL installer
        could not create the ORE operator class (requires superuser, …)
HINT:  Use public.integer_eq (equality) or public.integer_ord_ope (ordering) instead.
```

The affected users are exactly the ones this PR exists to help. The new test suite doesn't catch it because `#[sqlx::test(migrations = false)]` gives a fresh database, so no ORE column exists when the poison `ALTER` runs.

`NOT VALID` fixes it. For domains — unlike table constraints — it does not weaken enforcement: coercion applies every constraint regardless of `convalidated`. All four properties still hold; I checked each:

```sql
ALTER DOMAIN probe.d ADD CONSTRAINT p CHECK (probe.poison(VALUE,'probe.d')) NOT VALID;

SELECT count(*) FROM probe.t;              -- 1     existing rows still readable
INSERT INTO probe.t VALUES ('{"b":2}');    -- ERROR new values poisoned
INSERT INTO probe.t VALUES (NULL);         -- ERROR NULL still poisoned (non-STRICT holds)
SELECT '{"c":3}'::jsonb::probe.d;          -- ERROR casts still poisoned
```

It also skips a heavy-lock sequential scan of every dependent table across all 38 domains, which is worth having even on a fresh install.

In `crates/eql-codegen/templates/ore_fallback.sql.j2`:

```sql
ALTER DOMAIN {{ e.name }} ADD CONSTRAINT eql_ore_unavailable
  CHECK ({{ internal_schema }}.ore_domain_unavailable(VALUE, '{{ e.name }}', '{{ e.alternatives }}')) NOT VALID;
```

Regression test worth adding alongside: create a table with an `integer_ord` column, insert a row, *then* run the installer under `SET ROLE`; assert the install succeeds, the old row still selects, and a new insert raises `0A000`. That test fails on the current branch.

---

## Docs: `"v":"3"` is wrong, not just inconsistent

The U-003 verification snippets in `docs/upgrading/v3.0.md` use `'{"v":"3",…}'` (string), while U-001/U-002 four lines up use `"v":3` (number). Both pass the domain `CHECK`, because `VALUE->>'v' = '3'` text-coerces. But `SchemaVersion` is `pub struct SchemaVersion(#[ts(type = "3")] u16)` — the JSON Schema emits `{"type":"integer","const":3}` and the Rust `Deserialize` rejects a JSON string. So the snippet demonstrates a payload the published bindings reject; it only "works" because Postgres is more lenient than the wire contract. Should be `"v":3`.

## `text_search` collateral is worth a sentence in U-003

`search` carries `[Hm, Ore, Bloom]` (`crates/eql-domains/src/lib.rs:341`), so poisoning it removes equality and match too — capabilities that need no ORE opclass. U-003's remedy ("splits into `text_eq` + `text_match` + `text_ord_ope` columns as needed") is accurate for a fresh managed install, where there's no data to migrate. It's the superuser→managed migration case that involves re-encryption, and that's the one worth calling out explicitly.

## `RAISE NOTICE` advertises a domain that mostly doesn't exist

The closing notice says "use the `_ord_ope` (ordering), `_eq` (equality), or `_match` (match) domains instead". A `<T>_match` domain exists **only for `text`** — the other eight ordered families have none (`ORDERED_INT_DOMAINS`). The per-domain `HINT` is correct, since `ore_alternatives()` computes real siblings; only the generic one-liner over-promises.

## Nits

- `{{ e.name }}` is interpolated into a SQL string literal unescaped while `{{ e.alternatives }}` goes through `sql_str()`. Harmless (catalog identifiers can't contain quotes), and minijinja doesn't autoescape a `.sql`-named template so nothing is masking it — but the asymmetry invites a future bug.
- `OreFallbackContext.count` duplicates `entries|length`.

## Test coverage

Beyond the install-over-existing-data gap above: nothing asserts that pre-existing rows in a poisoned domain column remain readable, which is the property that makes `NOT VALID` safe.

The `ord_ope_term` btree recipe that U-003 prescribes *is* covered, in `tests/sqlx/tests/encrypted_domain/ope/integer_ord_ope.rs`, though under the superuser install. Since `ord_ope_term` returns `ope_cllw` (a domain over `bytea` with a native, install-role-independent opclass), asserting it under `NOSUPERUSER` too is nice-to-have rather than a real gap.

---

Looks good once `NOT VALID` lands with a test that would have caught it. The generator work, the footgun unit tests, and the real-installer-under-`SET ROLE` harness are all well judged.

<sub>Note: the working tree's `eql-codegen` unit tests fail against a stale incremental cache (`integer_ord` renders `ord_ope_term`) — `cargo clean -p eql-domains -p eql-codegen` restores 97/97. Not a defect in this PR; noting it because it looks alarming and wastes time.</sub>

PR #389, opened 2026-07-09
feat(v3)!: back the `_ord` domains with CLLW-OPE instead of block-ORE

tobyhede:
## Summary

Switches the **default** ordering domain family from block-ORE to CLLW-OPE.

`public.<T>_ord` (every ordered scalar) and `public.text_ord` now carry the `op` term, extract via `eql_v3.ord_term(col)` returning `eql_v3_internal.ope_cllw`, and order by native `bytea` comparison.

**Nothing is removed.** Block-ORE remains fully supported behind `public.<T>_ord_ore`, whose extractor is now the qualified `eql_v3.ord_term_ore(col)`. `public.text_search` stays `[hm, ob, bf]`. `_ord` / `_ord_ope` are now the byte-identical twins (previously `_ord` / `_ord_ore`).

## Why

`eql_v3_internal.ope_cllw` is a `DOMAIN … AS bytea`, so an ordered functional btree resolves to **`bytea_ops`** — the base type's default operator class, whose opfamily already contains the `<` `<=` `>` `>=` the planner must match. No custom comparator, no `pgcrypto` in the compare path, no superuser.

The block-ORE path depends on a hand-written opclass for a *composite* type, created by a `DO` block that silently skips on `insufficient_privilege`. The failure mode is quiet, not loud: with the opclass absent, `CREATE INDEX … btree (eql_v3.ord_term_ore(col))` still **succeeds** — PostgreSQL binds `record_ops` — and the index then never engages, because the ORE operators are not members of that opfamily. An index that costs writes and buys nothing. I verified this against a live non-superuser role with the opclass removed.

## Relationship to #388 (CIP-3468)

#388 landed after this branch forked. It capability-detects the skipped opclass and **poisons** every `Term::Ore`-bearing domain, telling callers to use `_ord_ope` instead.

This PR removes the largest cause of that poison. `ore_fallback.sql` is catalog-generated from `d.terms.contains(&Term::Ore)`, so once `_ord` carries `Term::Ope` the generator stops poisoning it — the poison set narrows to `_ord_ore` and `text_search`, and #388's remediation advice ("use `_ord_ope` for ordering") becomes satisfied by the default `_ord` itself.

**This branch needs a rebase onto `eql_v3` before merge.** Expected work:
- regenerate `src/v3/scalars/ore_fallback.sql` (catalog-driven; `_ord` drops out of the poison set)
- update `tests/sqlx/tests/v3_ore_fallback_tests.rs`, which asserts `public.integer_ord` is poisoned (should become `public.integer_ord_ore`)
- resolve the one conflict, in `docs/upgrading/v3.0.md` (both PRs add an upgrade note)

## Breaking changes

- Rows in an `_ord` column must carry `op` rather than `ob`. Existing `_ord` indexes must be rebuilt.
- `eql_v3.ord_term(public.<T>_ord_ore)` is renamed `eql_v3.ord_term_ore` — a pure rename, same term, same behaviour, DDL text only.
- The pre-release-only `eql_v3.ord_ope_term` is removed; its overloads fold into `eql_v3.ord_term`.

Rollback for a caller is `ALTER TABLE … ALTER COLUMN … TYPE public.<T>_ord_ore` — zero re-encryption.

## Behaviour changes (both verified against real ciphertext)

**`public.text_ord` now accepts the empty string** and sorts it first. Encrypting `""` yields an empty, unorderable ORE term (`ob: []`) that `text_ord_ore` still rejects (#262); OPE yields a well-formed one-byte `op` term that is a proper prefix of every non-empty term.

**`real` / `double` `_ord` now distinguish `-0.0` from `+0.0`** (`-0.0 < +0.0`, `-0.0 <> +0.0`). `orderable-bytes` canonicalizes the sign of zero for ORE; CLLW-OPE does not. This makes `_ord` **consistent with `public.<T>_eq`**, which already distinguished them — see #387. A float column that must treat ±0.0 as one value should use `_ord_ore`.

Both are pinned by tests and documented in U-003.

## Also in this PR

**`float_special` now runs in CI.** It never had, since the commit that created it: the sharded job compiles with default features so the `#[cfg(feature = "proptest-e2e")]` module was never built, and the `e2e` job's `e2e_oracle` name filter never matched it. Its NaN / ±0.0 / ±Inf encoder tripwires guarded nothing.

Turning it on surfaced **#387**: `-0.0` and `+0.0` hash to different `hm` terms, because the client feeds `f64::to_be_bytes()` — sign bit included — into the HMAC. `WHERE col = 0.0` misses rows stored as `-0.0`. Not caused by, and independent of, the `_ord` SEM.

Rather than `#[ignore]` it, `known_failure()` **inverts** the assertion: the test is written the way it should pass, the marker passes while the bug reproduces (printing the issue URL every run), and **fails the moment the assertion starts passing**. A marker cannot outlive its bug. The other half is a new credential-free `known-failures` CI job: every `ISSUE_*` constant must name an issue that exists, is still OPEN, and is referenced by a test.

## Verification

- catalog / codegen / bindings crate tests, matrix inventory, `codegen:parity`, `types:check`, `typescript:check`
- `self_contained_v3`, `splinter` (0 unmatched — the new extractors are recognised structurally), doc coverage 100%
- public-surface golden re-blessed; matrix baselines re-blessed
- empty-string and signed-zero behaviour confirmed against real ZeroKMS ciphertext
- non-superuser role: `CREATE INDEX … btree (eql_v3.ord_term(c))` on `_ord` binds `bytea_ops` and succeeds

## Review notes (not yet actioned)

Four review passes flagged, in priority order:

1. `tasks/test/known-failures.sh` — the "marker referenced by no test" diagnostic is unreachable (`set -euo pipefail` aborts on `grep`'s no-match exit before the check), and the `ISSUE_*` parse **fails open**: if the regex matches nothing the gate prints "nothing to verify" and exits 0.
2. `tests/sqlx/src/property.rs` — `assert_extractor_oracle`'s `Some(other) => bail!(…)` catch-all defeats compile-time exhaustiveness (`Term` is not `#[non_exhaustive]`); payload keys `"ob"`/`"op"` are hardcoded where `Term::json_key()` exists.
3. `float_special` covers `F8` only. The signed-zero change applies to `real_ord` (`F4`) too, untested.

`mapfile` in `known-failures.sh` is bash 4+; macOS `/bin/bash` is 3.2.


mise.toml
```diff
@@ -106,21 +106,42 @@ cargo test --features proptest-e2e
...
+# catches it for you.
 depends = ["test:sqlx:prep"]
 dir = "{{config_root}}/tests/sqlx"
 run = """
+#!/usr/bin/env bash
+set -euo pipefail
 echo "Running e2e property suite (fresh ZeroKMS encryption)..."
 cargo test --features proptest-e2e e2e_oracle
+echo "Running float special-value suite (fresh ZeroKMS encryption)..."
+cargo test --features proptest-e2e float_special
```
coderdan, line 134: **Blocking — `v3_scalar_query_operand_tests` runs in no CI job**

It's now the only `#![cfg(feature = "proptest-e2e")]` test binary left in the repo, and nothing executes it:

- `tasks/test/sqlx-archive.sh:43` runs a bare `cargo nextest archive` with no `--features proptest-e2e`, so the sharded `test` job compiles zero tests from it.
- `test:sqlx:e2e` names only the `e2e_oracle` and `float_special` filters, neither of which matches any test in that binary.

This PR edited that file — renaming `ord_term_only_operand_orders_via_the_ore_operator` to `..._ope_operator` and swapping `IndexKind::Ore` → `IndexKind::Ope` so the operand satisfies the new `integer_ord` `op` CHECK. That assertion has never executed. Break the `(integer_ord, query_integer_ord)` OPE ordering operator, or revert the OPE migration, and CI stays green.

This is the same orphaning bug the PR diagnosed and fixed for `float_special`, one file over — and the comment at line 122 says it out loud: *"Adding a new `proptest-e2e`-gated module means adding a line here; no glob catches it for you."*

```diff
 cargo test --features proptest-e2e float_special
+echo "Running scalar query-operand conformance (fresh ZeroKMS encryption)..."
+cargo test --features proptest-e2e scalar_query_operand
```

tasks/test/known-failures.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "==> verifying ${#entries[@]} known-failure marker(s) against ${repo}"
+
+# A marker whose constant is never referenced by a test is dead weight: the bug
+# may be long fixed and nobody noticed. Catch that too.
+fail=0
+for entry in "${entries[@]}"; do
+  name=${entry%% *}
+  number=${entry##* }
+
+  refs=$(grep -rl --include='*.rs' -- "$name" "${EQL_ROOT}/tests/sqlx/tests" 2>/dev/null | wc -l | tr -d ' ')
```
coderdan: **Blocking — `set -euo pipefail` makes the "referenced by no test" branch dead code**

When `grep` finds no match it exits 1, `pipefail` propagates that through `| wc -l | tr -d ' '`, and `set -e` kills the script at the assignment — so lines 48–52 are unreachable.

That's precisely the case this check exists to catch. Reproduced with a registered-but-unreferenced constant: the script exits 1 having printed **nothing**. The intended `✗ ... is registered but referenced by no test` never prints, the remaining markers go unchecked, and the `known-failure gate FAILED` summary never runs. The new required `known-failures` job fails with an empty log.

```diff
-  refs=$(grep -rl --include='*.rs' -- "$name" "${EQL_ROOT}/tests/sqlx/tests" 2>/dev/null | wc -l | tr -d ' ')
+  refs=$(grep -rl --include='*.rs' -- "$name" "${EQL_ROOT}/tests/sqlx/tests" 2>/dev/null | wc -l | tr -d ' ' || true)
```

tasks/test/known-failures.sh
```diff
@@ -0,0 +1,79 @@
...
+  if [ "$refs" -eq 0 ]; then
+    echo "  ✗ ${name} (#${number}) is registered but referenced by no test — remove it" >&2
+    fail=1
+    continue
+  fi
+
+  state=$(gh issue view "$number" --repo "$repo" --json state -q .state 2>/dev/null || echo "MISSING")
```
coderdan, line 58: **Blocking — A transient `gh` failure is indistinguishable from a nonexistent issue**

`|| echo "MISSING"` catches *any* non-zero exit — a 5xx, a rate-limit, an unauthenticated `gh` — and `2>/dev/null` throws away the reason. All of it lands in the `*)` branch, which fails the gate with *"could not be read ... A known-failure marker must name a real issue"* even when the issue is open and valid.

Consequences: `ci-required` fails a PR on a network blip while pointing at the wrong problem, and locally, anyone without `gh auth login` sees every valid marker reported as bogus.

Worth separating gh's exit status from a genuine `NOT_FOUND`, and surfacing stderr on the unexpected path rather than discarding it.

tests/sqlx/tests/encrypted_domain/float_special.rs
```diff
@@ -111,17 +130,83 @@ async fn two_encryptions_of_same_nan_bits_compare_equal() -> Result<()> {
...
+    );
+    assert!(
+        !ord_ore_cmp(&pool, &p[1], "<", &p[0]).await?,
+        "+0.0 not < -0.0 under block-ORE"
+    );
+    Ok(())
+}
+
+#[tokio::test]
+async fn negative_zero_orders_below_positive_zero_under_ope() -> Result<()> {
```
coderdan, line 158: **Blocking — `=` on `_ord` now splits ±0.0, and nothing pins it**

`double_ord_functions.sql:24` renders `eql_v3.eq(a,b) := ord_term(a) = ord_term(b)`, and `_ord` is `[Ope]`-only — so `=` compares OPE bytea. This test proves `op(-0.0) != op(+0.0)`.

So `WHERE double_ord_col = 0.0` now misses rows stored as `-0.0`. Under the previous block-ORE `_ord` it matched, because `orderable-bytes` canonicalizes `-0.0 → +0.0` (pinned two tests up in `negative_zero_and_positive_zero_share_ore_order`).

Only the `<` divergence is pinned here. The `=`/`<>` facet on `_ord` is untested, and `-0.0` is excluded from `DOUBLE_FIXTURES`/`REAL_FIXTURES`, so the matrix can't catch it either.

The changeset justifies the new behaviour as *"consistent with `public.<T>_eq`, which already distinguished them"* — but that `_eq` behaviour is #387, tracked as a bug to be **fixed**. Once #387's root cause is repaired for the `hm` term, `_eq` stops distinguishing the zeroes, `_ord` becomes the lone domain that does, the rationale evaporates, and no test turns red.

Suggest adding the `=`/`<>` assertion on `_ord` alongside the ordering one, so the two facets can't silently diverge.

tasks/test/splinter.sh
```diff
@@ -85,19 +85,29 @@ function_search_path_mutable	eql_v3_internal	hmac_256	function	HMAC equality ext
...
-function_search_path_mutable	eql_v3_internal	ope_cllw	function	CLLW-OPE extractor for the eql_v3 SEM fork (now in eql_v3_internal): inlinable SQL (jsonb) constructor used inside eql_v3.ord_ope_term, hex-decoding `op` to the bytea-backed eql_v3_internal.ope_cllw domain. The domain inherits bytea's native comparison operators and btree opclass, so the WHOLE comparison chain (wrapper -> ord_ope_term -> this) is inlinable SQL and the functional btree index on eql_v3.ord_ope_term(col) engages structurally — the hmac_256 pattern. Carries the `eql-inline-critical` COMMENT marker (bare jsonb arg escapes the structural domain-arg skip).
+function_search_path_mutable	eql_v3_internal	ore_cllw_eq	function	Inner comparator for the eql_v3_internal.ore_cllw composite type's `=` operator (self-contained SEM fork, DEFAULT FOR TYPE btree opclass eql_v3_internal.ore_cllw_ops). The outer same-type operators back the opclass; the planner only carries the inlined form through to functional-index match if this inner function is also inlinable (no SET, IMMUTABLE). The plpgsql FUNCTION 1 comparator (compare_ore_cllw_term) stays pinned by design.
```
coderdan, line 88: **Blocking — Twelve allowlist rows for `ore_cllw`, which doesn't exist**

`rg ore_cllw src/v3 crates/` matches only a doc comment in `src/v3/sem/ope_cllw/types.sql`. `src/v3/sem/` holds exactly `bloom_filter`, `hmac_256`, `ope_cllw`, `ore_block_256` — the `ore_cllw` type, its six comparators, and `has_ore_cllw` are all gone (removed on the base branch; v3.0.md U-004 lists them under Removed).

The eight `ore_cllw*` rows added here can never match a splinter finding, so they're inert today. But the allowlist is the record-of-record for *why* each lint is waived, and a future `ore_cllw` would arrive silently pre-waived.

Worse, two **live** rows had their prose regressed in the same hunk:

- `eql_v3.lt` (line 72) now says the jsonb_entry wrapper routes "via `ore_cllw`" — `src/v3/jsonb/operators.sql:307` shows it routes via `eql_v3.ord_term` → `ope_cllw`.
- `eql_v3.->` says the chained recipes are "`eq_term` / `ore_cllw`" — same problem.

Given the `fix(rebase): restore resolutions clobbered by later-commit replays` commit in this series, this file may not be the only casualty — the same reversed-rename artifact shows up in `fixture_oracle.rs`, `scalar_domains.rs`, `matrix.rs`, and `ope_cllw/types.sql` (all commented separately).

src/v3/sem/ope_cllw/types.sql
```diff
@@ -15,9 +14,9 @@
 --! outright — no hand-written operators, comparator, or operator class (the
 --! same pattern as eql_v3_internal.hmac_256 over text). That keeps the whole
 --! comparison chain inlinable, so a functional btree index on
---! `eql_v3.ord_ope_term(col)` engages structurally for the `_ord_ope`
---! domains' comparison operators — and likewise on
---! `eql_v3.ord_ope_term(col -> 'selector')` for SteVec entry ordering.
+--! `eql_v3.ord_term(col)` engages structurally for the `_ord` / `_ord_ope`
+--! domains' comparison operators. Contrast eql_v3_internal.ore_cllw (`oc`), the SteVec
```
coderdan, line 18: **Should fix — Shipped Doxygen contrasts against a type that no longer exists**

This is a `--!` block, so it ships into `docs/api/xml` and the published API reference.

`eql_v3_internal.ore_cllw` was removed on the base branch. And SteVec entries don't order by a "CLLW-*ORE* composite compared by a custom per-byte protocol" — they order via `eql_v3.ord_term(jsonb_entry)` → `ope_cllw`, i.e. *this very type* (`src/v3/jsonb/functions.sql:121`, `operators.sql:306-374`, `aggregates.sql:50`).

The edit also deleted the still-true sentence saying `ope_cllw` backs SteVec entry ordering, and line 8 now claims `op` appears only on "the `_ord_ope` domains" — omitting the `_ord` domains this PR just moved onto `op`.

Net effect: the published docs point a reader at a deleted type and give them the wrong ordering mechanism.

tests/sqlx/tests/encrypted_domain/property/fixture_oracle.rs
```diff
@@ -479,7 +479,7 @@ mod text_fn {
     }
 
     /// `text_search` ([Hm, Ore, Bloom]) — the eq/ord function facets plus
```
coderdan, line 481: **Should fix — Blanket `ord_term` → `ord_term_ore` rename inverted these comments**

`text_search` is `[Hm, **Ope**, Bloom]` after this PR — `[Hm, Ore, Bloom]` is now `text_search_ore`, a different domain.

And the oracle immediately below checks the OPE path: `search_fn_oracle` calls `assert_extractor_oracle(Variant::Search)`, which reads `spec.ord_term = ordering_term("text") = Term::Ope` and asserts the extracted term equals payload `op` via `eql_v3.ord_term`.

Same inversion at line 467 (`text_ord ... ord_term_ore(\`ob\`) identity`) and line 343 (`ord_term_ore==ob`). All three were changed by this PR in the wrong direction — the mechanical rename swept up the domains that moved *to* OPE. A maintainer reading these would believe the OPE oracle rides the block-ORE path and could "fix" the test toward `ord_term_ore`, breaking it.

tests/sqlx/src/scalar_domains.rs
```diff
@@ -1316,7 +1358,21 @@ pub struct ScalarDomainSpec {
     pub variant: Variant,
     pub placeholder_payload: &'static str,
     pub eq_extractor: fn(&str) -> String,
-    pub ord_extractor: fn(&str) -> String,
+    /// Takes the variant because one `T` spans domains with different ordering
+    /// terms. Call via [`ScalarDomainSpec::ord_extractor_expr`], which supplies
+    /// `self.variant`.
+    pub ord_extractor: fn(Variant, &str) -> String,
+    /// The variant's ordering term (`Term::Ope` for `_ord`, `Term::Ore` for
+    /// `_ord_ore` / `_search`), or `None` when the domain is not ordered.
```
coderdan, line 1366: **Should fix — This field's doc states the inverse of the change**

`crates/eql-domains/src/lib.rs:368-373` defines `search => [Hm, Ope, Bloom]` and `search_ore => [Hm, Ore, Bloom]`. So `_search` is the **Ope** one; `Term::Ore` belongs to `_ord_ore` / `_search_ore`.

This is the one field whose entire job is recording which SEM backs each variant, so getting it backwards is costly.

Line 1371's `orders by \`oc\`` is wrong too: `oc` is a v3-removed wire key that the `jsonb_entry` CHECK actively rejects (it requires `hm` XOR `op`). `tests/sqlx/src/matrix.rs:522` repeats the same `oc` error, two lines after the code comment correctly names `eql_v3.ord_term for entries`.

docs/upgrading/v3.0.md
```diff
@@ -129,10 +140,12 @@ comparisons ran as unindexable seq scans, while
...
 bare `ORDER BY` could not find an ordering operator. From 3.0.0 the installer
-capability-detects the skip and **poisons every ORE-carrying domain** — `_ord`
-/ `_ord_ore` on each ordered scalar family, `text_search`, and their
-`eql_v3.query_*` twins (38 domains) — with an always-raising `CHECK`
-constraint. The first value cast or inserted into one (including `NULL`)
+capability-detects the skip and **poisons every ORE-carrying domain** —
+`_ord_ore` on each ordered scalar family, `text_search_ore`, and their
+`eql_v3.query_*` twins — with an always-raising `CHECK`
+constraint. The set is derived from the catalog, so it shrinks automatically as
+families move off block-ORE: `_ord` (see U-004) and `text_search` (see U-005) are
```
coderdan, line 147: **Should fix — U-note cross-links are off by one**

The headings in this file are:

- `U-004: SteVec ordering terms are CLLW-OPE` (line 193)
- `U-005: _ord domains are backed by CLLW-OPE` (line 250)
- `U-006: text_search is backed by CLLW-OPE` (line 414)

So `_ord (see U-004)` should be **U-005**, and `text_search (see U-005)` should be **U-006**. Every other reference in the file maps them that way — these two anchors send an operator auditing their schema to the wrong section.

tests/sqlx/tests/v3_privilege_tests.rs
```diff
@@ -34,7 +34,7 @@ const EQ_QUERY: &str = "SELECT count(*) FROM fixtures.eql_v3_integer \
      WHERE payload::public.integer_eq = payload::public.integer_eq";
 
 /// A real ordering query using the `<` *operator* on `integer_ord`, which dispatches
```
coderdan, line 36: **Consider — Ordering-path comments and the `extensions` grant are now stale**

`<` on `integer_ord` inlines `eql_v3.ord_term` → `eql_v3_internal.ope_cllw` with native bytea comparison: no `ore_block_256`, no pgcrypto.

The suite still passes (ope_cllw lives in `eql_v3_internal`), so the module doc at line 8, this `ORD_QUERY` comment, and the `GRANT USAGE ON SCHEMA extensions` at lines 172-178 are stale — and the grant is now dead.

The consequence is a coverage hole: `docs/reference/permissions.md` was updated in this PR to claim pgcrypto is needed *only* on `_ord_ore`/`text_search_ore`, but this suite exercises only the OPE `integer_ord` path while still granting `extensions`. A change that reintroduced a pgcrypto dependency on the `_ord` ordering path would violate the newly-written doc and still pass here.

Re-pointing `ORD_QUERY`/`AGG_QUERY` at `integer_ord_ore`, or adding an OPE-path control that omits the extensions grant, would pin the distinction.

crates/eql-domains/src/lib.rs
```diff
@@ -187,7 +197,7 @@ const ORDERED_INT_DOMAINS: &[Domain] = &[
     },
     Domain {
         name: "ord",
```
coderdan, line 215: **Consider — `_ord` and `_ord_ope` are now a duplicated surface, not an alias**

With `_ord` on `Term::Ope`, the two domains are indistinguishable — same terms, same `eql_v3.ord_term` extractor — and the PR's own `ordered_files_byte_identical_modulo_typename` asserts they render identically modulo type name. The doc at line 177 acknowledges it ("`_ord` and `_ord_ope` are therefore the twins now").

But the twin is *materialized*, not aliased. Across 9 ordered families that's ~45 generated `*_ord_ope*.sql` files (~8k lines), roughly a thousand redundant DDL statements in every `release/cipherstash-encrypt.sql` install, plus 18 duplicate TS files, 18 duplicate JSON schemas, and 9 duplicate Rust structs shipped to crates.io and npm. `TextOrd` and `TextOrdOpe` in `eql-bindings/src/v3/text.rs:143/176` differ only in one string literal.

If `_ord_ope` is being kept purely for name stability, rendering it as an alias of `_ord` would halve the ordered surface and remove a duplicate that every future codegen change has to be diffed against twice.

mise.toml
```diff
@@ -369,14 +390,14 @@ discovered=$(printf '%s\\n' "$listing" \
...
-# (`_ord` / `order_by` / `routes_through_ob`). Derive that subset once into a
+# (`_ord` / `order_by` / `routes_through_ordering_term`). Derive that subset once into a
 # file, so an equality-only scalar (e.g. timestamp) needs NO second committed
 # snapshot — it is validated against this derivation from the single baseline.
 # Materialise to a file (not a shell variable) so both shapes compare via the
 # same `cmp`/`diff` path: a variable would strip trailing newlines on capture
 # and force the failure diff to reconstruct the string with `printf`.
 eq_only_expected="/tmp/matrix-eq-only-expected.txt"
-grep -vE '_ord|order_by|routes_through_ob' snapshots/matrix_tests.txt | LC_ALL=C sort -u > "$eq_only_expected"
+grep -vE '_ord|order_by|routes_through_ordering_term' snapshots/matrix_tests.txt | LC_ALL=C sort -u > "$eq_only_expected"
```
coderdan, line 404: **Consider — The third alternative in this filter is dead, and it's copy-pasted 4×**

`routes_through_ordering_term` contains the substring `_ord` (in `through_**_ord**ering`), so the first alternative already matches it. Verified: `grep -vE '_ord|order_by|routes_through_ordering_term'` and `grep -vE '_ord|order_by'` both leave 54 lines of `snapshots/matrix_tests.txt`.

The old token `routes_through_ob` had no `_ord` substring, so it *was* load-bearing — the rename silently made it inert.

The full filter is duplicated at mise.toml:400, mise.toml:410 (the regenerate hint echoed to the user on failure), mise.toml:556, and `tests/sqlx/snapshots/README.md:49`, plus several comment copies. A future test rename that drops the `ord` substring must be applied to all four in lockstep — in a filter no maintainer now realises is doing nothing.

.changeset/ord-defaults-to-ope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cipherstash/eql': minor
```
coderdan, line 2: **Consider — `minor` for a change the PR title marks breaking**

The commit is `feat(v3)!:` and the body of this changeset describes: renaming the public `eql_v3.ord_term(<T>_ord_ore)` to `ord_term_ore`, changing `eql_v3.ord_term`'s return type from `ore_block_256` to `ope_cllw`, removing `eql_v3.ord_ope_term`, and changing the required payload key for `_ord` from `ob` to `op` ("existing `_ord` indexes must be rebuilt").

CLAUDE.md: *"**Minor (`2.x.0`)** — additive changes, behaviour changes that don't break the public API (signatures, schema name, payload format, operator names)"* / *"**Major (`3.0.0`)** — only for changes that break the public API."* Signatures, operator names, and payload format are all on the major side of that line. `.changeset/eql-3469-stevec-ope.md`, the analogous OPE switch, is marked `major`.

The computed version is unaffected this cycle (other `major` changesets are already in the set, and changesets is in pre-mode), so this is classification only. `search-defaults-to-ope.md:2` has the same issue.

coderdan (commented):
## Review: `_ord` → CLLW-OPE

The core of this change is sound, and I verified the load-bearing parts rather than assuming them:

- Regenerated the codegen (`cargo run -p eql-codegen`, then `-- bindings`) and diffed — **no drift**, `codegen:parity` and the Rust half of `types:check` will pass.
- `cargo test -p eql-domains -p eql-codegen` — all green (98 + 93 + parity/CLI).
- Traced the inlining chain: wrapper → `eql_v3.ord_term` → `eql_v3_internal.ope_cllw` is `LANGUAGE sql` / `IMMUTABLE` / single-`SELECT` / unpinned end to end, and `ope_cllw` is a `DOMAIN … AS bytea` inheriting `bytea_ops`. The functional btree really does engage without a superuser opclass — which is the whole point of the change.
- All 1565 generated blockers are `LANGUAGE plpgsql` and none are `STRICT`. No domain-over-domain, no opclass on a domain.
- `ore_fallback.sql` poisons `text_search_ore` and correctly leaves `text_search` alone.
- The public-surface golden contains the new `text_search_ore` surface and both `ord_term` overloads.

I've left **15 inline comments**. Three I'd want resolved before merge:

1. **`v3_scalar_query_operand_tests` runs in no CI job** (`mise.toml:132`) — the exact orphaning bug this PR fixes for `float_special`, left in place for the sibling file the PR itself edited onto the new OPE `_ord`. That assertion has never executed.
2. **The new `known-failures` gate is broken in the case it exists for** (`known-failures.sh:47`) — `set -euo pipefail` kills the script before the "referenced by no test" diagnostic can print. Reproduced: exits 1, prints nothing.
3. **`=` on `_ord` now splits ±0.0 for `real`/`double`**, and nothing pins it. The changeset justifies this as consistency with `_eq` — but that `_eq` behaviour is #387, tracked as a bug to be fixed. When it is, the rationale evaporates and no test turns red.

Plus a likely rebase casualty: `splinter.sh` gained twelve allowlist rows for `ore_cllw`, a type that exists nowhere in `src/v3`. The same reversed-rename artifact shows up in `fixture_oracle.rs`, `scalar_domains.rs`, `matrix.rs`, and the shipped Doxygen in `ope_cllw/types.sql`.

---

### Two findings I couldn't anchor inline

**`tests/sqlx/tests/v3_ore_fallback_tests.rs:311`** — *the superuser positive control is now vacuous.*
`src/v3/scalars/ore_fallback.sql:90` only ALTERs the `_ord_ore` domains and `text_search_ore`, never `public.integer_ord`. This PR flipped `integer_ord` from `[Ore]` to `[Ope]`, so the `try_cast(&pool, "public.integer_ord", …)` at line 311 — whose stated purpose is *"ORE domains accept values on a superuser install"* — now exercises a domain that is unpoisonable by construction. The sibling negative tests (`nosuper_install_poisons_ore_domains`, `reinstall_over_ore_data`) were correctly re-pointed to `integer_ord_ore`; this one wasn't. The preceding `poison_constraints == 0` check still does real work, so this is a weakened guard rather than an absent one. Fix: `.find(|d| d.name == "ord_ore")` and cast to `public.integer_ord_ore`.

**`crates/eql-domains/src/fixtures/record.rs:82`** — *no scale-equivalent numeric pair.*
`lib.rs:289` asserts the ordering term agrees with `Decimal: Ord`, *"equivalent scales collide"* — a guarantee previously established for block-ORE via `ore-rs`. `numeric._ord` is now `[Term::Ope]`. Every other degenerate case the swap touches got a targeted test (float ±0.0 → `float_special.rs`; empty string → `v3_text_empty.rs`), but scale collision didn't, because `NUMERIC_FIXTURES` (`"-1.001", "-1", "-0.5", …, "1", "1.001"`) contains no two values that are `Decimal`-equal at different scales. If cipherstash-client's OPE for `Plaintext::Decimal` doesn't normalize scale, then `op(1) != op(1.0)`: `WHERE numeric_ord_col = 1` silently misses a row stored as `1.0`, and `ORDER BY` splits equal values. Adding `"1.0"` to the fixture list would close it.

<sub>Reviewed with Claude Code at `xhigh` effort: 10 finder angles → verify → gap sweep. Every finding was checked against the working tree; the shell and grep claims were reproduced.</sub>

tasks/test/known-failures.sh
```diff
@@ -0,0 +1,79 @@
...
+
+[ -f "$REGISTRY" ] || { echo "missing known-failure registry: $REGISTRY" >&2; exit 2; }
+
+# The repo the issue numbers refer to, read from the registry so it cannot drift.
+repo=$(sed -n 's/^pub const KNOWN_FAILURE_REPO: &str = "\(.*\)";$/\1/p' "$REGISTRY")
+[ -n "$repo" ] || { echo "could not read KNOWN_FAILURE_REPO from $REGISTRY" >&2; exit 2; }
+
+# Every `pub const ISSUE_<NAME>: u64 = <n>;` in the registry.
+mapfile -t entries < <(
+  sed -n 's/^pub const \(ISSUE_[A-Z0-9_]*\): u64 = \([0-9]*\);$/\1 \2/p' "$REGISTRY"
```
tobyhede, line 30: **Blocking — the gate fails open: a format tweak silently disables it**

This regex is fully anchored and hard-codes `u64` at column zero with no trailing content. Any deviation yields zero matches, and the script then prints `no known-failure markers registered — nothing to verify` and **exits 0**.

Tested each of these against the real script — all silently disable the entire gate:

| registry line | parsed? |
|---|---|
| `pub const ISSUE_FOO: u64 = 387;` | yes |
| `pub const ISSUE_FOO: u32 = 387;` | **no** |
| `    pub const ISSUE_FOO: u64 = 387;` (indented, e.g. moved into a `mod`) | **no** |
| `pub const ISSUE_FOO: u64 = 387; // note` | **no** |
| `pub const ISSUE_FOO: u64 = 1_000;` (valid Rust) | **no** |

Today's single constant matches, so the gate works; the fail-open is latent. This is distinct from the `pipefail` issue below — the gate can be neutered without ever reaching line 47.

Suggest counting the declarations independently and hard-erroring on a mismatch, rather than treating zero as "nothing to verify":

```bash
declared=$(grep -cE '^[[:space:]]*pub const ISSUE_[A-Z0-9_]*:' "$REGISTRY")
if [ "${#entries[@]}" -ne "$declared" ]; then
  echo "registry declares $declared ISSUE_ consts but ${#entries[@]} parsed — fix the parser" >&2
  exit 2
fi
```

tasks/test/known-failures.sh
tobyhede, line 58: **Should fix — `gh issue view` resolves pull-request numbers, so a marker can "name" a PR**

Issues and PRs share one numbering space and `gh issue view` resolves both. Confirmed against this repo:

```
$ gh issue view 389 --repo cipherstash/encrypt-query-language --json state -q .state
OPEN
```

389 is this pull request. An `ISSUE_*` constant that (by typo, or by pasting a PR link) holds a PR number passes the gate as a valid open issue, and the marker it guards is then anchored to something that will close on merge.

Request the entity type and reject PRs — `--json state,isPullRequest` (or the GraphQL `__typename`).

This is a different failure from your transient-`gh` comment on the same line: that one is a false negative on a network blip, this is a false positive on the wrong entity. Both live here.

tasks/test/known-failures.sh
```diff
@@ -0,0 +1,79 @@
...
+REGISTRY="${EQL_ROOT}/tests/sqlx/src/known_failure.rs"
+
+[ -f "$REGISTRY" ] || { echo "missing known-failure registry: $REGISTRY" >&2; exit 2; }
+
+# The repo the issue numbers refer to, read from the registry so it cannot drift.
+repo=$(sed -n 's/^pub const KNOWN_FAILURE_REPO: &str = "\(.*\)";$/\1/p' "$REGISTRY")
+[ -n "$repo" ] || { echo "could not read KNOWN_FAILURE_REPO from $REGISTRY" >&2; exit 2; }
+
+# Every `pub const ISSUE_<NAME>: u64 = <n>;` in the registry.
+mapfile -t entries < <(
```
tobyhede, line 29: **Consider — `mapfile` is bash 4+; this dies on a stock macOS clone**

`mise.toml` invokes `bash tasks/test/known-failures.sh`, which bypasses the `#!/usr/bin/env bash` shebang and uses whatever `bash` is on PATH.

On my machine that's Homebrew bash 5.3, so it runs. But `/bin/bash` is 3.2.57, which has no `mapfile`:

```
$ /bin/bash -c 'mapfile -t x < <(echo hi)'
/bin/bash: line 1: mapfile: command not found
```

CI is Linux/bash 5, so this never turns CI red — it just breaks for any contributor without a Homebrew bash, at the first line of the loop. A 3.2-safe read loop costs two lines:

```bash
entries=()
while IFS= read -r line; do entries+=("$line"); done < <(sed -n '...p' "$REGISTRY")
```

(Checked the adjacent worry and it's fine: `${#entries[@]}` under `set -u` on an empty `mapfile` array does **not** error on bash 4+.)

tests/sqlx/src/property.rs
```diff
@@ -458,41 +480,71 @@ pub async fn assert_extractor_oracle<T: ScalarType>(
...
+                let got: Option<String> = sqlx::query_scalar(&sql)
+                    .fetch_one(pool)
+                    .await
+                    .with_context(|| format!("ord_term identity query: {sql}"))?;
+                anyhow::ensure!(
+                    got.as_deref() == Some(op),
+                    "ord_term identity on {domain}: extractor returned {got:?}, payload op={op:?}",
+                );
+            }
+            Some(other) => anyhow::bail!(
```
tobyhede, line 544: **Should fix — this catch-all discards the compile-time exhaustiveness check**

`Term` (`crates/eql-domains/src/lib.rs:81`) has exactly four variants and is **not** `#[non_exhaustive]`, so a downstream crate may match it exhaustively. A catch-all is only mandatory across a crate boundary for a `#[non_exhaustive]` enum.

So this arm converts a compile error into a runtime one. Its own message — "add an identity branch for it" — is exactly the advice `rustc` would give at the point the fifth term lands, for free. As written, the advice arrives only when a test happens to run against a domain carrying that term.

This compiles and restores the guard:

```rust
Some(Term::Hm) | Some(Term::Bloom) => unreachable!(
    "ordering_term() cannot return a non-ordering term"
),
```

Relatedly: `"hm"` (line 469), `"ob"` (487), and `"op"` (524) are hardcoded, but `Term::json_key()` is a `pub const fn` returning exactly those. It's the same argument this PR makes for catalog-deriving `ord_extractor_expr` — worth applying one function over.

tests/sqlx/tests/encrypted_domain/float_special.rs
```diff
@@ -11,15 +11,20 @@
...
+//! the flip happened when the `_ord` default moved from block-ORE to CLLW-OPE.
+//! Both behaviours are pinned below, one test per domain.
 
 use anyhow::Result;
 use eql_tests::fixtures::cipherstash::encrypt_store;
 use eql_tests::fixtures::index_kind::IndexKind;
+use eql_tests::known_failure;
+use eql_tests::known_failure::ISSUE_FLOAT_SIGNED_ZERO_EQ;
 use eql_tests::property::{connect_pool, ensure_eql_installed};
 use eql_tests::scalar_domains::F8;
```
tobyhede, line 32: **Should fix — this suite covers `F8` only; `real` / `F4` signed-zero is asserted nowhere**

`F4` appears nowhere in this file, and every helper hard-codes `public.eql_v3_double*`. But `crates/eql-domains/src/lib.rs:429` says it plainly:

> Both float widths encrypt through the SAME f64 crypto path (`Plaintext::Float`), so `real` vs `double` is purely a Postgres-surface distinction.

So the new `-0.0 < +0.0` behaviour applies to `real_ord` identically, and both U-005 and `.changeset/ord-defaults-to-ope.md` promise it for `real`. The matrix can't pick up the slack: `crates/eql-domains/src/fixtures/record.rs:111` excludes `-0.0` from both `REAL_FIXTURES` and `DOUBLE_FIXTURES`, and `tests/sqlx/src/scalar_domains.rs:1116` actively asserts the float fixture set contains none.

This is a different axis from your `=`-facet comment further down, and that one is the sharper finding. But fixing it without parameterising this suite over `F4` still leaves `real_ord` uncovered on **both** facets — `<` and `=`.

A `F4`/`F8` type parameter over `encrypt_specials` + the three domain names would close it.

tests/sqlx/tests/encrypted_domain/float_special.rs
```diff
@@ -114,17 +133,83 @@ async fn two_encryptions_of_same_nan_bits_compare_equal() -> Result<()> {
...
+    let assertion = if equal {
+        Ok(())
+    } else {
+        Err(anyhow::anyhow!(
+            "encrypted `=` on public.double_eq returned false for -0.0 vs +0.0"
+        ))
+    };
+    known_failure(
+        ISSUE_FLOAT_SIGNED_ZERO_EQ,
+        "-0.0 == +0.0 under public.double_eq",
```
tobyhede: **Nit — this names a domain that doesn't exist, and prints on every CI run**

`eq_cmp` queries `public.eql_v3_double_eq`. This `what` string, and the error literal at line 203, both say `public.double_eq` — the pre-CIP-3472 name.

By design `known_failure()` is never a silent skip: it `eprintln!`s the `what` on every run while the bug reproduces. So this is the most-read stale name in the diff. Same literal in the doc example at `tests/sqlx/src/known_failure.rs:62`.

tobyhede (commented):
## Second pass: 9 items not covered by the existing threads

I verified each of these against `7fc9f5ce` before filing (adversarial check, default-to-refuted). @coderdan's review already covers the `splinter.sh` rows, the `known-failures.sh` pipefail branch, the `scalar_domains.rs` doc inversion, the `fixture_oracle.rs` comment inversions, `ope_cllw/types.sql`, the `_ord`/`_ord_ope` twin, the U-note cross-links, `v3_privilege_tests`, `mise.toml:400`, and the changeset bump — I agree with all of them and won't restate. Seven inline comments below, plus three items that anchor off-diff.

One correction I owe: the `ore_cllw` count. Field extraction on the diff gives **10 stale rows** (`ore_cllw_{eq,neq,lt,lte,gt,gte}`, `eql_v3.{ore_cllw,has_ore_cllw}`, `eql_v3_internal.{ore_cllw,has_ore_cllw}`) plus **2 live rows whose prose regressed** (`eql_v3.lt`, `eql_v3.->`). That's the twelve added lines; the "eight" in that thread's body is low.

---

### Should fix — no `search_ore_fn_oracle` (`fixture_oracle.rs:485`, off-diff)

`search_fn_oracle` covers `Variant::Search`. Grep for `SearchOre` across `fixture_oracle.rs` and `e2e_oracle.rs`: zero hits.

The asymmetry is sharper than it looks, because the `Ord`/`OrdOre` pair **is** symmetric — `ord_fn_oracle` iterates `for variant in [Variant::Ord, Variant::OrdOre]` (`fixture_oracle.rs:471`), and `e2e_oracle.rs:113-114` covers both. `text_search_ore` is the only ORE twin in the suite without an all-pairs oracle.

What it gets instead is genuinely weaker, not a variant of the same thing: the 85 `search_ore` tests in `snapshots/matrix_tests_text.txt:283-367` are all `..._pivot_{min,mid,max}_correctness` / `_cross_shape`, generated by `matrix.rs:859`, which binds one `let pivot: $scalar = $pivot_val;` and compares `col op pivot` across the fixture. Pivots are only `min_pivot()`/`max_pivot()`/`mid_pivot()` (`matrix.rs:195-198`). The suite README states the contrast directly: the fixture oracle "checks **every ordered pair**", whereas "the fixed-pivot scalar matrix only checks hand-picked values."

So this PR created a domain and gave it three anchor values where its twin gets O(n²). `search_fn_oracle` is pre-existing (the whole base→HEAD diff of this file is comment-only); the catalog and matrix gained `SearchOre`, the oracle didn't.

### Should fix — U-005 argues the default flip entirely on indexability, never on leakage (`docs/upgrading/v3.0.md`)

`docs/upgrading/v3.0.md` contains zero instances of *leak*, *security*, or *threat*. Neither changeset mentions leakage. Every argument — `bytea_ops`, superuser, silent index non-engagement — is correct and necessary, and none of it is about what the ciphertext reveals.

But this flips the SEM under every existing `_ord` column. The PR's own text observes that the OPE term for `""` is "a proper prefix of every non-empty term" — i.e. `op` terms for text are prefix-structured, so a btree over `ord_term(col)` places ciphertexts sharing a plaintext prefix adjacent, and the shared byte-prefix length is readable off the index. Block-ORE leaks the index of the first differing block.

Those may well be equivalent for CLLW. I'm **not** asserting a downgrade — I'm asserting that a reader of U-005 currently has no way to tell, and for a searchable-encryption product the default's leakage profile is the load-bearing property. One paragraph comparing the two, in U-005, before this becomes the default.

### Nit, drive-by (not this PR's doing) — stale index-kind comment (`fixture_oracle.rs:449`, off-diff)

> `// [Unique, Ore, Match], so its payload carries hm+ob+bf and casts cleanly to`

The real spec is `scalar_fixture.rs:120`: `[Unique, Ore, Match, Ope]` — as that file's own doc at `:33-35` correctly says. To be clear about provenance: base `4d1f19c2` already carries `Ope` there, and this comment line isn't a `+`/`-` line in this PR. It was **already stale before this PR** (from the earlier CLLW-OPE work). Worth fixing while the file is open; not a regression.

tasks/test/splinter.sh
tobyhede: Agreed, and the rows are inert **structurally**, which suggests a second fix alongside the revert.

The awk split at `splinter.sh:157-174` loads the allowlist into `allow[key]`, then for each **finding** tests `if (key in allow)`. It never iterates `allow[]`. The only guards are "refuse to run on an empty allowlist" (`:149`) and "fail on unmatched findings" (`:262`). So no allowlist row can ever be reported as unused, and the PR's own verification claim — *"splinter (0 unmatched)"* — is true while carrying no information about allowlist health. A future `ore_cllw` arrives pre-waived exactly as you say, and nothing in CI notices in the meantime.

One `for (k in allow)` loop closes it.

The pointed part: this PR introduces `known-failures.sh` specifically to enforce *"registered but referenced by no test"* on `ISSUE_*` constants — and trips over the absence of that same invariant in the sibling gate, in the same diff. Worth fixing both, or neither.

Also worth settling the count in this thread, since the title and body disagree. Extracting field 3 from the added lines: **10 stale rows** (`ore_cllw_{eq,neq,lt,lte,gt,gte}`, `eql_v3.{ore_cllw,has_ore_cllw}`, `eql_v3_internal.{ore_cllw,has_ore_cllw}`) plus **2 live rows with regressed prose** (`eql_v3.lt`, `eql_v3.->`). Twelve added lines mentioning `ore_cllw`; ten of them rows.

mise.toml
tobyhede: Confirmed independently — and I think the fix wants to be structural, because the comment right above already predicted this.

`mise.toml:118-122` says: *"Adding a new `proptest-e2e`-gated module means adding a line here; no glob catches it for you."* The very next `proptest-e2e` module then did not get a line. A comment is not a gate, and the next gated module will land the same way.

`test:matrix:inventory` already solves the equivalent problem for scalar types — it discovers them from the test binary's `--list` and cross-checks against `cargo run -p eql-codegen -- list-types`, so a type missing its wiring fails CI. Same shape here: enumerate the `#[cfg(feature = "proptest-e2e")]` modules and fail when one is absent from the task, rather than adding a third filter line and a fourth comment.

PR #390, opened 2026-07-09
feat(jsonb)!: switch SteVec ordering from CLLW-ORE (oc) to CLLW-OPE (op)

freshtonic:
Closes [CIP-3469](https://linear.app/cipherstash/issue/CIP-3469/switch-jsonbstevec-support-from-cllw-ore-to-ope)

## What

SteVec (encrypted JSONB) entries now carry `hm` XOR **`op`** (CLLW-OPE) instead of `hm` XOR `oc` (CLLW-ORE), and `jsonb_entry` ordering extracts the new `eql_v3.ord_ope_term(entry public.jsonb_entry)` — the same `eql_v3_internal.ope_cllw` bytea domain the scalar `_ord_ope` surface uses. OPE ciphertext bytes order under **native byte comparison**, so the entire `src/v3/sem/ore_cllw/` surface is deleted: the composite type, the per-byte CLLW comparator, six hand-written operators, and the superuser-only `DEFAULT FOR TYPE` btree operator class.

## Why

`CREATE OPERATOR CLASS` requires superuser, so SteVec entry ordering was the **last EQL surface that could not index on cloud-hosted Supabase / managed Postgres** (the same motivation as CIP-3468's ORE-domain poisoning). With OPE the whole comparison chain is inlinable `LANGUAGE sql` and a plain functional btree index on `eql_v3.ord_ope_term(doc -> '<selector>'::text)` engages structurally on any install — no operator class at all.

## Changes

- **SQL**: `hm` XOR `op` in all three domain CHECKs/validators; `to_ste_vec_query` / `@>` needles / `eql_v3.jsonb_array` deterministic keys now `s`/`hm`/`op`; `lt`/`lte`/`gt`/`gte` and the `min`/`max` aggregate state functions compare via `ord_ope_term` (op-less entries keep the same skip semantics `oc`-less had); `eql_v3.ore_cllw(entry)` / `eql_v3.has_ore_cllw(entry)` removed (the extractor's SQL-NULL-on-absent covers the `has_` use).
- **eql-bindings**: `SteVecTerm::OpeCllw { op }` replaces `OreCllw { oc }`; the `OreCllw` newtype is gone. `from_v2` **rejects** `oc` sv entries with a new `UnconvertibleOreTerm` error — ORE bytes misorder under byte comparison and older Compat-mode OPE-in-`oc` is indistinguishable from the payload alone, so re-encryption is the only conversion. sv-level `op` is accepted ahead of the v2.3 schema file (same precedent as scalar `op`).
- **Fixtures**: SteVec generation pinned to `SteVecMode::Compat` (real OPE bytes from the pinned client) with a fail-closed `oc`→`op` remap shim in `v3_convert.rs`, to be removed once cipherstash-client emits sv-level `op` natively (cipherstash-suite counterpart PR).
- **Tests**: `ore_cllw_v3_opclass_tests.rs` deleted; jsonb suite migrated (ladders verified monotone under plain byte order); index-engagement test now exercises the default bytea opclass; snapshots regenerated.
- **Lint/tasks**: splinter allowlist and `pin_search_path_v3.sql` drop the ore_cllw rows (the new extractor is covered by the structural domain-arg skip).
- **Docs**: json-support / sql-support / eql-functions / database-indexes updated; upgrade note **U-004** in `docs/upgrading/v3.0.md` (renumbered from U-003 after rebasing over CIP-3468, whose note took that slot; its "SteVec is out of scope" caveat is now marked resolved by this change, closing CIP-3471); changeset (`major`, folds into the pending 3.0.0 pre-mode).

## Verification

- `codegen:parity`, `types:check`, `typescript:check`, `test:self_contained_v3` all pass; release installer has zero `ore_cllw` references.
- All eql-bindings test binaries pass (including new `UnconvertibleOreTerm` coverage); `cargo test --lib` in tests/sqlx passes; all DB test binaries compile.
- Cross-repo validation: protect-ffi's full lib suite (202 tests) passes against this branch's eql-bindings + the suite branch's client via local path patches.

## Coordinated changes

- cipherstash-suite: `james/cip-3469-ste-vec-emit-sv-level-op-term` (client emits sv-level `op`, `SteVecMode` defaults to Compat)
- protectjs-ffi: `james/cip-3469-stevec-ope` (fixture migration, release-gated)


freshtonic (commented):
## Independent review (fresh-context agent, given only the PR + CIP-3469)

**Verdict: APPROVED — no blocking findings.** Verified against a live PG 17: fresh install of the built release, `cargo test --workspace` (140 eql_tests lib + 100 eql-bindings unit tests), `v3_public_surface_tests` 7/7, `v3_operator_equivalents_tests` 3/3, `payload_schema_tests` 32/32, inlinability/validator-parity pins 38/38, and all gates (`codegen:parity`, `types:check`, `typescript:check`, `test:matrix:inventory`, `test:self_contained_v3`, `docs:validate` 241 files). Installer has 0 `ore_cllw` refs / 76 `ope_cllw`.

The 9 local `v3_jsonb_tests` failures + 20 `jsonb_entry` matrix failures are environmental: this workspace's gitignored pre-PR fixtures still carry `oc` terms, which the new domain CHECK **correctly rejects** (SQLSTATE 23514) — the PR's intended fail-closed behaviour. CI regenerates fixtures each run.

**Specifically checked and found sound:** NULL/STRICT semantics of the new extractor chain (live-tested); the op-less-seed aggregate guard (logic preserved verbatim, seed-position test retained); the `eq_term` coalesce collapse assumption (negative byte-collision test retained; hm=32 bytes vs op length ≢ 32 rationale documented); `from_v2` fail-closed ordering (`oc` check precedes term matching); REQUIRE-edge rewiring (tsort clean); blocker non-regression; breaking-change hygiene (major changeset + U-004 + compat table).

**Non-blocking findings** — three fixed in 9e3a7d5a (changelog-safe root-relative U-004 link in the changeset, blank line before U-003's Verification block, "an CLLW-OPE" grammar); one deferred to the shim-removal follow-up (the shared fixture preamble still says "HMAC + ORE block terms"; generalise when `ste_vec_oc_to_op` is removed — it's pinned by `spec.rs:370`).

**Highlights:** `UnconvertibleOreTerm` fail-closed conversion with the *why* at the point of failure; the fixture shim engineered to fail loudly if dropped prematurely; net −1503/+727 while keeping capability parity (`has_ore_cllw` → `ord_ope_term(...) IS NOT NULL`); U-003's "SteVec out of scope" caveat recorded as resolved (CIP-3471) rather than silently deleted.

PR #392, opened 2026-07-09
Eql v3

coderdan:
EQL 3.0 replaces the untyped `eql_v2_encrypted` composite column type with a family of **typed, per-capability encrypted domain types**, and removes the `eql_v2` schema entirely. The database no longer owns the encryption configuration model — the client (CipherStash Proxy / ProtectJS) does — and EQL ships as a single self-contained SQL surface plus first-class Rust and TypeScript packages generated from one catalog.

This is a major, breaking release. Nothing in `eql_v2` survives.

---

## The core change: typed columns instead of one opaque type

In v2 every encrypted column was `eql_v2_encrypted`, regardless of what it held or how it could be searched. Capability was a property of the *payload*, discovered at query time.

In v3 capability is a property of the **type**. A column is declared as the domain matching exactly what it supports, and the domain's `CHECK` enforces that the payload actually carries the required index terms — so a missing term fails on insert or cast, not later at query time.

```sql
CREATE TABLE users (
  age        public.eql_v3_integer_ord,   -- equality + ordering + MIN/MAX
  email      public.eql_v3_text_eq,       -- equality only
  bio        public.eql_v3_text_match,    -- bloom containment only
  is_admin   public.eql_v3_boolean,       -- encrypted at rest, not searchable
  profile    public.eql_v3_json           -- searchable encrypted JSONB
);
```

### Capability suffixes

| Suffix | Operators | Backing index term |
|---|---|---|
| *(none)* | — (storage only) | `c` |
| `_eq` | `=` `<>` | `hm` (HMAC-256) |
| `_ord` | `=` `<>` `<` `<=` `>` `>=`, `MIN`/`MAX` | `op` (CLLW-OPE) |
| `_ord_ope` | as `_ord` (byte-identical twin) | `op` |
| `_ord_ore` | as `_ord` | `ob` (block-ORE) |
| `_match` | `@>` `<@` | `bf` (bloom filter) |
| `_search` | all of the above | `hm` + `op` + `bf` |
| `_search_ore` | all of the above | `hm` + `ob` + `bf` |

On the integer-like families the order term is exact, so `_ord` / `_ord_ore` get equality from `op` / `ob` directly. On **text** the order term is not equality-lossless, so `text_ord` and `text_ord_ore` additionally carry `hm` and always route `=` / `<>` through the exact HMAC.

**51 column-type domains** across 10 scalar families — `smallint`, `integer`, `bigint`, `real`, `double`, `numeric`, `date`, `timestamp` (each: storage / `_eq` / `_ord` / `_ord_ope` / `_ord_ore`), `text` (adds `_match`, `_search`, `_search_ore`), and `boolean` (storage-only) — plus `eql_v3_json` and `eql_v3_jsonb_entry`.

`boolean` is deliberately storage-only: a two-value column has so little cardinality that even an HMAC equality index would leak the plaintext distribution.

Every operator also has a **callable function equivalent** (`eql_v3.eq`, `eql_v3.lt`, `eql_v3.contains`, …). Supabase/PostgREST exposes the database through RPC that calls functions, not operators — `WHERE col = $1` isn't expressible there, but `eql_v3.eq(col, $1)` is. A CI gate (`v3_operator_equivalents_tests.rs`) fails if any supported operator loses its wrapper.

## Schema layout

Three namespaces, split along a deliberate line:

- **`public`** — the column-type domains, prefixed `eql_v3_`. They live in `public` because dropping an EQL-owned schema must never drop application columns. The prefix means EQL types no longer shadow built-in names (`integer`, `text`, `json`), so an unqualified reference can't resolve to the wrong type through `search_path`, and future EQL versions can coexist in one database.
- **`eql_v3`** — the public API: query-operand domains, index-term extractors (`eq_term` / `ord_term` / `ord_term_ore` / `match_term`), comparison wrappers, aggregates, the JSONB access surface, `version()`, and `lints()`.
- **`eql_v3_internal`** — internal only: the SEM index-term types (`hmac_256`, `bloom_filter`, `ope_cllw`, `ore_block_256`), operator blockers, aggregate state functions, and CHECK validators. Hidden so they don't clutter type pickers that enumerate every type in every visible schema.

**EQL never grants permissions automatically.** The installer issues no `GRANT`/`REVOKE`; access is strictly opt-in. See [`docs/reference/permissions.md`](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/reference/permissions.md).

## Ordering: CLLW-OPE by default, block-ORE opt-in

The biggest operational change. `_ord` is now backed by **CLLW-OPE** (`op` term), not block-ORE.

`eql_v3.ord_term(col)` returns `eql_v3_internal.ope_cllw`, a domain over `bytea`. Because it's a bytea domain it inherits the **default** btree operator class, so a plain functional index engages structurally, and the whole comparison chain stays inlinable:

```sql
CREATE INDEX ON users (eql_v3.ord_term(age));
```

Block-ORE needs a hand-written operator class on a composite type, and `CREATE OPERATOR CLASS` requires superuser. On cloud-hosted Supabase and most managed Postgres that class silently didn't exist — `CREATE INDEX` *succeeded* by binding `record_ops`, then never engaged. Ordered queries fell back to seq scans with no error.

So v3 does two things:

1. **Ordering works without superuser by default.** `_ord` / `text_search` use OPE.
2. **Block-ORE fails loudly when unavailable.** If the installer can't create the ORE operator class, it capability-detects the skip and poisons all 20 ORE-carrying domains with an always-raising `CHECK`. First cast or insert raises `feature_not_supported` (`0A000`) naming the domain and pointing at the alternatives. The constraint is `NOT VALID`, so ORE data written under an earlier superuser install stays readable. Superuser installs are unchanged.

Block-ORE remains fully supported behind `_ord_ore` / `text_search_ore`, extracted via `eql_v3.ord_term_ore(col)`.

## Query-operand domains

Every term-bearing domain has a paired `eql_v3.query_<name>` twin (**40 of them**) carrying `{v, i, <terms>}` — the storage envelope **minus the ciphertext `c`**, enforced by the domain CHECK.

```sql
WHERE age = $1::eql_v3.query_integer_ord
WHERE profile @> $1::eql_v3.query_jsonb
```

Query operands travel through PostgREST query strings, URL logs, proxies, and SQL logs. They should carry index terms only — never a decryptable ciphertext. A `c`-bearing operand is rejected.

## Encrypted JSONB (SteVec)

`public.eql_v3_json` documents are searchable without decryption: containment (`@>`, `<@`), path access (`->`, `->>`, `jsonb_path_query`, `jsonb_array_*`), entry equality on extracted leaves, and entry-level ordered range.

Entries carry `hm` **XOR** `op`. Entry ordering extracts `eql_v3.ord_term(entry)` — the same bytea-domain path as the scalars, so ordered JSONB indexes also work without superuser. The old `oc` (CLLW-ORE) term, its composite type, comparator, operators, and operator class are **removed**; `oc`-bearing documents must be re-encrypted.

```sql
-- containment
CREATE INDEX ON users USING GIN (eql_v3.to_ste_vec_query(profile) jsonb_path_ops);
-- ordered leaf
CREATE INDEX ON users (eql_v3.ord_term(profile -> 'age'::text));
```

Comparisons are **leaf-level only**. Root-document `=`, `<`, `>` and every native `jsonb` operator reachable through domain fallback (`?`, `?|`, `@?`, `#>`, `||`, …) raise rather than silently falling through to plaintext-jsonb semantics.

> **Typed operands.** `eql_v3_json` is a `jsonb` domain, so selector literals must be typed: `col -> 'sel'::text`. A bare `col -> 'sel'` resolves to the native `jsonb` operator, because PostgreSQL reduces a domain to its base type for unknown-typed literals. The Proxy interface passes typed parameters.

## Wire format: `v: 3`

Every v3 domain CHECK pins `VALUE->>'v' = '3'`. The bindings accept exactly `3` and reject legacy `2` at the type boundary. A client emitting `v: 2` cannot insert into a v3 column.

`eql_bindings::from_v2` / `from_v2_query` convert v2.3 payloads to v3, failing closed (`MissingTerm`, `UnconvertibleOreTerm`) rather than guessing — the target domain is explicit input, since every v2 index term is optional on the wire and capability can't be inferred.

## One catalog, three languages

Scalar domains are defined as rows in `eql-domains::CATALOG` (Rust). `cargo run -p eql-codegen` materialises from it:

- the SQL domain/operator/aggregate surface,
- the Rust `eql-bindings` crate (payload structs, `DomainPayload` / `QueryPayload` enums, the domain inventory),
- TypeScript bindings and JSON Schemas → the **`@cipherstash/eql` npm package**.

Adding a scalar type is one catalog row. Drift between the three is CI-gated (`types:check`). The previous Python codegen toolchain is gone — building and testing EQL no longer needs Python.

Both language packages **bundle the exact SQL installer they were generated against** (`eql_bindings::sql`, or the npm package's `./sql` subpath export), so consumers pin wire types and DDL together. Releases are cut in lockstep from Changesets; from 3.0.0 the changelog is [`packages/eql/CHANGELOG.md`](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/packages/eql/CHANGELOG.md).

## Removed

The `eql_v2` schema and everything in it, with **no v3 replacement** for:

- `eql_v2_encrypted` and its operators (`=`, `<>`, `LIKE`/`ILIKE`, `@>`/`<@`, ORE comparisons)
- database-side configuration management — `eql_v2_configuration`, `add_search_config`, `add_column`, `migrate_config`, `diff_config`, `create_encrypted_columns`
- the `encryptindex` migration machinery
- `GROUP BY` / `grouped_value` on the encrypted column type
- operator-class-on-column indexing

Searchable-encryption capability is provided by the v3 domain families. Configuration is now owned by the client.

Build variants are gone too: **one artifact**, `release/cipherstash-encrypt.sql` (+ uninstaller), under the name the combined build used — existing install URLs keep working. The Supabase and Protect variants no longer exist because there is now a single surface that installs everywhere.

## Breaking changes

Full detail and migration recipes in [`docs/upgrading/v3.0.md`](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/upgrading/v3.0.md):

| | Change |
|---|---|
| [U-001](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/upgrading/v3.0.md#u-001-eql_v3-payloads-carry-v-3) | Payloads carry `v: 3` |
| [U-002](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/upgrading/v3.0.md#u-002-query-operand-domains-are-eql_v3query_name) | Query operands are `eql_v3.query_<name>` |
| [U-003](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/upgrading/v3.0.md#u-003-non-superuser-installs-disable-the-ore-backed-domains) | Non-superuser installs disable the ORE domains |
| [U-004](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/upgrading/v3.0.md#u-004-stevec-ordering-terms-are-cllw-ope-op) | SteVec ordering is CLLW-OPE (`op`, not `oc`) |
| [U-005](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/upgrading/v3.0.md#u-005-_ord-domains-are-backed-by-cllw-ope) | `_ord` is backed by CLLW-OPE — existing `_ord` indexes must be rebuilt |
| [U-006](https://github.com/cipherstash/encrypt-query-language/blob/eql_v3/docs/upgrading/v3.0.md#u-006-text_search-is-backed-by-cllw-ope) | `text_search` is backed by CLLW-OPE |

Two behaviour changes ride along with U-005: `text_ord` now **accepts** the empty string (OPE encodes `""` to a well-formed term; block-ORE produced an unorderable `ob: []`), and `real`/`double` `_ord` columns now distinguish `-0.0` from `+0.0`. A float column needing IEEE `±0.0` semantics should be typed `_ord_ore`, which canonicalizes. Both splits are pinned by `known_failure` markers against #387.

## Testing

18 SQLx suites, plus a property-test harness asserting SQL operator results agree with a plaintext oracle across a generated input space — a pure-Rust catalog suite, a fixture suite over committed real ciphertexts, and an e2e suite (behind the `proptest-e2e` feature) that batch-encrypts fresh plaintexts through ZeroKMS every run.

Dedicated gates cover the things that silently rot: `eql_v2`-freeness of the artifact and its dependency closure, public/internal schema placement, operator↔function equivalence, uninstall completeness, privilege behaviour, and Rust/TS/JSON-Schema drift.

Two performance regressions found during the release benches were fixed by converting hot functions the planner can never inline from `LANGUAGE sql` to plpgsql (#353, #354), restoring or beating v2 numbers on ORE ordered scans and the JSONB needle cast.

---

## Branch state

Brought up to date with `main` via merge (`6d7e87c9`), not rebase. The only conflict was `.github/workflows/release-plz.yml` — resolved in favour of `eql_v3`, which carries the full publish workflow; main's copy is the registration stub added in #383, whose own comment says to replace it when this branch lands. The merge changed no content: the resulting tree is byte-identical to the branch tip before it.

A rebase was attempted first and abandoned. `--rebase-merges` cannot reproduce the hand-made resolutions in 32 of this branch's 50 merge commits, and silently reverted part of the `eql_v3_internal` schema split across 166 commits.


freshtonic (commented):
## Review: EQL 3.0 (`eql_v3` → `main`)

Reviewed the **hand-written** surface only. Generated artefacts were excluded by marker (`AUTOMATICALLY GENERATED FILE` / `@generated`) and by path (`release/`, `src/v3/scalars/`, `crates/eql-bindings/{sql,bindings,schema}/`, `packages/eql/{src/generated,sql}/`, `pnpm-lock.yaml`) — 654 of 949 changed files, leaving ~295 to read. I also took the branch's prior per-PR reviews as given and did not re-litigate settled design; the focus is merge-level and release-level risk, plus invariant violations.

Worth flagging up front: **CodeRabbit skipped this PR entirely** ("876 files exceed the limit of 150"), so nothing automated has looked at it.

Overall this is high-quality work. The footgun invariants the architecture exists to protect are intact, and the test suite is unusually well-guarded. Three things need attention, one of them before `3.0.0` can be cut at all.

---

### 🔴 Blocking (before versioning, not necessarily before merge) — Changesets pre-mode reaches `main`, so the production path can never emit `3.0.0`

`.changeset/pre.json` (`mode: "pre"`, `tag: "alpha"`) exists on `eql_v3` and **not** on `origin/main`. This merge carries it onto `main`. `release.yml` classifies any push to `main` as `mode=production` and runs `pnpm run version` → `changeset version`, which in pre-mode bumps to the next *prerelease*.

`docs/development/releasing.md` documents this exact footgun:

> **Footgun — exit pre-mode before a final release.** `.changeset/pre.json` lives on `eql_v3`. If it reaches `main`, `changeset version` there will emit alpha-suffixed "final" versions. Run `changeset pre exit` (and merge that) before cutting a production release from `main`.

I confirmed it empirically rather than inferring it — ran `changeset version` in a detached worktree at this branch's head:

- **As-is (pre.json present):** `3.0.0-alpha.4` → **`3.0.0-alpha.5`**, all 43 changesets left unconsumed, no `3.0.0` CHANGELOG section. Changesets itself warns: *"You are in prerelease mode … you should revert these changes and run `changeset pre exit`."*
- **After `changeset pre exit`:** → **`3.0.0`**, all 43 changesets consumed, full `3.0.0` changelog section written.

The 43 changesets (6 `major` incl. `eql-v2-removed.md`, 30 `minor`, 7 `patch`) do aggregate correctly to `3.0.0` — but only once pre-mode is exited.

**Not auto-catastrophic:** publishing is gated behind the human-reviewed "Version Packages" PR, and the production path uses `changeset publish`, which in pre-mode tags npm `alpha` — so `latest` isn't poisoned. The failure mode is "3.0.0 never ships / another alpha publishes from the production path," not "GA is corrupted."

**Fix:** `changeset pre exit` and commit — either in this PR or immediately after merge, before the Version Packages PR is versioned.

---

### 🟠 Major — `docs/upgrading/v3.0.md` doesn't cover the `eql_v2` removal, and says so

The guide's own header:

> This guide currently covers the `eql_v3` envelope-version bump; the release owner extends it with notes for the other pending breaking changes (**notably the `eql_v2` schema removal**) as the release is prepared.

`grep -cE 'eql_v2_encrypted|config_add' docs/upgrading/v3.0.md` → **0**. U-001…U-006 and the compatibility table cover only intra-v3 changes (envelope `v:3`, query-domain renames, ORE→OPE) — i.e. migrations for *pre-release alpha adopters*, not for an actual EQL 2.x production user.

So a v2 user opening the canonical upgrade guide for the release whose headline change is "the entire `eql_v2` schema is removed" finds nothing about their `eql_v2_encrypted` columns, the removed `config_add_table`/`config_add_column`/`config_add_index` API, or the removed `eql_v2.*` operators.

**No data-loss trap** — the v3 installer globs `src/v3` only and leaves an existing `eql_v2` schema intact, and the uninstaller drops only `eql_v3`/`eql_v3_internal`. This is a documentation gap, not a destructive one. Defensible while on the `-alpha` line; must land before `3.0.0` GA.

---

### 🟠 Major — encrypted empty JSONB array (`sv: []`) breaks the array accessors

`eql_v3.ste_vec()` builds its `jsonb[]` with `array_agg(elem)`, which returns SQL `NULL` (not `'{}'`) over zero rows. The document CHECK **accepts** `sv: []` (`jsonb_typeof(val->'sv')='array'` holds, and `NOT EXISTS` over an empty array is trivially true), so this is a storable value.

Reproduced against a real PostgreSQL 16 with the surface built from this branch (`mise run build` → `release/cipherstash-encrypt.sql`), payload `{"v":"3","i":{...},"a":true,"sv":[]}`:

| Call | Actual | Expected (native jsonb parity) |
|---|---|---|
| cast to `public.eql_v3_json` | accepted (`t`) | accepted |
| `eql_v3.jsonb_array_length(doc)` | `NULL` | `0` |
| `eql_v3.jsonb_array_elements(doc)` | **`ERROR: upper bound of FOR loop cannot be null`** | 0 rows |
| `eql_v3.jsonb_array_elements_text(doc)` | **same ERROR** | 0 rows |

Both errors come from `FOR idx IN 1..array_length(sv, 1)` with `sv` NULL (`src/v3/jsonb/functions.sql:446`, and the same shape in `jsonb_array_elements_text`).

Note the obvious fix is **not** sufficient: `coalesce(array_agg(elem), '{}')` still yields `NULL` from `jsonb_array_length`, because `array_length('{}'::jsonb[], 1)` is itself `NULL` in Postgres. I verified a fix that does hold:

- `jsonb_array_length`: `RETURN coalesce(array_length(sv, 1), 0);`
- `jsonb_array_elements` / `_text`: `FOREACH item IN ARRAY coalesce(sv, '{}'::jsonb[]) LOOP`

→ length `0`, elements `0` rows, matching native `jsonb` semantics.

**Reachability caveat, stated honestly:** the SQL layer is definitely wrong for this input, but the trigger requires the client to emit `a: true` with `sv: []` for an encrypted `[]`. `a` is the documented array marker (`eql_v3_internal.is_ste_vec_array`), and `sv: []` documents already appear in `v3_jsonb_operator_surface_tests.rs:485` (without `a`), so it looks constructible — please confirm whether cipherstash-client produces it. Worth a regression test either way.

---

### 🟡 Minor — `edge_cases.rs` `_ord` arm tests a stale requirement

`tests/sqlx/tests/encrypted_domain/property/edge_cases.rs:154-164` — `check_rejects_payload_missing_ob` asserts "the `_ord` and `_ord_ore` domain CHECKs both require `ob`". Stale after the CLLW-OPE flip: the catalog gives `_ord` `Term::Ope` (`crates/eql-domains/src/lib.rs:215-216`), so its CHECK requires **`op`**, not `ob`. The `no_ob` fixture happens to omit both, so `_ord` still rejects it — for the wrong reason. A payload carrying `op` but no `ob` would be accepted by `_ord` and this test would never notice.

Not a can't-fail test (a fully-broken `_ord` CHECK is still caught), and the catalog-derived matrix check covers the real per-variant requirement. Suggest splitting into a missing-`op` case for `_ord` and missing-`ob` for `_ord_ore`, with corrected comments.

---

### Nits

- **`.gitignore:7-11,246-248`** — stale ignore entries for build variants deleted in this PR (`deps.txt`, `deps-supabase.txt`, `src/deps-protect.txt`, …). `tasks/build.sh` now only produces `deps-v3.txt` / `deps-ordered-v3.txt`. Harmless, cosmetic.
- **`.github/workflows/release.yml:17-20,48`** — the workflow holding `id-token: write` (npm OIDC) uses floating tags (`actions/checkout@v4`, `jdx/mise-action@v3`) while `test-eql.yml` SHA-pins. The highest-privilege workflow deserves the tightest pinning. Also, workflow-level `permissions` are inherited by jobs that don't need them (`classify` only runs `git log`); prefer `contents: read` at the top with per-job escalation.
- **`README.md:205`** — performance table still captioned "EQL 2.3 on PostgreSQL 17" on a 3.0 release. Honestly labelled, just stale provenance.
- **`docs/development/2026-07-04-release-tasks-implementation-plan.md`** (1538 lines) — superseded internal scratch. It carries a clear "SUPERSEDED (historical record)" banner and isn't linked anywhere, so this is defensible; consider deleting rather than shipping.

---

### Verified clean (checked, nothing found)

- **All four encrypted-domain footguns hold.** Programmatically audited all **1,565** generated blockers: every one is `LANGUAGE plpgsql`, none is `STRICT`. No domain-over-domain (all domains resolve to a base type). The single operator class is `FOR TYPE eql_v3_internal.ore_block_256`, a composite type, not a domain. Every `LANGUAGE sql` extractor/wrapper is `IMMUTABLE`, single-`SELECT`, and free of a `SET` clause, so inlining survives.
- **Self-containment.** No `eql_v2.<symbol>` anywhere in `src/v3` outside doc-comment prose.
- **Codegen determinism.** The only `HashSet`/`HashMap` in the generator is used for `contains()` membership, never iterated into output — so "identical CATALOG ⇒ byte-identical SQL" holds. `mise run build` on this branch produced a zero-byte diff.
- **Rust workspace.** `cargo check --workspace` clean; `eql-domains` (93), `eql-codegen`, `eql-bindings`, `eql-tests-macros` (22) test suites all pass. `serde(flatten)`/`deny_unknown_fields` are correctly kept off the same structs. `from_v2` fails closed on `oc` (CLLW-ORE) entries.
- **No silently-skipped tests.** The `proptest-e2e` gate is genuinely exercised — the dedicated `e2e` job runs `mise run test:sqlx:e2e`, which names each gated suite explicitly, and `ci-required` `needs` it.
- **No synthetic ciphertexts.** Hand-built hex payloads appear only in structural CHECK/blocker/comparator-wiring tests; every crypto ordering/equality claim rides real generated fixtures.
- **CI gating is sound.** `ci-required` lists every job in `needs` and passes only on `success|skipped` — a `failure`/`cancelled` cannot be masked. The drift gates (`codegen-parity.sh`, `schemas-parity.sh`) correctly fail on **untracked** files too, which is the usual blind spot.
- **v2 removal is coherent.** No live reference to any deleted task/artifact (`tasks/uninstall.sql`, `tasks/pin_search_path.sql`, `src/version.template`) — the only hits are removal-note comments. No doc instructs installing a build variant that no longer exists.
- **Uninstall is correct by design.** `eql_v3`/`eql_v3_internal` drop CASCADE takes the `public` ORE operators with them; the `public.eql_v3_*` column domains deliberately survive so application tables do, and reinstall is idempotent.

---

**Summary:** the `changeset pre exit` step must happen before `3.0.0` is versioned, and the v3.0 upgrade guide needs its `eql_v2`-removal section before GA. The empty-`sv` accessor bug is a real, reproduced defect worth fixing regardless of whether the client currently emits that payload. Everything else is nits. Nice work.

PR #398, opened 2026-07-14
feat(eql_v3): add storage-only public.eql_v3_json; rename JSON family for consistency

tobyhede:
## Summary

Makes the encrypted-JSON domains follow the same naming convention as every scalar family — the **bare family name is the storage-only domain**, and a suffix marks the searchable variant.

- **New:** `public.eql_v3_json` — a ciphertext-only, **storage-only** encrypted-JSON domain (`{v,i,c}` envelope, no index terms; the JSON analogue of `public.eql_v3_boolean`), generated by the scalar materializer like any storage-only scalar.
- **Renamed** (3.0.0-alpha surface): the searchable SteVec document `public.eql_v3_json` → `public.eql_v3_json_search`, `public.eql_v3_jsonb_entry` → `public.eql_v3_json_entry`, `eql_v3.query_jsonb` → `eql_v3.query_json`.

The `json` family is now a **mixed** family: three hand-written SteVec domains plus one generated scalar storage domain. `eql-codegen`'s SQL/bindings drivers iterate a new `families_with_scalar_domains()` seam and render each family's scalar domains only; the family stays non-scalar so `scalar_families()` (`list-types`, the matrix) still excludes it. The generated `Json` struct is re-exported from the hand-written `json.rs` so `inventory`/`payload` resolve `super::json::Json`.

Related: #397

## Naming

| Before (3.0.0 pre-release) | After | What it is |
| --- | --- | --- |
| `public.eql_v3_json` | `public.eql_v3_json_search` | searchable SteVec document |
| `public.eql_v3_jsonb_entry` | `public.eql_v3_json_entry` | one `sv` element |
| `eql_v3.query_jsonb` | `eql_v3.query_json` | containment needle |
| — (new) | `public.eql_v3_json` | ciphertext-only, storage-only |

## Changes

- **Catalog** (`crates/eql-domains`): family `jsonb`→`json`, `JSONB`/`JSONB_DOMAINS`→`JSON`/`JSON_DOMAINS`, document domain name `json`→`search`, new bare scalar storage domain; `families_with_scalar_domains()` helper; dropped the `full_name` `"json"` special-case (storage bare name now yields `eql_v3_json` naturally).
- **Codegen** (`crates/eql-codegen`): SQL + bindings drivers render scalar domains of any scalar-bearing family; `render_scalar_only_bindings(&JSON)` → `json_storage.rs`; `dump.rs` skips the scalar domain from the SteVec list; `query_payload_domains()` needle → `query_json`.
[...]

PR #406, opened 2026-07-15
feat(eql_v3)!: rename text bloom fuzzy match to `@@` / `eql_v3.matches`

tobyhede:
## What

Renames the text bloom fuzzy-match surface on the text match/search domains (`public.eql_v3_text_match`, `public.eql_v3_text_search`, `public.eql_v3_text_search_ore`) from the containment spelling `@>` / `<@` (backed by `eql_v3.contains` / `eql_v3.contained_by`) to a single directional operator **`@@`** backed by **`eql_v3.matches`**.

The old spelling promised containment semantics the operation does not provide: it is array containment over two bloom filters built from the downcased 3-gram token set — a probabilistic, order- and multiplicity-insensitive, one-sided match (a `true` may be a false positive; a `false` never is). `@>` / `<@` now **raise** "operator not supported" on these domains. Genuine JSON containment (`public.eql_v3_json_search`, `eql_v3.ste_vec_contains`) is unaffected and keeps the containment vocabulary.

This is the SQL-layer counterpart of the SDK's `contains()` → `matches()` rename.

## Text domains & SEM index terms

The `text` family's public domains and the `eql_v3_internal` SEM index-term types each carries (payload key in parentheses):

| Public domain | SEM index terms | Search capability |
| --- | --- | --- |
| `public.eql_v3_text` | — | storage only (encrypt at rest, no search) |
| `public.eql_v3_text_eq` | `hmac_256` (`hm`) | equality |
| `public.eql_v3_text_match` | `bloom_filter` (`bf`) | bloom fuzzy match (`@@`) |
| `public.eql_v3_text_ord` / `_ord_ope` | `hmac_256` (`hm`), `ope_cllw` (`op`) | equality + ordering (OPE) |
| `public.eql_v3_text_ord_ore` | `hmac_256` (`hm`), `ore_block_256` (`ob`) | equality + ordering (block-ORE) |
| `public.eql_v3_text_search` | `hmac_256` (`hm`), `ope_cllw` (`op`), `bloom_filter` (`bf`) | equality + ordering (OPE) + `@@` |
| `public.eql_v3_text_search_ore` | `hmac_256` (`hm`), `ore_block_256` (`ob`), `bloom_filter` (`bf`) | equality + ordering (block-ORE) + `@@` |

## Breaking change

- `WHERE col @> $1` / `eql_v3.contains(col, …)` on text match domains → **must** migrate to `WHERE col @@ $1` / `eql_v3.matches(col, …)`.
- The reverse `contained_by` direction is dropped.
- The functional GIN index recipe is unchanged (`GIN (eql_v3.match_term(col))`): `col @@ $1` inlines through `match_term` to the same array-containment the index supports.

See upgrade note **U-008** in `docs/upgrading/v3.0.md` and the changeset (`major`).

## Review fixes applied

The last two commits address review findings:
- corrected backwards rollback guidance in the v3.0 upgrade note;
- bumped the changeset `minor` → `major` (operator removal/rename breaks the public API per the versioning policy);
- replaced stale "bloom containment" wording with "bloom fuzzy match" across the docs;
- added `text_search_ore` to the sql-support matrices and the eql-functions `@@` section;
- retargeted the BloomFilter term doc and the unsupported-blocker Doxygen template, regenerating the committed bindings + SQL surface;
- regenerated the bundled package installers so packaged installs match the advertised API.

## Verification

- `cargo test -p eql-codegen` passes (incl. the `@brief`/`@param`/`@return` tag-parity assertions).
- `mise run codegen:parity` → `PARITY OK`; `mise run types:check` → bindings synced.
- Generated SQL diffs are comment-only — no function-body changes.

## Note for reviewers

Stacked on **#398** (`feat/eql-v3-json-storage-and-search-rename`); the diff here is only the text-match-operator rename. Base retargets to `main` automatically once #398 merges. (This replaces the earlier #405, which GitHub closed when the head branch was renamed.)


freshtonic (commented):
## Review — `@@` / `eql_v3.matches` text-match rename (CIP-3517)

Reviewed the **hand-written** surface only; per request I skipped the generated/bundled artifacts (`src/v3/scalars/**`, `crates/eql-bindings/{bindings,schema,sql}`, `packages/eql/{sql,src/generated}`, `*_expanded.rs`). That leaves the catalog/codegen, hand-written SEM SQL, docs, changeset, splinter allowlist, and tests (~28 files).

**Overall: strong, well-engineered, and internally consistent.** The design is the right shape and the diff is coherent end-to-end. One issue worth resolving before merge (the changeset bump), plus a couple of minor notes.

### What it does
Renames the text bloom fuzzy-match surface on `text_match` / `text_search` / `text_search_ore` from the containment spelling `@>`/`<@` (`eql_v3.contains`/`contained_by`) to a single directional `@@` backed by `eql_v3.matches`. `@>`/`<@` now raise "operator not supported" on those domains; genuine JSON containment is untouched.

### Design — verified correct
- `Term::Bloom` operators `["@>","<@"]` → `["@@"]`; `extractor_for_operator` routes `@@`→`match_term` and `@>`/`<@`→`None` (fall through to blockers). Clean single source of truth.
- `body_operator()` reduces `@@` to bloom array-containment `@>` on the extracted `smallint[]` terms, so `eql_v3.matches` inlines to the *identical* `match_term(a) @> match_term(b)` expression the old `contains` produced — the GIN index recipe (`GIN (eql_v3.match_term(col))`) is genuinely unchanged, and `bare_matches_operator_uses_functional_index` proves the bare `col @@ $1` form still engages it.
- The `blocker_only` flag on the `@@ (domain, jsonpath)` overload is a nice catch: `@@` is also native jsonb's jsonpath predicate, so that overload is forced to a blocker on every domain (incl. match domains), preserving the "no silent native jsonb ops" guarantee. Well-tested (`match_operator_carries_symmetric_and_blocker_only_signatures`).
- `match_metadata()` correctly drops commutator/negator (directional, no reverse) while keeping `contsel`/`contjoinsel`.

### Test coverage — excellent
- New `text_smoke::text_match_containment_operators_are_blocked` pins that `@>`/`<@` now **raise** — the key breaking behaviour.
- `text_match.rs` correctly drops the `<@`/`contained_by` tests and reduces the mixed-overload test to the 3 symmetric shapes that survive.
- `v3_operator_equivalents_tests`, `match_smoke`, `assert_match_smoke`, the match matrix cases, and the domains catalog tests are all updated coherently. Splinter allowlist swaps `contains`/`contained_by` → `matches` and refreshes the `match_term` justification.
- U-008 is thorough (compat table, who's-affected, migration, verification), and U-006 is de-"containment"-ed consistently.

### Issue — please resolve before merge

**The changeset bump is `patch`, but this is a breaking change.** `.changeset/rename-text-match-operator.md` declares `'@cipherstash/eql': patch`, which contradicts:
- the PR title `feat(eql_v3)!:` (the `!` marks a break),
- the PR body's own "Review fixes applied" note ("bumped the changeset `minor` → `major`"), which is **not** reflected in the committed file, and
- the versioning policy in `CLAUDE.md` — removing/renaming public operators & functions (`@>`/`<@`, `eql_v3.contains`/`contained_by` disappear from the public surface; +15/−30 in `eql_v3_public_surface.txt`) is called out as a **major** trigger ("operator names").

Shipping this as `patch` would let a public-API break land in a patch release. Per policy it should be `major` (or, if the intent is that this pre-GA surface is exempt, that decision should be made explicitly and the `feat!` marker / body reconciled). Right now the committed bump disagrees with everything else in the PR.

### Minor notes (non-blocking)
- `Operator::wrapper_function_name()` still maps `@>`→`"contains"` (asserted in a unit test), but `@>` is now a blocker on every generated domain, so that arm is never hit in generation — harmless, just dead-ish for the generated path (the JSON containment surface is hand-written).
- Directionality: dropping `<@` also drops the "is `col` contained by needle" form, and `$1 @@ col` (the `(jsonb, domain)` overload) won't hit the GIN index the way `col @@ $1` does (indexed expr must be the left `@>` operand). That's consistent with the stated "fuzzy match is directional" intent and the documented `col @@ $1` recipe is tested — noting only so it's a conscious choice.

Nice work — the `body_operator`/`blocker_only` decomposition keeps the whole thing catalog-driven and preserves the proven single-level inline to the GIN opclass. Just the changeset bump to sort out.

PR #410, opened 2026-07-16
feat(eql_v3): json_entry <-> query_<T> cross-type comparison operators

tobyhede:
Closes #404.

This PR now contains the final combined encrypted-JSON work from this branch and #413. It adds natural, indexable field-range comparisons while moving exact field equality to value-selector containment, and it introduces the new SteVec document envelope used by cipherstash-client 0.42.

## Final SQL model

Extracted entries are for field selection and range comparisons:

```sql
SELECT * FROM examples
WHERE encrypted_json -> 'age_selector'::text > $1::eql_v3.query_integer_ord;
```

The comparison can use a functional btree index on:

```sql
eql_v3.ord_term(encrypted_json -> 'age_selector'::text)
```

Exact field equality is expressed as encrypted-document containment:

```sql
SELECT * FROM examples
WHERE encrypted_json @> $1::eql_v3.query_json;
```

The query document contains a value-inclusive selector, so equality is based on the presence of the exact typed path/value token rather than equality of the lossy ordering term. This can use the functional GIN index on `eql_v3.to_ste_vec_query(encrypted_json)::jsonb`.

### Generated JSON-entry range surface

`eql-codegen` derives `json_entry <-> query_<T>_ord` and `_ord_ope` range operators from the scalar catalog for JSON-native numeric and text families. The operators work in both directions and have public function equivalents for environments that cannot invoke custom operators directly.

Unsupported or unsafe pairs have exact-signature blockers. They raise a clear error instead of silently falling through to PostgreSQL's native `jsonb` operators. This includes extracted-entry `=` / `<>`, temporal query operands, and text-search operands. JSON has no native date or timestamp leaf; temporal values in JSON use the text surface.

`eql_v3.ope_term(json_entry)` is the primary ordered-term extractor. The former `eq_term(json_entry)` spelling remains as a deprecated compatibility alias, avoiding a needless break while keeping ordering-specific behavior out of the general equality abstraction.

### Exact value selectors

Each scalar JSON value now emits a selector derived from its type, path, and canonical value. This makes equality injective for cases where ordering terms collide, including accented text and adjacent large integers. Extracted-entry equality is blocked for every family because a path entry alone does not carry the value selector.

Document containment delegates to the inlinable `ste_vec_contains` path, including the document-to-document overload. PostgreSQL can match the functional GIN index when the indexed document is compared with a stable query value. A genuinely column-to-column predicate still cannot use a conventional GIN lookup because its right-hand key varies for every candidate row.

### SteVec envelope and ciphertext binding

The encrypted document format is now:

```json
{"v":3,"k":"sv","i":"…","h":{},"sv":[]}
```

- key-retrieval material is stored once in the document-level `h` header;
- each entry stores only its AEAD ciphertext;
- each entry nonce is derived from its selector;
- ciphertext is therefore bound to its selector, and grafting ciphertext between entries fails authentication;
- value entries encrypt a fixed, versioned sentinel, keeping them distinct from a genuine empty-string value;
- the old per-entry `hm` equality term and independently decryptable ciphertext stream are removed.

The fixture document shrinks by roughly 30% compared with the repeated-header format.

### Validation and compatibility hardening

- SteVec document and query validators reject unknown or inconsistent root fields.
- Legacy conversion fails closed for SteVec documents that cannot be translated safely.
- The optional legacy key marker is validated when present.
- Bindings and schemas include the new `KeyHeader`, `EntryCiphertext`, document, entry, and query-entry shapes.
- Public documentation contains no private tracker identifiers.

## Breaking change and migration

Existing SteVec documents must be re-encrypted with a producer that emits the new envelope and value selectors. There is no safe mechanical conversion from the old document format. The changeset marks this as a major change, and the v3 upgrade guide documents the migration.

Range comparisons still use the existing ordered term and therefore retain its normalization and numeric-precision characteristics. Exact equality must use containment; it must not be reconstructed from `ope_term` equality.

## Performance and safety

- range predicates inline to the ordered-term expression used by the functional btree index;
- equality and containment inline to the expression used by the functional GIN index;
- unsupported comparisons fail loudly instead of producing silent native-JSONB false results;
- selector-derived nonces prevent same-document ciphertext reuse and substitution;
- value-inclusive selectors avoid equality false positives from ordering-term collisions.

The test PostgreSQL configuration also includes sufficient lock-table capacity for concurrent full-surface uninstall tests, together with a regression test and user-facing upgrade guidance.

## Verification

The final branch has passed:

- exact-version SQL regeneration and release-manifest hash parity;
- encrypted-domain codegen and schema parity;
- Rust compile, clippy, unit tests, and package checks;
- TypeScript and JSON Schema validation;
- SQL documentation, public-identifier, and anchor gates;
- clean PostgreSQL 17 installation and self-containment checks;
- Supabase splinter;
- fresh-encryption end-to-end property tests;
- catalog/matrix coverage and all four PostgreSQL 17 test shards.

All required CI checks are green after merging current `main`; the conflict-resolution commit is GPG verified.

## Related work

- #413 supplied the value-selector and envelope changes and is included here.
- #412 is superseded by the final combined implementation.
- cipherstash-client 0.42 is the corresponding producer/consumer implementation; protect-ffi is being updated to consume that client surface.


freshtonic (commented):
## Review — `json_entry` ↔ `query_<T>` cross-type operators (CIP-3526)

Reviewed the generator (`generate.rs`), a spot-check of the generated SQL, `context.rs::wrapper_entry`, the integration + generator-unit tests, the negative guard, the harness edits, the changeset, docs, and the public-surface snapshot.

**Overall: excellent — merge-ready.** This is a clean, purely-additive feature with an elegant catalog-driven design and unusually thorough tests/docs. One thing to confirm (the changeset bump), and the inherent field-context caveat is worth stating explicitly — both below.

### What it does
Binds the fixed `public.eql_v3_json_entry` leaf to every per-type query operand carrying `Term::Ope` (`query_<T>_ord`, `query_<T>_ord_ope`, and text's `query_text_search`), so a JSONPath selector-with-constraint (`col -> '$.age' > 21`, `col -> '$.email' = x`) is expressible in operator form and engages a functional btree on `eql_v3.ord_term(...)` instead of falling back to a seq-scan function form.

### Design — verified correct
- **Everything derives from `Term::Ope`.** Domain selection (`d.terms.contains(&Ope)`), operator set (`operators_for_terms(&[Ope])`, *not* the operand's own terms — which correctly keeps text's `search`=`[Hm,Ope,Bloom]` from dragging in Bloom's `@@`), extractor (`Ope.extractor()`), and the `json_entry` type name (derived from `eql_domains::JSON` via `full_name`/`domain_name`, not a literal — guards against a `#398`-style catalog rename). This is the right axis: `json_entry` serves exactly one term, so capability-driven selection makes the excluded cases (`_eq`, `_ord_ore`, `_match`, `_search_ore`) unrepresentable rather than asserted.
- **The forced `ord_term` extractor is the key subtlety and it's right.** On a dual-term `[Hm,Ope]` operand, the generic `extractor_for_operator` would route `=`/`<>` through `eq_term` (per-value HMAC) — which a SteVec scalar leaf never carries, yielding an operator that silently matches nothing. Forcing `ord_term` routes equality through the deterministic `op` term (equal plaintext at a fixed selector ⇒ equal `op` bytes). The `_eq` exclusion + the `no query_<T>_eq bound` guard lock this in.
- Spot-checked generated `json_entry_integer_functions.sql`: `eql_v3.eq(json_entry, query_integer_ord)` → `ord_term(a) = ord_term(b)`, both directions, inlinable (`LANGUAGE sql IMMUTABLE STRICT`, no `SET search_path`), no `_eq` operand, no jsonb cast. Confirmed `wrapper_entry`'s `dom` arg is inert here (only feeds the jsonb→domain cast, which never fires since neither operand is jsonb).

### Tests — outstanding
- Integration suite over real fixture ciphertext: structural (public wrappers, no `_eq` binding), index engagement (all five order/eq ops on `_ord` + `_ord_ope`), operator≡function, and — the crux — **equality correctness asserted as the full match set against a plaintext oracle, positive and negative, with field-context operands** (`json_entry_eq_cross_type_matches_plaintext_equality` for integer, and the text string-leaf variant, which even "guards the guard" that the operand carries non-NULL terms so the assertions can't pass vacuously).
- Generator unit tests pin the catalog derivation, the term-derived op set (`@@` excluded), and specifically that **text** routes `=` through `ord_term` (integer alone wouldn't catch a regression there).
- The negative guard in `v3_jsonb_operator_surface_tests.rs` is correctly *tightened*, not weakened — it now permits the `query_*` partner while still forbidding a raw-`jsonb` partner, plus a positive-presence test and a `_eq`-never-bound lock.
- Harness edits (`property.rs`, `inlinability.rs` type aliases; `ope/support.rs` `crate::`→`$crate::`) are legit clippy/macro-hygiene cleanups, no coverage change.

### To confirm — changeset bump
The changeset is `'@cipherstash/eql': patch`, but this is **additive public API** (new operators/wrappers; snapshot is +228/−0, "nothing existing changes"). Per the `CLAUDE.md` versioning policy, additive changes are **minor**. `patch` is only defensible if you're intentionally landing the feature in the pre-GA `3.0.1` line rather than cutting a `3.1.0` — a reasonable release-management call, but worth a conscious confirmation (this is under-classification, not the dangerous kind — unlike a breaking change marked patch).

### Operational caveat (well-handled, flagging for visibility)
The load-bearing runtime assumption is that the operand is encrypted in the *exact selector's field context* — an operand for a plain scalar or a different selector silently returns zero rows, and SQL can't detect it. This is inherent to term-comparison encryption, correctly pushed to the client/Proxy, and clearly documented (the ⚠️ block in `json-support.md` and the changeset). Also nicely documented: `query_text_ord`'s CHECK requires an `hm` key that the comparison doesn't use (because the same operand type also serves scalar `text` columns).

Nice work — the `Term::Ope`-as-single-source design and the op-vs-hm equality reasoning are the standout parts.

freshtonic (approved):
Approving. Full review is in my earlier comment — the design (everything derived from `Term::Ope`, the forced `ord_term` extractor so equality runs on the deterministic `op` term, catalog-derived type name) is correct and the tests are excellent (real-ciphertext equality correctness as the full match set against a plaintext oracle, positive and negative; negative guard tightened not weakened; generator unit tests pinning the text `=` → `ord_term` routing). Purely additive — public surface +228/−0.

The one item I raised — the `patch` changeset for an additive public-API change — is a **conscious, accepted call**: this surface is pre-GA with no consumers yet, so landing it in the 3.0.1 line rather than cutting a minor is fine. No blockers.

crates/eql-codegen/src/generate.rs
```diff
@@ -1470,4 +1729,202 @@ mod tests {
...
+            .map(|d| d.name)
+            .collect();
+        assert_eq!(text, vec!["ord", "ord_ope", "search"]);
+        // Excluded BY CONSTRUCTION, no name list needed:
+        //   `_eq`        [Hm]              — no Ope; a SteVec scalar leaf has no
+        //                                    per-value `hm`, so it'd be dead surface.
+        //   `_ord_ore`   [Hm, Ore]         — block-ORE; json_entry cannot produce it.
+        //   `_search_ore`[Hm, Ore, Bloom]  — likewise.
+        //   `_match`     [Bloom]           — Bloom only.
+        for excluded in ["eq", "ord_ore", "match", "search_ore"] {
```
coderdan: There's no match/bloom capability for SteVec

src/v3/scalars/integer/json_entry_integer_functions.sql
```diff
@@ -0,0 +1,202 @@
...
+-- REQUIRE: src/v3/scalars/integer/query_integer_ord_ope_functions.sql
+
+--! @file encrypted_domain/integer/json_entry_integer_functions.sql
+--! @brief Functions for public.eql_v3_json_entry.
+
+--! @brief Operator wrapper for public.eql_v3_json_entry.
+--! @param a public.eql_v3_json_entry
+--! @param b eql_v3.query_integer_ord
+--! @return boolean
+CREATE FUNCTION eql_v3.eq(a public.eql_v3_json_entry, b eql_v3.query_integer_ord)
```
coderdan: Is there an `eql_v3.query_integer_eq` type? Given OPE is always used for equality it seems redundant but semantically it would be easier to reason about (even if the internal type is still an `op`.

coderdan (approved):
LGTM. A small suggestion to the API. Would also be good to have some E2E tests in place for the `x->'y' > z` form queries.

freshtonic (commented):
## Overview

Binds the fixed `public.eql_v3_json_entry` leaf to every query operand carrying `Term::Ope`, giving `col -> '$.age' > $1::eql_v3.query_integer_ord` natural operator form plus `ord_term` btree engagement. 456 new objects (24 operators + 24 wrappers per ordered family; 36+36 for `text`).

The codegen is the strongest part of this PR. Deriving *everything* from `JSON_ENTRY_TERM = Term::Ope` — domain selection, operator set, extractor, and the `json_entry` type name via `Domain::full_name` — makes `_eq`/`_ord_ore`/`_match` exclusion structural rather than asserted, and the doc comment explaining why `Term::operators_for_terms(&[Ope])` is used instead of the operand's own term list (text's `search` would drag in Bloom's `@@`) is exactly the reasoning that should be captured in code. Deriving the type name rather than spelling it (#398 `jsonb`→`json`) is a real save. COMMUTATOR/NEGATOR resolution is correct for the cross-type case (name + reversed types for commutator, name + same types for negator), and both directions are emitted, so every reference resolves.

The `query_<T>_eq` exclusion analysis, traced to `ste_plaintext_term.rs` and pinned by a negative surface guard, is excellent work — that's a bug avoided, not a bug fixed.

Findings below, roughly in order of how much they'd change my mind about merging.

---

### 1. All three `text` operands require an `hm` the JSON-leaf crypto never emits

This is filed in the PR body as a cosmetic "known limitation" — *"callers supply a value the comparison ignores"* — but I think it's sharper than that, and it lands on the case the changeset and `json-support.md` both lead with (`-> 'email_selector' = $1::eql_v3.query_text_ord`).

The catalog asymmetry drives it:

| operand | terms | CHECK requires |
|---|---|---|
| `query_integer_ord` / `_ord_ope` | `[Ope]` | `v, i, op` |
| `query_text_ord` / `_ord_ope` | `[Hm, Ope]` | `v, i, hm, op` |
| `query_text_search` | `[Hm, Ope, Bloom]` | `v, i, hm, op, bf` |

By this PR's own crypto analysis, a SteVec string leaf yields `op` and nothing else. So for `text` a caller has three options, and none is a plain encryption:

- encrypt in field context → `{op}` → **fails the domain CHECK** (hard error);
- encrypt as a scalar text column → `{hm, op}` → CHECK passes, but `op` is column-context → **silent zero rows**;
- mint a composite: field-context `op` + an `hm` from a separate scalar-path encryption → works.

The numeric families are unaffected — `query_integer_ord` needs only `op`, so it's constructible directly.

The test is the evidence here: it can't build the text operand the way a client would, and instead scavenges an `hm` from an unrelated `hm`-carrying entry of the same document — material a query-side client, which has a plaintext value and no document, doesn't have. The commit retitling that splice from fabricated to real fixture bytes fixed the "real ciphertext" rule but not the constructibility question underneath it.

**So: does the stack adapter (cipherstash/stack#623) actually mint a `query_text_ord` operand for a JSON leaf, and where does its `hm` come from?** If it composes the two paths, this is fine and I'd just ask that the docs say *that* — "the client composes a field-context `op` with a scalar-path `hm`; the `hm` is inert" — rather than the current framing, which reads as though the `hm` is incidental. If it doesn't, the text examples in `json-support.md` and the changeset don't work as written and that half of the feature isn't shippable yet.

Relatedly, `query_text_search` demands an inert `bf` *and* an inert `hm` for zero benefit over `query_text_ord`. The PR rigorously excludes `_eq` as surface that "could never match real data"; binding `search` is the mirror image — it matches, but its CHECK demands two terms the seam ignores and the caller must manufacture. Term-derivation says bind it; usability says it's a trap. Worth a deliberate decision either way.

### 2. The shipped uninstaller now exceeds the stock lock budget — fixed only in the test harness

The `docker-compose.yml` comment is a genuinely good piece of diagnostic work, and the measurement in it is doing more than the PR acts on:

> the uninstaller takes 6,433 locks ... against the 6,400 the default (64) affords for the WHOLE cluster ... A lone teardown survives only because Postgres grows the table into spare shared memory

This PR adds 456 objects, so the uninstall goes from ~5,977 locks to 6,433 — **this change is precisely what crosses the stock 6,400 default.** And `tasks/uninstall-v3.sql` is concatenated verbatim into `release/cipherstash-encrypt-uninstall.sql`, so that single-transaction `DROP SCHEMA eql_v3 CASCADE` is what users run.

The bump landed in `tests/docker-compose.yml`. Users get nothing: no upgrade note, no changeset line, no note beside the uninstall instructions. The failure mode on a busy production DB with no spare shared memory is `out of shared memory / You might need to increase max_locks_per_transaction` — and the remedy needs a **restart**, which is not a small ask.

At minimum this warrants a user-facing note. Better would be splitting the uninstaller into per-schema (or batched) transactions so it stays under the default. Either way it shouldn't ship silently — the test harness bump documents the problem for us and hides it from users.

### 3. No range-ordering oracle for the headline feature

Test coverage on equality is genuinely strong — #4 and #5 assert the full match set against the plaintext oracle, positive and negative, and the "guard the guard" NULL check on the text operand shows real care about vacuous passes.

But `col -> '$.age' > 21` is the motivating example, and nothing asserts `>` returns the right rows:

- #2 proves the index engages — a wrong result set engages it just as well;
- #3 proves operator ≡ function — both can be identically wrong;
- the existing `jsonb_entry_integer_ord_ope_injectivity` proves distinct plaintexts ⇒ distinct terms — injectivity is not ordering.

`entry_x` already carries `plaintext`, and the operand already comes from row 1, so this is a handful of lines mirroring #4:

```rust
let expected_gt: Vec<i64> = sqlx::query_scalar(
    "SELECT id FROM entry_x WHERE plaintext > (SELECT plaintext FROM entry_x WHERE id = 1) ORDER BY id",
).fetch_all(&mut *tx).await?;
// assert `value > operand` matches expected_gt
```

I'd expect this to pass — both sides go through the same `ope_cllw` comparison the scalar OPE suite covers — so I'm flagging a gap, not predicting a failure. But it's the one assertion that directly states the feature's promise, and it's cheap. Ideally pick a pivot mid-range so `>` and `<` are both non-empty.

### 4. Unrelated commits bundled, and the PR body doesn't mention them

Three commits are outside CIP-3526: `test(ci): gate doc anchors and the known-failure parser` (new `doc-anchors.sh`, `known-failures-parser.test.sh`, two mise tasks, a new required CI job), and `style(rust): clean rustfmt and clippy across the workspace` (`property.rs`, `inlinability.rs`, `ope/support.rs`). They're fine changes — the doc-anchors gate's "deliberately NOT relevance-gated" reasoning is right — but they're a separate concern.

"Nothing existing changes" and the Verification list don't mention any of them, and with **squash-merge recommended** they'd all land under a `feat(eql_v3):` commit, making the CI gates unbisectable and invisible in history. I'd split them out, or at least disclose them in the body. (The `max_locks` bump is different — that one is *caused* by this PR and belongs here, modulo finding #2.)

### 5. Minor: `SEL_HELLO_OP` is duplicated rather than shared

`v3_json_entry_cross_type_tests.rs:285` hardcodes the selector with a comment pointing at `v3_jsonb_tests::SEL_HELLO_OP` — same hex, third copy (it's also inline in `v3_convert.rs:387`). The same file imports `SELECTOR` from `eql_tests::fixtures::v3_doc_integer`, so the pattern for sharing already exists and is used one line away. Given pinned SteVec selectors are a known regen-drift point, a fixture regen currently needs three synchronized edits, and separate test binaries mean the compiler won't catch a missed one. Worth hoisting to `eql_tests::fixtures::v3_ste_vec`.

---

## Assessment

The engineering here is high quality and the reasoning is unusually well documented in-code. Findings #3–#5 are straightforwardly addressable. #2 needs a decision (doc note vs. batched uninstall) before this ships. #1 is the one I'd want answered first, because it determines whether the `text` examples in the docs and changeset are accurate — and if they aren't, that's a correctness problem in user-facing documentation rather than a limitation footnote.

freshtonic (changes requested):
## Re-review — requesting changes

Since my last pass, `fa9a7f27` (text `=` returned false positives) and `44fb90b6` (bigint/numeric f64 rounding) are the two best commits in the PR. Tracing `"café" == "cafe"` to `cllw-ore`'s `orderize_string`, and `2^53 == 2^53+1` to `as_f64()` running *before* `orderable_to_u64`, then correcting the `JSON_ENTRY_TERM` doc that had asserted the opposite ("determinism gives ordering; equality also needs injectivity") — that's finding your own P1s and killing them. My earlier finding #1 asked a weaker question than the one you answered. The blockers are also implemented exactly right: `eql_v3_internal`, `LANGUAGE plpgsql`, non-`STRICT`, with `CREATE OPERATOR` still pointing at them, so `jsonb = jsonb` domain fallback can't leak through.

Two things stop me approving. The first is mechanical. The second is the same class of bug those two commits just fixed, in the half of the gate they didn't revisit.

---

### 1. Blocking (mechanical): the golden snapshot is stale — CI cannot go green

`tests/sqlx/snapshots/eql_v3_public_surface.txt` was last regenerated at `82153947`; both blocking commits postdate it and neither touched it. `eql_v3_public_surface_matches_golden` (`tests/sqlx/tests/v3_public_surface_tests.rs:152`) is an exact compare (`if actual != GOLDEN`) over `WHERE n.nspname = 'eql_v3'` only, so every function you moved to `eql_v3_internal` has to disappear from it:

- **28 phantom entries** — `eql_v3.eq`/`eql_v3.neq` for `json_entry` × `query_{text,bigint,numeric}_{ord,ord_ope}` (e.g. `:230`, `:116`). These are now `eql_v3_internal.eq` (`src/v3/scalars/bigint/json_entry_bigint_functions.sql:1`), invisible to the snapshot query.
- **12 entries for a binding that doesn't exist** — `json_entry` × `query_text_search` across all six operators (`:121`, `:232`, …). `git grep query_text_search -- src/v3 | grep json_entry` returns **zero**. The docs (`json-support.md:145`) correctly say `query_text_search` is "not bound at all"; the snapshot disagrees.

`EQL_UPDATE_SNAPSHOTS=1 mise run test:surface:snapshot:regen`. The relevant CI jobs are still `pending`, so this hasn't surfaced yet — but it's deterministic.

### 2. Blocking (design): the equality gate asks about the leaf, never about the operand

`json_leaf_equality_is_exact` (`crates/eql-domains/src/fixtures/kind.rs:237`) is a well-reasoned predicate, and its doc is explicit about its own scope — the operand side is "an EQL client contract … not something this predicate speaks to." That boundary is where three of the six kept families fail. An operator is only real if *both* sides can produce comparable bytes, and in `cipherstash-client 0.38.1` they're different functions of the value:

| path | encoding | source |
|---|---|---|
| stored **leaf** (`Value::Number`) | `orderable_to_u64(f64.to_orderable_bytes())` | `ste_plaintext_term.rs:212` |
| operand `Plaintext::Float` | `orderable_to_u64(f64.to_orderable_bytes())` | `:169` — **matches** |
| operand `Plaintext::Int` | `*n as u64` — raw cast, never orderable-encoded | `:163` — **mismatch** |
| operand `Plaintext::SmallInt` | `orderable_to_u64(i16.to_orderable_bytes())` | `:176` — **mismatch** |
| operand `Plaintext::NaiveDate` / `Timestamp` | `Err(invalid_type(..))` | `:190-191` — **cannot construct** |

`pad_orderable_to_8` left-pads into the low bytes, so `SmallInt(2)` lands near `0x8002` while the leaf `2` lands near `0xC000…` — not a near-miss, and it breaks *ordering* too, not just `=`.

So of the six families that keep `=`:

- **`date` / `timestamp`** — `TryFrom<&Plaintext> for OrderableTerm` returns `Err`. No client can build the operand via the documented `QueryOp::SteVecTerm` path, so **all six operators** are unreachable, not just `=`. The row `| date, timestamp | = <> < <= > >= | ISO-8601 / RFC3339 passes through collation unchanged |` (`json-support.md:138`) is true about the leaf and moot in practice. There is no e2e arm because one can't be written — `ste_vec_query_term`'s own doc (`cipherstash.rs:339`) says "strings and numbers only".
- **`smallint`** — never matches. Untested.
- **`integer`** — works *only* if the caller encrypts as `F8`. You know this: `v3_json_entry_query_operand_e2e_tests.rs:142-150` documents the `Int` vs `Float` divergence in a comment and the test passes `F8(2.0)` precisely to dodge it. But that reasoning lives in a test comment, not in the gate or the docs.
- **`real` / `double`** — genuinely sound.

The net: `Plaintext::Float` is the only numeric plaintext, and `Plaintext::Text` the only string one, that produces a comparable operand. The family-named domains are labels the caller casts into; the *encoding* is chosen by the plaintext variant, and the gate never checks it.

This is the standard the PR set for itself — `query_<T>_eq` was dropped because it "could never match real data." `query_date_ord` is in the same position, and `query_smallint_ord` for a JSON leaf too. I'd like either an e2e arm proving each kept family's operand round-trips, or the gate extended to ask the operand question and block what can't answer it. `date`/`timestamp` in particular: a date leaf and a text leaf are byte-identically encoded (`Value::String` → `OrderableTerm::String`), so `= query_date_ord` is also the one cast that walks straight around the `text` block you just added.

Two related notes, non-blocking but same root:
- **Timestamp `=` compares representations, not instants.** Nothing normalizes the stored JSON string, so `"…T00:00:00Z"` and `"…T05:00:00+05:00"` are the same instant with different `op` terms. `ScalarKind::Timestamp` says "UTC-normalized", but that's the scalar-column contract; a JSON leaf is stored verbatim.
- **The block is type-level, not value-level.** `query_date_ord`'s CHECK (`{v,i,op}`, no `c`) accepts a text-derived term, so `::eql_v3.query_date_ord` re-opens collated equality. Caller error, but it's the cheap bypass of the fix.

### 3. Docs: the field-context rule doesn't lead a caller to the working path

`json-support.md:153` is otherwise the best paragraph in the doc, but its failure model is number-vs-string:

> an operand encrypted for a *different column*, or for a different JSON scalar type (a number term against a string leaf), has non-corresponding term bytes and **silently returns zero rows with no error**

A caller who follows this literally — integer field, encrypt as an integer — gets `Plaintext::Int`, which *is* a number against a number leaf, and still silently returns zero rows. The rule that actually holds is "encrypt numbers as `Float` and strings as `Text`, whatever the operand domain is named." Worth stating, since the docs lead with the `query_integer_ord` example.

The rest of the doc work is genuinely strong — the per-family table, both ⚠️ callouts with concrete collision pairs, and the `@>` remedy. I verified the table against the shipped SQL and it matches exactly. One nit: "They are absent because…" (`:143`, and the changeset) — they're not absent, they're bound to a raising blocker; "absent" is used three lines later (`:145`) for the genuinely-unbound operands, so the same word carries two meanings.

### 4. Carry-overs from my last review

- **`max_locks_per_transaction` — still no user-facing note.** `git grep` finds it in `tests/docker-compose.yml` only (4 hits). The measurement in that file — *"the uninstaller takes 6,433 locks … against the 6,400 the default affords for the WHOLE cluster"* — is a user-facing fact living in a test fixture. A user running the shipped uninstaller on stock Postgres hits `out of shared memory` and the remedy needs a restart. A `U-NNN` in `docs/upgrading/v3.0.md` would close it.
- **Range oracle — ~⅓ closed.** `>` now has real plaintext-oracle coverage for text (`:481`) and e2e (`:206`, `:290`) — good. `<`, `<=`, `>=` still rest on `json_entry_ord_cross_type_operator_equals_function` (`:188`), which compares operator-vs-function *counts* and is near-tautological (the operator's `oprcode` is that function). Also: that test pivots on row 1, which is `i32::MIN` (`record.rs:30`), so its `<` arm asserts `0 == 0` unguarded. The text pivot (row 2) is well chosen and guarded.
- **`SEL_HELLO_OP` — value fixed, duplication not.** `d7049da3` fixing a constant that named `$.number` while claiming `$.hello` is the strongest possible argument for hoisting it. Two copies remain (`v3_json_entry_cross_type_tests.rs:300`, `v3_jsonb_tests.rs:57`) with nothing tying the hex to the selector. The cross-type suite is now protected by the ordering oracle; `v3_jsonb_tests` is not — its own comment concedes its uses are type-agnostic, which is exactly how the original mis-pin survived. The e2e suite derives the selector from the client and is the right pattern, but it's creds-gated.
- **Bundled commits** — `4890ec14` (rustfmt/clippy) and `1bedb8c0` (doc-anchors + known-failures CI gates) are still here and still unmentioned in the body. Fine changes; with squash-merge they land under `feat(eql_v3):` and become unbisectable. The `max_locks` bump genuinely belongs here (this PR causes it); those two don't.

**Two stale references** from the late reshuffle: `ScalarKind::ope_is_injective` no longer exists (renamed `json_leaf_equality_is_exact`) but is linked from `generate.rs:381` (a rustdoc intra-doc link — likely a `cargo doc` warning), `generate.rs:2081`, and `e2e:224`; and `e2e:223` names `json_entry_text_has_no_equality_operator`, which doesn't exist (it's `json_entry_lossy_family_equality_is_blocked`).

**One test gap worth a decision:** `query_text_search` gets no wrapper *and* no blocker — `json_entry_bloom_operand_is_excluded_from_the_served_list` (`generate.rs:1960`) asserts that on purpose ("the net owns it"). But `json_entry = query_text_search` then flattens to native `jsonb = jsonb` → silent zero rows, which is the exact failure the blockers exist to prevent. If the catch-all net covers it, a test asserting the raise would prove it; right now nothing does.

---

### Summary

Codegen and docs are still the strongest parts of this PR, and the two self-found P1s raised my confidence in the approach rather than lowering it. But the gate that came out of them is half a gate: it proved the leaf encoding injective and never asked whether the operand can produce those bytes — and for `date`/`timestamp` it provably can't, for `smallint` it doesn't, and for `integer` it only does via a plaintext variant that appears in a test comment and nowhere else.

Blocking on the snapshot regen (mechanical) and on a decision for the kept-`=` set — e2e arms per family, or extend the gate. Everything in §3–§4 is happy-to-follow-up.
</content>
</invoke>

src/v3/scalars/integer/json_entry_integer_functions.sql
coderdan: Resolving my own question as the new owner: `eql_v3.query_integer_eq` does exist (it serves scalar `_eq` columns), but it can't serve this seam even as a semantic alias — its CHECK carries `hm` only, never `op`, so an op-routed wrapper would have nothing to read; reshaping the operand domains would ripple into the scalar surface. And equality is now family-conditional (blocked for text/bigint/numeric, and date/timestamp no longer participate at all — JSON has no temporal type), so a uniform `_eq` alias would misrepresent the surface. Equality stays spelled through `query_<T>_ord` where it's sound; `json_entry <op> query_<T>_eq` remains covered by the negative surface guard. If we later want the friendlier spelling, that's a follow-up CIP to introduce op-carrying eq operands — not this PR.

coderdan (approved):
LGTM. A small suggestion to the API. Would also be good to have some E2E tests in place for the `x->'y' > z` form queries.

freshtonic (approved):
## Re-review — approving

My previous pass was `CHANGES_REQUESTED` on two blockers. Both are resolved — the second via a redesign that's stronger than the fix I asked for. Verified against the current head (`cd66ff72`), which has merged `main` (cipherstash-client `0.38.1` → `0.42`, value-inclusive SteVec selectors, new envelope wire format) since I last looked, so I re-derived everything from the committed code rather than the PR body.

### Blocker 1 (mechanical): golden snapshot — **resolved**

`tests/sqlx/snapshots/eql_v3_public_surface.txt` was regenerated (`f6a8240d`, after the surface-changing commits). The phantoms I flagged are gone: **0** `json_entry × query_text_search` entries (was 12), **0** `json_entry × query_{date,timestamp}` entries, and the only `json_entry` eq/neq left are the pre-existing `eq(json_entry, json_entry)` / `neq(json_entry, json_entry)`. The cross-type pairs are ordering-only (`lt/lte/gt/gte`); the `eq`/`neq` for `integer/smallint` now live in `eql_v3_internal` as blockers, correctly invisible to the `nspname = 'eql_v3'` snapshot query. `codegen:parity` is green locally, so the committed surface matches the catalog.

### Blocker 2 (design): the equality gate that "asks about the leaf, never the operand" — **resolved by removing the gate**

Rather than extend `json_leaf_equality_is_exact` to interrogate the operand, the surface stopped offering extract-equality at all. `json_leaf_equality_is_exact` is deleted; **equality never binds the extract surface for any family** — exact field equality is now document containment on a value-inclusive selector (`col @> $1::eql_v3.query_json`), which is injective for *every* type (`"café"` ≠ `"cafe"`, `9007199254740993` ≠ `…992`). That dissolves the "half a gate" objection by construction — there's no per-family equality gate left to be half-right.

The operand-constructibility half is handled directly:
- **`date`/`timestamp` — fully blocked.** All six operators in `eql_v3_internal`; `has_native_json_leaf` returns `false` for them (a date-in-JSON is a text leaf). No client can mint a temporal SteVec ordering term, so nothing is lost — documented at `json-support.md:167`.
- **The "encrypt numbers as floats" contract is now documented and tested.** My worry that `query_smallint_ord` / `query_integer_ord` couldn't produce comparable bytes assumed a `Plaintext::SmallInt`/`Int` operand. The actual contract — `json-support.md:173`, "encrypt numbers as *floats*" with the `Int` raw-cast divergence spelled out — is that the family domain is an inert label for the `op` term; the bytes come from the plaintext encoding. `fresh_numeric_range_operand_matches_plaintext_oracle` proves it: an `F8(2.0)` operand cast as `query_integer_ord` reproduces the `::int > 2` plaintext oracle with fresh ZeroKMS crypto. Because the domain name doesn't touch the `op`, that proof generalizes across the numeric families.

### Carry-overs from my last review — all addressed

- **U-009** added to `docs/upgrading/v3.0.md` for the uninstaller lock budget (`5a57908a`) — the measurement is out of the test fixture and in front of users.
- **Range oracles** at a median pivot + unserved-pair raise coverage (`dee77c51`).
- **`SEL_HELLO_OP`** hoisted to one shared copy (`dee77c51`).
- **Changeset bump** is now `major` — correct, since this rides a breaking SteVec wire-format change (stored documents require re-encryption), not the `patch` I queried.

### Two non-blocking notes for follow-up

1. **The PR description is stale.** It still describes the old "ordering everywhere, equality only where the leaf encoding is exact" design against client 0.38.1, including the per-family equality table — none of which is what ships. Since squash-merge is recommended and the body becomes the commit message, please update it to the shipped design (ordering-only extract surface; equality via containment) before merging.
2. **e2e ordering breadth.** The numeric ordering arm exercises `query_integer_ord` as the representative; the earlier per-family sweep (smallint/real/double + `_ope`) is gone. Fine given the inert-domain argument above, but a one-line comment in the test noting *why* one arm covers the family set would save the next reader the same derivation I just did.

Codegen/catalog unit tests (94) and `codegen:parity` pass locally; the SQLx + e2e suites are creds-gated to CI. No blockers. Nice work — the pivot to value-inclusive-selector equality is a cleaner answer than the gate I was pushing for.

PR #413, opened 2026-07-19
feat(eql_v3): SteVec value-inclusive selectors + envelope wire format

coderdan:
Reworks the encrypted-JSON (SteVec) surface so exact field equality is **value-selector presence** for every scalar type, and restructures the SteVec document wire format around a per-document key header with selector-derived entry nonces.

Stacks on #410 (json_entry ↔ query_<T> cross-type operators) — GitHub will retarget this to `main` when #410 merges. This is the EQL half of a cross-repo change; the encryption-client and protect-ffi halves ship in their own branches and merge bottom-up (client release first).

## What changed

**Value-inclusive selectors.** Each JSON node emits a path entry `{s, c, op?}` plus a value entry whose selector `SEL(type-tag ‖ path ‖ canonical(value))` bakes in the value, so the selector's *presence* in a stored document is an exact, injective match — correct for `text` / `bigint` / `numeric`, where the old `op`-based comparison collided (`"café"` == `"cafe"`, `9007199254740993` == `…992`). Field equality is now document containment: `WHERE col @> $1::eql_v3.query_json`. The per-value `hm` term is retired; extract-surface `=` / `<>` is **blocked** for every family (a path entry carries no value selector), while ranges stay on the extract surface. The single-entry containment operators are dropped.

**Envelope wire format.** A document is now `{v, k:"sv", i, h, sv:[…]}`. The key-retrieval material (IV, tag, descriptor, keyset) is hoisted **once** to a document-level `h` field instead of being repeated inside every entry; there is no root `c` (the root ciphertext is the root `sv` entry). Each entry's `c` is the raw AEAD output only, and its nonce derives from the entry's own selector (`hex_decode(s)[..12]`) rather than a shared per-document IV. This shrinks a document ~30% (measured on the fixture: 2238 → 1505 chars) and closes an intra-document leak — equal values at different paths, and the constant value-entry plaintext, no longer produce byte-identical ciphertexts, and each ciphertext is cryptographically bound to its selector (a within-document graft fails to decrypt). Value entries encrypt a fixed versioned sentinel (not the empty string), so a value entry stays distinguishable from a genuine `""` leaf. `eql_v3.jsonb_array_elements_text` is removed (a bare-ciphertext stream is no longer independently decryptable).

Full design rationale (including the options considered and the sentinel / descriptor / array-marker decisions) is in `docs/plans/stevec-envelope-key-header-rfc.md`.

## Breaking

Stored SteVec documents must be **re-encrypted** — there is no mechanical v2→v3 conversion (`from_v2` fails closed on a SteVec document and on `hm`-bearing query entries). The changeset is a `major` bump and upgrade note **U-010** documents the migration.

## Verification

- Bindings gain the `KeyHeader` / `EntryCiphertext` newtypes and a term-less-optional `SteVecTerm`; TS/JSON-Schema regenerated; `types:check` / `codegen:parity` are a deterministic no-op after regeneration.
- Fixtures regenerated through the new client; the SteVec document confirms the shape (`h` present, no root `c`, no `hm`, all term-less ciphertexts distinct).
- Green: `encrypted_domain` 2403, `v3_jsonb_tests` 68, `v3_json_entry_cross_type` 7, `v3_json_entry_query_operand_e2e` 9, `v3_json_storage` 8, `v3_jsonb_bindings` 2, `v3_jsonb_operator_surface` 7, `payload_schema` 32, `v3_public_surface` 3 (snapshot regenerated), `v3_privilege` 5, `v3_operator_equivalents` 7, `v3_uninstall` 3; eql-bindings all suites.

Docs updated: `json-support.md`, `sql-support.md`, `docs/upgrading/v3.0.md` (U-010).

https://claude.ai/code/session_01FnhDhqdknNLWrHWPZhQtWq

Supersedes #412 after removing private issue identifiers from its public branch and history.


freshtonic (commented):
## Review — SteVec value-inclusive selectors + envelope wire format

### Overview

This reworks the encrypted-JSON (SteVec) surface along two axes:

1. **Value-inclusive selectors.** Exact field equality becomes *value-selector presence* — `SEL(tag ‖ path ‖ value)` bakes the value into the selector, so containment (`col @> $1::eql_v3.query_json`) is an injective, keyed-MAC match. This fixes real correctness bugs where the old `op`-based comparison collided (`"café" == "cafe"`, `9007199254740993 == …992`). The per-value `hm` term is retired; extract-surface `=`/`<>` is blocked for every family; ranges stay on the extract surface.
2. **Envelope wire format.** Key-retrieval material is hoisted once to a document-level `h` header; each entry's `c` is raw AEAD output with a selector-derived nonce (`hex_decode(s)[..12]`). This shrinks documents ~30% and closes an intra-document leak (equal values at different paths, and the constant value-entry plaintext, no longer produce byte-identical ciphertexts).

It's a `major` bump: stored documents must be re-encrypted (`from_v2` fails closed), documented in upgrade note U-010.

Engineering quality is high — the rationale is captured in-line and in the RFC, the design is internally consistent across SQL / Rust bindings / codegen / catalog, and test coverage is broad (fail-closed conversion, term-optional parsing, distinct-ciphertext fixture assertions).

### Strengths worth calling out

- **The `to_ste_vec_query(b)` normalization on the operator RHS is a genuine correctness fix, not just tidy-up.** Native jsonb `@>` requires the needle element be a *subset* of a container element; a needle carrying `op` against an `s`-only normalized container (`{s:X}` vs `{s:X,op:Y}`) would never match. Normalizing both sides through `to_ste_vec_query` is required. The new `to_ste_vec_query(eql_v3.query_json)` overload handles document-derived needles carrying a stray `op`.
- **Replacing `ste_vec_contains` (opaque plpgsql loop) with `to_ste_vec_query(a) @> to_ste_vec_query(b)` is an indexability *improvement*** — the LHS now structurally matches a `GIN (eql_v3.to_ste_vec_query(col)::jsonb jsonb_path_ops)` functional index, which the plpgsql loop could never use.
- **The Rust bindings are strictly better.** The flattened `SteVecTerm` enum forced serde to silently disable `deny_unknown_fields`; replacing it with an explicit `Option<OpeCllw> op` (plus explicit optional `i`/`v`/`h` for `->`-grafted metadata) restores `deny_unknown_fields`, so the client parser and published JSON Schema now enforce the same boundary as the SQL CHECKs.
- **Blockers are correctly `LANGUAGE plpgsql`, non-`STRICT`** — matching the project footgun rules; a `STRICT`/`LANGUAGE sql` blocker could be elided or NULL-bypassed.
- **`from_v2` fails closed cleanly** — `UnconvertibleSteVecDocument` / `UnconvertibleSteVecQuery` with clear messages, and the SQL CHECKs reject stale `hm`-bearing payloads loudly rather than degrading.

### Findings

**1. (Minor — polish) The CIP-identifier scrub left mangled comments.** The string-based removal of private issue identifiers left grammatical debris in several places:

- `crates/eql-domains/src/fixtures/kind.rs` — a rustdoc line reading `///: a family whose values…` (leading colon where `(CIP-3526)` was removed).
- `crates/eql-domains/src/spec.rs` — a bare `///.` doc line (was `(CIP-3442).`).
- `crates/eql-domains/src/tests.rs` — `// prefix —  — pinned separately below` (double em-dash gap).
- `crates/eql-codegen/src/generate.rs` — test-section comment `// --- : json_entry <-> query_<T>…` (was `CIP-3526:`).
- Two more bare `///.` doc lines and a `@@ fuzzy match —  — and are blockers` gap in the fixtures/docs prose.

None break compilation or CI (comments/prose only; `docs:validate` greps SQL doxygen), but the `kind.rs` one renders as a rustdoc line starting with `:`, and this PR is the v3 reference. Worth one cleanup pass — grep for `///:`, `///.`, `—  —`, `--- :`.

**2. (Question — design) Asymmetric `=` availability on `json_entry`.** `=`/`<>` between `json_entry` and a `query_<T>_ord` operand are now blocked (raise), which is well-justified. But `=`/`<>` between two `json_entry` values (`src/v3/json/operators.sql`) are retained, backed by `eq_term` → `op` byte-comparison — the exact lossy comparison (`"café" == "cafe"`) the rest of the change routes *away* from. It's documented in the function `@note`, so it's deliberate — but is it intended that entry-to-entry `=` stays offered (lossy) while entry-to-operand `=` is blocked? If there's no concrete use case for entry-to-entry equality, consider blocking it too for consistency; if there is, the asymmetry deserves an explicit line in `json-support.md`.

**3. (Confirm) Dropped operators fail loud.** The single-entry `@>`(json_search, json_entry) and its `<@` reverse are removed; a query still using them errors with "operator does not exist" — the right behavior for a major break. Worth a glance that `v3_jsonb_operator_surface` asserts the *absence* of the dropped operators, not just the presence of the survivors.

### Risk assessment

Low-to-moderate, appropriate for a `major`. The cryptographic core (selector-derived nonces, value-selectors, `h` hoisting) lives in cipherstash-client, not here — the EQL side carries/grafts opaque material and enforces shape, which it does correctly and fail-closed. The main residual risk is the cross-repo lockstep the PR itself flags (client must release first); nothing in this diff can validate that ordering, so it rests on release discipline. Generated artifacts (schema JSON, TS, bundled SQL, fixtures) are deterministic and out of scope.

**Recommendation:** Approve after the finding-1 comment cleanup; treat findings 2 and 3 as questions to confirm rather than blockers.

<sub>🤖 Generated with [Claude Code](https://claude.ai/code)</sub>

tobyhede (commented):
Excellent work.

Must fix

  1. Collision cases untested. café/cafe and 9007199254740993 appear only in prose. Add two assertions to
  the e2e suite — these are the PR's headline claims.
  2. Fresh-operand range coverage deleted. ste_vec_query_term / ste_vec_query_selector
  (tests/sqlx/src/fixtures/cipherstash.rs:315,338) now have zero callers. Ranges lost their
  independently-encrypted-operand test; oracle correctness is still covered.
  3. Empty-string sentinel untested. Add "empty": "" to documents()
  (tests/sqlx/src/fixtures/v3_ste_vec.rs).
  4. Multi-selector needle lacks AND-discrimination. v3_jsonb_containment_mixed
  (tests/sqlx/tests/v3_jsonb_tests.rs:425-450) uses two selectors present in every row — an OR
  implementation passes. Pair the constant selector with a row-specific one.
  5. Ticket-ID strip artifacts, 15 sites, 4 in live assertion messages.
    - crates/eql-codegen/src/bindings.rs:864,1115
    - crates/eql-codegen/src/dump.rs:73
    - tests/sqlx/src/property.rs:115,141,215,284,577,603,608,613
    - tests/sqlx/src/fixtures/cipherstash.rs:545
    - tests/sqlx/tests/encrypted_domain/ope/support.rs:259,261
    - tests/sqlx/tests/v3_jsonb_operator_surface_tests.rs:321
    - tests/sqlx/snapshots/README.md:196
  6. Stale comments contradicting current behaviour.
  tests/sqlx/tests/encrypted_domain/jsonb_entry.rs:139,171 still say entry = reduces through eq_term; it
  now raises. tests/sqlx/src/fixtures/v3_ste_vec.rs:69,110 still reference hm leaves.

  Should fix

  7. Ciphertext-graft binding asserted by comment only (v3_jsonb_tests.rs:1366-1369). Testable —
  cipherstash-client 0.42 is already a dep and exposes decrypt_eql_fallible.
  8. Containment implemented twice. src/v3/json/operators.sql:135-141 duplicates
  src/v3/json/functions.sql:328-334 byte-for-byte. Verified on live PG 17 that delegating produces an
  identical GIN plan, so the duplication buys nothing.
  9. Document validator missing unknown-key rejection. This PR added val - ARRAY[…] = '{}' to the entry
  and query validators but not to eql_v3_is_valid_ste_vec_document_payload
  (src/v3/json/types.sql:94-114).

  Nits

  10. Rename eql_v3.eq_term(json_entry) (src/v3/json/functions.sql:121) — it and ord_term now read the
  same op field; a function named eq_term documented as "never use for equality" invites misuse. Not a
  bypass; retention is deliberate and changeset-declared.
  11. from_v2_query doesn't check k (crates/eql-bindings/src/from_v2/mod.rs:253-291). Unreachable with
  real payloads. One defensive line.
  12. RFC array-wildcard residual (docs/plans/stevec-envelope-key-header-rfc.md:57-60) — sibling
  ciphertexts under a shared wildcard selector remain permutable; the "freely permutable" past-tense
  framing implies that's fully closed.
  13. Document the trust-boundary shift in docs/reference/json-support.md: "a needle constrains a value"
  is now a client obligation, unenforceable database-side.

  Verified clean

  Snapshot baselines (no masked test drops) · deny_unknown_fields strictly stricter than the old form ·
  from_v2 stored-path match exhaustive · codegen/bindings regenerate byte-identical · blocker discipline
  correct and asserted · major bump and U-010 appropriate.

PR #421, opened 2026-07-23
fix(v3): guard empty-bloom needle in eql_v3.matches (CIP-3606)

freshtonic:
## Summary

`eql_v3.matches` (the `@@` fuzzy-match operator on `public.eql_v3_text_match` / `public.eql_v3_text_search` / `public.eql_v3_text_search_ore`) reduces to bloom array-containment `match_term(a) @> match_term(b)`. An **empty needle bloom** (`{}`) is contained by every value, so a search term with no n-gram tokens (a sub-trigram string below the tokeniser floor) matched **every row in the table**, silently.

Fixes [CIP-3606](https://linear.app/cipherstash/issue/CIP-3606) / closes #416.

## The fix

The codegen renders the `@@` wrapper with an inline empty-needle guard giving `LIKE`-shaped semantics — an empty needle matches only a value whose own bloom is also empty (`'' LIKE ''` → true; `'catty' LIKE ''` → false):

```sql
SELECT match_term(a) @> match_term(b)
   AND (cardinality(match_term(b)) > 0 OR cardinality(match_term(a)) = 0)
```

| stored value | needle | want | before | after |
| --- | --- | --- | --- | --- |
| empty | empty | true | true | true |
| **non-empty** | **empty** | **false** | **true** ❌ | **false** ✅ |
| empty | non-empty | false | false | false |
| non-empty | non-empty | containment | containment | containment |

### Why inline, and why non-`STRICT`

The [issue's proposed fix](https://github.com/cipherstash/encrypt-query-language/issues/416#issuecomment-5041731628) (a `CASE` in `matches()`, or a `bloom_match` helper) would **break GIN index engagement** — `operator_surface.rs`, `tasks/test/splinter.sh`, and the `bare_matches_operator_uses_functional_index` test all require `matches()` to inline to a single top-level `match_term(col) @> needle`. Burying that `@>` inside a `CASE`/helper makes it non-indexable.

The guard is therefore **inline**: the top-level `@>` conjunct stays an indexable qualifier, so `WHERE col @@ $1` (bind parameter) or a literal needle still inlines and engages a **Bitmap Index Scan**, with the guard riding along as a cheap recheck filter. The wrapper is now **non-`STRICT`**: PostgreSQL will not inline a `STRICT` SQL function whose body carries the guard's `AND`/`OR`, and the body propagates `NULL` on a `NULL` operand on its own — so dropping `STRICT` preserves the NULL semantics while restoring inlinability.

**One narrow, documented trade-off:** the guard references the needle twice, so a needle supplied as an **uncorrelated subquery** (`col @@ (SELECT …)`) no longer inlines and falls back to a sequence scan. The result is still correct; only the index acceleration is lost. Parameters and literals — the normal query shapes — are unaffected. Verified with `EXPLAIN` across constant / bind-param / subquery needles.

## Tests

- New **real-ciphertext** fixture `v3_text_empty_bloom` (`"pq"` → a genuine `bf: []` from the actual crypto; `"aardvark"` → populated), following the `v3_text_empty` precedent — no synthetic blobs.
- Four behavioural tests covering the full LIKE truth table, plus a premise guard asserting the fixture really produces an empty bloom.
- Index-engagement tests switched from an uncorrelated subquery to a literal needle (real-usage shape).
- Corrected the pre-existing `text_smoke` test that had encoded the **bug** as intended behaviour ("every filter must match the empty needle").
- Changeset (patch) + upgrade note **U-011**.

### Verification (all green locally)

codegen (113), text_match (21), text_smoke (8), match_smoke, public-surface (12), operator-equivalents (3), lint (7), splinter (`ERROR=0`), `types:check` (no bindings drift), `codegen:parity`, build + self-contained.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fuzzy text matching so empty-needle `eql_v3` queries no longer match all rows.
  * Empty-needle behavior now follows `LIKE ''` semantics (only empty blooms match empty blooms).
  * Kept functional GIN index usage for non-empty queries; added coverage for index behavior when empty-needle guards are applied.
* **Documentation**
  * Updated upgrade notes with the breaking-change guidance for empty-needle fuzzy matching (U-011), including compatibility considerations.
* **Tests**
  * Added SQLx fixtures and regression tests for the empty-bloom guard behavior.


tests/sqlx/tests/encrypted_domain/text/text_match.rs
```diff
@@ -287,3 +304,154 @@ async fn bloom_matches_where_like_would_not(pool: PgPool) -> anyhow::Result<()>
...
+    let pq = empty_bloom_payload_for(&pool, "pq").await?;
+
+    let (op_hit, fn_hit): (bool, bool) = sqlx::query_as(
+        "SELECT ($1::jsonb::public.eql_v3_text_match) @@ ($2::jsonb::public.eql_v3_text_match),
+                eql_v3.matches($1::jsonb::public.eql_v3_text_match, $2::jsonb::public.eql_v3_text_match)",
+    )
+    .bind(&aardvark)
+    .bind(&pq)
+    .fetch_one(&pool)
+    .await?;
```
coderdan, line 373: Not necessary for this pull request, but this pattern would be quite easy to create a property test for, because the properties in plain text are easy to establish. Have something we can do in a follow-up. There might be more pathological examples.

coderdan (approved):
Nice work.

tests/sqlx/tests/encrypted_domain/text/text_match.rs
freshtonic: Good idea, will create follow up ticket

PR #423, opened 2026-07-27
feat(eql_v3): add grouped_value aggregate

freshtonic:
Re-creates the eql_v2 grouped_value aggregate on the self-contained eql_v3 surface. It returns a representative (first non-null) value per group so an encrypted column can be projected while grouping by its equality term: GROUP BY eql_v3.eq_term(col) groups encrypted rows by equality, but PostgreSQL then rejects a bare SELECT col because it cannot prove the column is constant within each group. Wrapping the column in eql_v3.grouped_value(col) satisfies the GROUP BY rule.

A single aggregate over jsonb serves every encrypted-domain type (all are jsonb-backed domains). The STRICT state function reproduces the v2 COALESCE semantics; PARALLEL SAFE with a combine function matches the generated min/max shape.

CIP-3657

## Summary by CodeRabbit

* **New Features**
  * Added `eql_v3.grouped_value` for projecting encrypted values when grouping by their equality term.
  * Returns a representative non-null value for each group without decrypting or comparing data.
  * Supports parallel aggregation.

* **Bug Fixes**
  * Enables grouped queries that previously failed when selecting the original encrypted column alongside the grouping expression.

* **Tests**
  * Added coverage for null handling, ordering, grouping errors, and successful grouped projections.


tests/sqlx/tests/v3_grouped_value_tests.rs
```diff
@@ -0,0 +1,131 @@
...
+//! contract):
+//!
+//!   * `first_non_null_per_group` — with an explicit intra-aggregate ORDER BY the
+//!     result is the first non-null value; NULLs are skipped.
+//!   * `all_null_group_is_null` — a group of only NULLs aggregates to NULL.
+//!   * `bare_encrypted_column_projection_is_rejected` — the motivating failure:
+//!     `SELECT col ... GROUP BY <expr on col>` raises grouping_error (42803).
+//!   * `grouped_value_makes_group_by_term_projection_valid` — the fix: the same
+//!     shape with `grouped_value(col)` succeeds and returns a representative
+//!     member of each group.
```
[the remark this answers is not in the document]
freshtonic: `grouped_value` does not require encrypted data to prove it works.

src/v3/aggregates.sql
```diff
@@ -0,0 +1,106 @@
...
+--! @file v3/aggregates.sql
+--! @brief Schema-generic aggregates over the eql_v3 encrypted-domain surface.
+--!
+--! Re-creates the eql_v2 `grouped_value` aggregate on the self-contained eql_v3
+--! surface. Unlike the per-type `min`/`max` aggregates (generated into
+--! src/v3/scalars/<T>/ from the catalog, one per ORD-term domain), this one is
+--! hand-written and generic: every eql_v3 encrypted-domain column type is a
+--! jsonb-backed domain, so a single aggregate over `jsonb` accepts any of them
+--! (a domain value implicitly casts to its base type) and returns a value
+--! unchanged. There is nothing type-specific to generate.
```
coderdan: This reads more like a commit message than docs. Would be better to frame the docs for a user who is using group by in EQL for the first time instead of for a stashie who already has the context.

coderdan (approved):
Worth adding some tests for SELECT DISTINCT too? I assume it's the same mechanism?

src/v3/aggregates.sql
freshtonic: Good call — reframed in 88dae689. The header now leads with the task (running `GROUP BY` over an encrypted column and getting the encrypted value back), walks a first-time EQL user through it with worked examples, and moves the eql_v2/codegen provenance into a short `@internal` note.

src/v3/aggregates.sql
```diff
@@ -0,0 +1,125 @@
...
+--!
+--! Encryption in EQL is non-deterministic: encrypting the same value twice gives
+--! two different ciphertexts. So you cannot group rows by comparing the stored
+--! ciphertext. Instead you group by the column's *equality term*,
+--! `eql_v3.eq_term(col)` — a deterministic keyed hash that is identical for equal
+--! plaintexts. Counting how many rows share each encrypted value looks like this:
+--!
+--!   SELECT eql_v3.eq_term(email), count(*)
+--!   FROM users
+--!   GROUP BY eql_v3.eq_term(email);
```
freshtonic:
```
--!   SELECT eql_v3.eq_term(email), count(*)
--!   FROM users
--!   GROUP BY eql_v3.eq_term(email);
```

Should be:

```
--!   SELECT count(*)
--!   FROM users
--!   GROUP BY eql_v3.eq_term(email);
```

There is zero point in returning the eq_term itself.

src/v3/aggregates.sql
```diff
@@ -0,0 +1,125 @@
...
+--! `count`), you do not need `grouped_value` — use `DISTINCT ON` on the equality
+--! term, which lets you project the column directly:
+--!
+--!   SELECT DISTINCT ON (eql_v3.eq_term(email)) email
+--!   FROM users
+--!   ORDER BY eql_v3.eq_term(email);
+--!
+--! Deduplicate on `eql_v3.eq_term(col)`, never on the column itself: a plain
+--! `SELECT DISTINCT email` compares the raw ciphertext, so two encryptions of the
+--! same value are NOT collapsed.
```
coderdan, line 52: This approach would not be decryptable. Can it use the same function we use for GROUP BY?

coderdan (changes requested):

tests/sqlx/tests/v3_grouped_value_tests.rs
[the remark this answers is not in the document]
freshtonic: Fixed in 5e9a6493. Reworded the module doc to say it straight: `grouped_value` returns its input unchanged and has nothing to do with encryption, so plain `jsonb` literals fully exercise it — dropped the "no real ciphertext needed" framing that made it sound like a stand-in.

src/v3/aggregates.sql
freshtonic: Fixed in 5e9a6493 — dropped `eql_v3.eq_term(email)` from the SELECT list, now just `SELECT count(*) ... GROUP BY eql_v3.eq_term(email)`.

src/v3/aggregates.sql
freshtonic: I'm not sure what you mean here - it's not clear which fragment you're referring to.

coderdan:
```
SELECT DISTINCT ON eql_v3.eq_term(email)
```

It isn't very useful because the decryptable ciphertext isn't returned.

My question is can we use the same grouping function added in this PR instead?

freshtonic: That fragment doesn't appear in the SQL doc.  The closest I can find is this one (which does use `eql_v3.eql_term(email)` to get a distinct value then projects `email` which _is_ decryptable):

```
--!   SELECT DISTINCT ON (eql_v3.eq_term(email)) email
--!   FROM users
--!   ORDER BY eql_v3.eq_term(email);
```

coderdan: Yes, that's the fragment I'm referring to and NO it isn't decryptable. `eq_term` returns an `HMAC` which is a one-way function.

coderdan: I'll try to make this as clear as I can:

Can the DISTINCT queries use the form?

```sql
SELECT DISTINCT ON eql_v3.grouped_value(email) FROM users;
```

freshtonic: I know SEM terms are not decryptable. But the doc I pointed out to you is not returning the SEM term: it's returning the email column.

There are two possibilities for where a misunderstanding could be: 

1. You didn't notice the email column in the projection, or 

2. I misunderstood how DISTINCT works - perhaps it only works on the physical representation of its argument (does not use operators)

coderdan: The doc says in this PR:

> SELECT DISTINCT email` compares the raw ciphertext

Because the ciphertext is randomized, this will never actually return distinct email values. Because the type is a domain type the planner will compare the values as `jsonb` (the base type).

My question was can we use the same solution we used for GROUP BY in the DISTINCT queries. I'm not making a comment about the docs.

freshtonic: I don't think we can use a function with this form:`SELECT DISTINCT foo FROM blah`. There's nowhere to select the thing you want to use for DISTINCT separately from the columns you want to select.

But this already works: `SELECT DISTINCT ON (eql_v3.eq_term(col)) col` and being able to means in `eql-mapper` we can convert the first form to the second form.

coderdan (approved):

src/v3/aggregates.sql
coderdan: I've approved now based on a conversation with @freshtonic. I missed that `DISTINCT ON` in the example above uses the first term `(eql_v3.eq_term(col))` to determine uniqueness but then also returns the full value `col` (I thought it was an alias).

PR #433, opened 2026-08-04
docs(reference): document eq_term/ord_term for SELECT DISTINCT + ORDER BY

freshtonic:
## CIP-3677 — Document `SELECT DISTINCT` + `ORDER BY` on encrypted columns (EQL docs)

Makes the `DISTINCT` / `ORDER BY` behaviour that landed in the Proxy EQL mapper discoverable from the EQL reference docs, via the `eq_term` / `ord_term` extractors.

### Changes (docs only)
- `docs/reference/database-indexes.md` — extends the existing **GROUP BY / DISTINCT** section:
  - `SELECT DISTINCT` on an encrypted column is keyed on `eql_v3.eq_term(col)` (dedup by plaintext; raw-column `DISTINCT` never collapses identical plaintexts because ciphertext `c` is randomised). Includes the rewrite and the capability-error consequence for storage-only domains (`eql_v3_boolean`).
  - `SELECT DISTINCT … ORDER BY <encrypted col>` additionally requires `eql_v3.ord_term(col)`, with the subquery-projecting-the-term rewrite; the ordering term is never returned to the client and aliases are preserved.
- `docs/reference/eql-functions.md` — cross-reference from the Index Term Extraction section.

### Scope note
This covers the **encrypt-query-language repo** portion of CIP-3677 only. The public **website-docs** portion of the ticket lives in a separate repository and is tracked/handled there.

No SQL/Rust/test changes; no changeset (docs-only, describing Proxy behaviour rather than an EQL code/behaviour change).

Part of CIP-3677

### DIFF
--- a/docs/reference/database-indexes.md
+++ b/docs/reference/database-indexes.md
@@ -189,6 +189,38 @@ Why the raw column does not scale: `GROUP BY col` uses the entire encrypted payl
 
 Pick the extractor the domain actually has: `eq_term` exists only on the `hm`-carrying domains. On a numeric-and-time `_ord` / `_ord_ope` / `_ord_ore` column, group on the ordering extractor instead — `GROUP BY eql_v3.ord_term(col)` (or `ord_term_ore(col)`). The ordering term is injective for those types, so it is an exact grouping key, and the column's ordering btree covers it.
 
+#### `SELECT DISTINCT` is keyed on `eq_term`
+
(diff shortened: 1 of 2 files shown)


freshtonic (commented):
Comment-only review.\n\nI did not find an issue in the documentation changes. The explanation correctly separates randomized payload ciphertext from deterministic equality/ordering terms, describes the capability requirements, and keeps the ordering term out of the client projection. Documentation validation is green.

tobyhede (approved):

PR #436, opened 2026-08-16
Retarget ste_vec_contains rename changeset to patch

tobyhede:
## Summary
- `.changeset/rename-ste-vec-contains.md` was declaring `major`. It should be `patch`: 3.0.1 shipped an operator behaviour change (`@>`/`<@` → `@@`) as a patch, and this rename leaves the operators (and the `jsonb_contains`/`jsonb_contained_by` function-form entry points) untouched — same level. As `major` this would have released 4.0.0; the downstream absorbed tree (`cipherstash/stack`) expects `3.0.5`.
- Rewrote the changeset body: verified against `src/v3/json/functions.sql` that `eql_v3.jsonb_contains(jsonb, jsonb)` and `eql_v3.jsonb_contained_by(jsonb, jsonb)` are untouched by the rename (a PostgREST deployment calling those by name sees no change). The previous prose overstated the blast radius. The rewritten body scopes what actually breaks (hand-written SQL naming `ste_vec_contains` directly — queries, views, RLS, per-function grants) and notes the schema-wide grant recipes in `docs/reference/permissions.md` pick the new name up automatically, while a stale per-function grant fails loudly rather than silently. Also notes the installer already opens with `DROP SCHEMA IF EXISTS eql_v3 CASCADE` on every install/upgrade — not new behaviour from this rename.

This is the last release from this repository — the EQL subsystem has been absorbed into `cipherstash/stack`, and 3.0.5 is what puts a version on npm matching what the absorbed tree already carries.

## Test plan
- [x] Confirmed `eql_v3.jsonb_contains`/`jsonb_contained_by` (raw jsonb, 2-arg) are separate overloads from the renamed `jsonb_document_contains`, unaffected by the rename
- [x] Confirmed `release/cipherstash-encrypt.sql` opens with `DROP SCHEMA IF EXISTS eql_v3 CASCADE`
- [ ] Merge triggers Changesets Version PR bumping `@cipherstash/eql` 3.0.4 → 3.0.5 (patch)

## Summary by CodeRabbit

- **Changes**
  - Renamed `eql_v3.ste_vec_contains` to `eql_v3.jsonb_document_contains`.
  - Existing containment operators and raw JSONB function entry points remain unchanged.
  - Direct SQL callers using the former function name should update their queries.
  - Function-specific access grants now apply to the renamed function.
  - This update is delivered as a patch release.

### DIFF
--- a/.changeset/rename-ste-vec-contains.md
+++ b/.changeset/rename-ste-vec-contains.md
@@ -1,10 +1,23 @@
 ---
-'@cipherstash/eql': major
+'@cipherstash/eql': patch
 ---
 
 **`eql_v3.ste_vec_contains` is renamed to `eql_v3.jsonb_document_contains`.** This
 consolidates the last `ste_vec_*`-named public object into the `jsonb_*` family,
 matching the earlier renames of the SteVec entry/query surface (`jsonb_entry`,
-`jsonb_query`). The function backs the `json` `@>` / `<@` containment operators;
-its behaviour is unchanged. Callers that invoke the function by name (Supabase /
-PostgREST, which call functions rather than operators) must update the name.
+`jsonb_query`). The `@>` / `<@` containment operators are untouched, and so are
+the raw-jsonb function-form entry points, `eql_v3.jsonb_contains(jsonb, jsonb)`
+and `eql_v3.jsonb_contained_by(jsonb, jsonb)` — a PostgREST deployment calling
+those by name sees no change.
+
+What breaks is hand-written SQL that names `ste_vec_contains` directly: queries,
+views, and RLS policies that call it, and any per-function `GRANT EXECUTE ON
+FUNCTION eql_v3.ste_vec_contains(...)`. The schema-wide grant recipes in
+`docs/reference/permissions.md` (`GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA
+eql_v3 ...`) pick the new name up automatically on their next run; a stale
+per-function grant referencing the old name fails loudly at apply time
+(`function eql_v3.ste_vec_contains(...) does not exist`) rather than silently
+granting nothing. And underneath all of it: the installer opens with `DROP
+SCHEMA IF EXISTS eql_v3 CASCADE`, so every EQL install already drops and
+recreates grants, functional indexes, and dependent views on upgrade — that is
+not new behaviour introduced by this rename.


freshtonic (changes requested):
Requesting changes for two blocking issues:

1. The `patch` bump contradicts this repository's documented versioning policy. This changeset explicitly says the public function rename breaks hand-written queries, views, RLS policies, and per-function grants. `CLAUDE.md` defines patch releases as bug fixes with no behaviour changes and major releases as changes that break the public interface, including function names. Please either keep this as `major` or retain `eql_v3.ste_vec_contains` as a compatibility alias so the release is genuinely non-breaking. A previous release using a patch bump, or a downstream expectation of 3.0.5, does not make this rename patch-compatible.

2. The final sentence is factually incorrect: the installer does not "recreate grants, functional indexes, and dependent views." `DROP SCHEMA ... CASCADE` drops those dependent/user-owned objects, while EQL deliberately issues no `GRANT`/`REVOKE` and cannot recreate application views, indexes, or RLS policies. Please say that this destructive upgrade behaviour is pre-existing and that deployments must recreate/reapply those objects.

Also, `CLAUDE.md` asks changeset entries to be a single dense paragraph; please consolidate the three-paragraph body when revising it.

tobyhede:
Closing — superseded by #437.

This PR's goal (retarget the `ste_vec_contains` rename changeset from `major` to `patch`) was already achieved on `main` by 142f41d8 ("fix(json): restore deprecated ste_vec_contains aliases"), which made the same frontmatter change *and* went further: it restored the deprecated alias in `src/v3/json/functions.sql` and registered it in `tests/sqlx/snapshots/eql_v3_public_surface.txt`.

Rebasing this branch onto `origin/main` conflicts on `.changeset/rename-ste-vec-contains.md`, and resolving it correctly leaves the branch byte-identical to `main` — git drops cb0b3633 as empty.

The conflict is substantive, not just textual. Because #437 restored the alias, this PR's body is now inaccurate:

> What breaks is hand-written SQL that names `ste_vec_contains` directly: queries, views, and RLS policies that call it, and any per-function `GRANT EXECUTE ON FUNCTION eql_v3.ste_vec_contains(...)`.

Nothing there breaks — the old name still resolves. Merging this would replace `main`'s accurate "the old name remains available as a deprecated compatibility alias" wording with a breakage warning that no longer applies, in the 3.0.5 changelog.

`main` carries the correct `patch` changeset, and Version Packages PR #434 is bumping 3.0.4 → 3.0.5 with the right entry. No action needed here.

PR #439, opened 2026-08-18
fix(deps): bump jsonwebtoken to 10.4.0

tobyhede:
Bumps `jsonwebtoken` to 10.4.0, pulled in transitively via `stack-auth` 0.42.2 (now depends on `jsonwebtoken ^10.3.0` instead of the previous `^9.3.1` pin). No direct usage of jsonwebtoken in this repo — clean transitive bump, workspace builds successfully.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4885,7 +4885,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,9 +612,9 @@ dependencies = [
 
 [[package]]
 name = "cipherstash-config"
-version = "0.42.0"
+version = "0.42.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "56cde3aaa5e2916a40530932f142c37ad6202835a6d221ce6c46fc14bfec8fc5"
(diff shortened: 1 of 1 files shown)


freshtonic (approved):
Approving.

Verified the lock-only bump does what the description says:

- `jsonwebtoken` 9.3.1 → 10.4.0, arriving via `stack-auth` 0.42.2, which declares `jsonwebtoken = "10.3.0"` with `default-features = false, features = ["aws_lc_rs", "use_pem"]`. That swaps the crypto backend from `ring` to `aws-lc-rs`, which is the real substance of this diff — worth naming explicitly, since the PR body reads as a routine version-number bump.
- No `Cargo.toml` change is needed here: nothing in this workspace pins the 0.42.x crates exactly, so the lock moves on its own. Confirmed no direct `jsonwebtoken` usage.
- The backend swap adds no new build-toolchain burden. `aws-lc-rs`/`aws-lc-sys` are already in the lock on `main` (via `rustls`), so nothing new has to compile and no CMake/C requirement is introduced by this change.

One incidental change worth noting, not a blocker: `winapi-util` 0.1.11's `windows-sys` resolves 0.61.2 → 0.48.0. That's within its declared range (`>=0.48.0, <=0.61.*`, `cfg(windows)`-gated only), and it collapses a duplicate `windows-sys` copy rather than adding one — harmless, just unrelated to the stated change.

Full CI green, including the PG17 shards and the e2e property suite.

Cargo.lock
```diff
@@ -4882,7 +4885,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
```
coderdan: That's a fairly substantive drop. Any idea why that's changed so much?

coderdan (approved):
Looks fine, but there's one dependency that downgraded by quite a lot. Maybe worth checking before merging.

